PR #2, opened 2021-07-06
tlcs900h: Fix immediate mask in BIT #,r

invertego:
The BIT (imm, reg) instruction uses the low 4 bits of its immediate operand even
when the register operand is only 8 bits wide. When the register is 8 bits and
the immediate value is between 8 and 15, ZF will be always be set. ares was
erroneously using only the bottom 3 immediate bits in this case (so 15 would
behave the same as 7).

With this fix, sprites now render correctly in the Neo Geo Pocket game Pocket
Tennis.

This bug was discovered by comparing against an instruction trace from MAME.
This change is the functional equivalent of commit
57e42cda31430c75566403c5f27ec290dc123da3 in the MAME git repo.

Before:
before
After:
after

### DIFF
--- a/ares/component/processor/tlcs900h/instruction.cpp
+++ b/ares/component/processor/tlcs900h/instruction.cpp
@@ -829,9 +829,9 @@ auto TLCS900H::instructionRegister(R register) -> void {
   }
   case 0x33: {  //BIT #,r
     if constexpr(Long) return undefined();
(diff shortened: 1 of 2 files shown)


LukeUsher:
Thank you for your contribution.

It looks good to me but I have asked for someone with better knowledge of this to give it a once-over before we merge, just to be on the safe side.

PR #3, opened 2021-07-06
Fix building mia on Linux.

Screwtapello:
Although mia doesn't use threads itself, it includes `ares/ares.hpp` and ares *does* use threads, so we need to build with `-pthread` in order to link successfully.


LukeUsher:
Thanks

PR #7, opened 2021-07-10
Fixes for "Fast Boot" option for Neo Geo Pocket (Color)

invertego:
Perform more initialization to match the side effects of booting from
the BIOS.

1) Initialize XIX register. Puyo Pop World relies on this (likely a game
   bug).
2) Initialize interrupt vector table. SNK Gals' Fighters and the Pocket
   Casino Series games rely on this.
3) Initialize interrupt request levels used by the VECT_INTLVSET system
   call. The Pocket Casino Series games rely on this.

### DIFF
--- a/ares/ngp/cpu/cpu.cpp
+++ b/ares/ngp/cpu/cpu.cpp
@@ -338,6 +338,9 @@ auto CPU::fastBoot() -> void {
   address.byte(2) = cartridge.read(0, 0x1e);
   store(PC, address);
   store(XSP, 0x6c00);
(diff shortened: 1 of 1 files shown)


invertego:
This fixes all of the fast boot bugs [documented here](https://www.retrowide.com/stuff/ares_compat.htm).

This change keeps with the original design of the fast boot option, faithfully replicating the side effects of the BIOS boot procedure. That said, I think it would be considerably simpler to always let the BIOS handle the boot process and for the fast boot option to instead apply a code patch to skip the boot animation. I haven't tried this myself; it's just an idea for a future improvement.

PR #12, opened 2021-07-14
ngp: re-enable saving ram to disk on unload

invertego:
This code was commented out in commit
4e8232118288a0db8cb266bb07f278ee4184adbd which otherwise only affected
TLCS900/H timing. I believe this was left in by mistake. Uncommenting
this code restores the ability to persist BIOS settings across play
sessions.


LukeUsher:
Thanks!

PR #13, opened 2021-07-14
ngp: copy SNK logo from BIOS during fast boot

invertego:
The game Metal Slug - 2nd Mission compares a region of character memory
to a copy of the SNK logo in the game rom. Normally this memory is
initialized for the boot animation played by the BIOS. If the logo isn't
an exact match, the game starts but you are unable to fire any weapons.

Conceptually, this is the reverse of what the Game Boy does on boot.
There, the boot rom validates a Nintendo logo on the cartridge.

Since we already have the BIOS loaded, we can copy the logo from there
during a fast boot, though some unpacking is required.

### DIFF
--- a/ares/ngp/cpu/cpu.cpp
+++ b/ares/ngp/cpu/cpu.cpp
@@ -404,6 +404,22 @@ auto CPU::fastBoot() -> void {
     kge.write(address, data[address & 15]);
   }
 
+  //SNK logo: Metal Slug - 2nd Mission compares this memory to a copy in the rom
+  //and disables shooting controls if it doesn't match.
+  u32 logo = Model::NeoGeoPocketColor() ? 0x8eaf : 0x8123;
+  for(u32 address = 0xa1c0; address < 0xa200; address++) {
+      n8 packed = system.bios.read(logo + ((address >> 1) & 0x1f));
+      if (address & 1) packed >>= 4;
+
+      n8 unpacked;
+      unpacked.bit(7) = unpacked.bit(6) = packed.bit(3);
+      unpacked.bit(5) = unpacked.bit(4) = packed.bit(2);
+      unpacked.bit(3) = unpacked.bit(2) = packed.bit(1);
+      unpacked.bit(1) = unpacked.bit(0) = packed.bit(0);
+
+      kge.write(address, unpacked);
+  }
+
   //interrupt request levels used by syscall VECT_INTLVSET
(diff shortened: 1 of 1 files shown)


invertego:
Thanks to @FitzRoyX for discovering this issue and noting it [here](https://www.retrowide.com/stuff/ares_compat.htm)!

This change further reinforces my feeling that the fast boot path would benefit from a redesign. I looked at the Game Boy core, and it implements fast boot as a boot rom patch, so switching the NGP core to do the same would actually be more consistent with other cores in addition to being more robust. I found a one byte patch that will skip the boot animation in the BIOS (write 0x0e to address 0x4618 or address 0x530c in the color BIOS). However, this exposes fast boot to some of the problems currently exclusive to the non-fast boot path, mostly inherited from MAME by following the precedent it set of booting from 0xff1800 instead of the real hardware reset vector. The whole boot process could use a rethink.

FitzRoyX:
Fast boot has been a real troublemaker. What exactly does the feature offer besides not having to wait through BIOS splash? Fast forward skips it, savestates skip it. It's not like the BIOS file is undumped or hard to find. Hopefully this is all worth your talents, I'd hate for devs to keep chasing bugs that aren't really bugs.

invertego:
The current implementation of fast boot is radically different from the non-fast boot code path. Fast boot skips *all* of the normal BIOS initialization code but attempts to replicate its side effects (of which there are many) in emulator code. What I'd like to do is get rid of almost all of this code and just let the BIOS handle the boot process all the time, and when the fast boot option is selected just patch the BIOS to skip the boot animation. To the end user, the end result should be virtually the same.

The complicating factor is that the current fast boot mode is hiding some other emulation bugs, and those will need to be dealt with first. The most notable side effect of these issues is that in non-fast boot mode, resetting certain games (like Metal Slug - 2nd Mission) will always trigger the BIOS setup. MAME suffers this same issue, but it's even worse - just playing games like Metal Slug will permanently mess up your nvram save and you'll start getting the BIOS setup before every game you launch. ares fares better because it makes some attempt to clean things up even in the non-fast boot case.

LukeUsher:
I'll merge this for now, but just a heads up that I'm considering removing "fast boot" functionality as a whole. 

It needs more discussion with other developers, but it feels like a hack that we shouldn't need to keep working around, especially when it doesn't actually remove the requirement for a bios in most cases.  Boot logos are part of the experience!

PR #15, opened 2021-07-19
ngp: allow BIOS to handle the boot process

invertego:
Neo Geo Pocket is an "always on" system. On first boot, it quickly
enters a low power state via a halt instruction. From there, the user
must press the power button to wake the sytem. At this point, the BIOS
setup is shown, requiring the user to pick a preferred language among
other choices. When power is pressed again, the system returns to a low
power state. This state is maintained by an internal battery should the
external batteries discharge or be removed. The hardware reset entry
point will not execute again until the internal battery fails.

The current design of ares appears to be motivated by a few goals:
- Avoid requiring the user to press the emulated power button
- Avoid showing the BIOS setup more times than necessary
- Enable "fast boot" mode to skip the boot animation

The required power button press is avoided by skipping the first boot
code entirely in CPU::power(). Emulation never begins at the hardware
reset vector, but rather at 0xFF1800, following MAME's lead. This entry
point appears to have been for development purposes only and is not
referenced by the BIOS itself. Crucially, it does not run some
bookkeeping logic that only occurs when entering a low power state
either after a cold boot, in response to a power button press, or after
a software requested shutdown. Skipping this logic can cause the BIOS
setup to be shown on next boot.

The BIOS setup is avoided by simulating the bookkeping done by the BIOS
in the emulator itself. This logic is located in CPU::load(), meaning
that it does not run between system resets. Therefore, resetting a game
can still cause the BIOS setup to be shown. The BIOS setup is also
skipped by fast boot mode.

Fast boot mode bypasses the BIOS entirely on boot, jumping straight to
the cartridge entry point. This means all of the many important side
effects of the boot process must be replicated by CPU::fastBoot().
Currently, it also gives the user no way to select Japanese as a
preferred language.

The existing solutions to these problems fall short of providing a
consistent user experience, tightly couple ares to the internals of the
BIOS rom, and (particularly with fast boot) have been a source of many
bugs. This change removes the existing solutions in favor of ones that
let the BIOS handle the entirety of the boot process.

The new code in this change does the following:
- Begin execution at the hardware reset vector on first boot
- Begin execution at the VECT_SHUTDOWN syscall on subsequent boots
- Simulate a power button press when entering a low power state
- Patch the BIOS to skip the boot animation in fast boot mode

Starting at the hardware reset vector ensures the BIOS setup will be
shown on first boot, affording the user the opportunity to pick a
preferred language. Subsequent boots will begin at the VECT_SHUTDOWN
system call. It may seem counterintuitive, but this allows the BIOS to
perform cleanup that would normally only occur when the user shuts off
the device before switching cartridges. In both of these cases, we can
avoid requiring the user to press the emulated power button by
simulating a button press on their behalf. It's a low tech solution to
be sure, but a reliable one.

Finally, the implementation of fast boot mode is replaced with a single
byte patch to the BIOS to immediately return from the routine that plays
the boot animation.

### DIFF
--- a/ares/ngp/cpu/cpu.cpp
+++ b/ares/ngp/cpu/cpu.cpp
@@ -20,31 +20,6 @@ auto CPU::load(Node::Object parent) -> void {
 
   if(auto fp = system.pak->read("cpu.ram")) {
     ram.load(fp);
-
-    //hack: the BIOS checks (0x6c14) to determine if the setup menu needs to be executed.
-    //for unknown reasons, the BIOS from (0xffff00) only writes 0x00dc.
-    //it is not updated when language/color choices are made.
-    //to prevent the setup from being run every boot, manually calculate and set this value.
-    n16 data;
-    data += ram[0x2f87];  //language: Japanese, English
-    for(u32 address = 0x2c25; address <= 0x2c2b; address++) data += ram[address];  //always 0x00dc
-    if(Model::NeoGeoPocketColor()) {
-      data += ram[0x2f94];  //K1GE color mode: Black & White, Blue, Green, Red, Classic
-    }
-    ram[0x2c14] = data.byte(0);
-    ram[0x2c15] = data.byte(1);
-
-    //signature check
-    ram[0x2e96] = 'N';  //'N'eo Geo
-    ram[0x2e95] = 'P';  //'P'ocket
-
-    //this byte seems to indicate system state (0x00 = BIOS UI, 0x10 = setup, 0x40 = game playing)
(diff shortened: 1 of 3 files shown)


invertego:
Investigating a regression in the boot animation audio with this change. May want to hold off on merging until that's sorted.

Edit: Never mind, looks like more instances of the general problem of audio going silent. These issues occur under the same circumstances (switching games at certain points in time) on v120 as well, so I believe my change is in the clear.

LukeUsher (approved):
Excellent job, thanks for this!

PR #19, opened 2021-07-25
lucia: support digital presses from analog inputs

### LINKED ISSUE #16: Analog deadzone/sensitivity is busted
I'm using an XBOX one gamepad and assigned the right analog to the c-buttons.

version 121.a has the right analog deadzone disabled so it presses the buttons by itself, it's impossible to control.

invertego:
lucia maps the N64 C-buttons to the right thumbstick of the virtual pad.
These are analog inputs on the virtual pad, so sampled analog values are
not converted to digital. This means any nonzero value is treated as
button press, which is not ideal when mapped to an analog input device.

This change introduces a helper to convert raw input values to a binary
"pressed" state at time of use, currently only defined for digital and
analog input mappings - for relative and absolute inputs this concept
would not make much sense.

This fixes #16.

### DIFF
--- a/lucia/emulator/emulator.cpp
+++ b/lucia/emulator/emulator.cpp
@@ -190,8 +190,8 @@ auto Emulator::input(ares::Node::Input::Input input) -> void {
       for(auto& inputNode : inputDevice.inputs) {
         if(inputNode.name != input->name()) continue;
         if(auto button = input->cast<ares::Node::Input::Button>()) {
-          auto value = inputNode.mapping->value();
-          return button->setValue(value);
+          auto pressed = inputNode.mapping->pressed();
+          return button->setValue(pressed);
         }
         if(auto axis = input->cast<ares::Node::Input::Axis>()) {
           auto value = inputNode.mapping->value();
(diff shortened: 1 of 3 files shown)


remutro:
I'll point out that this is a regression from supporting per core controller configurations introduced in WIP 120.15 ( commit: 532aafe074388b1a368334f7c25b085f73c34936 ). 
This line (150) used to exist in lucia/emulator/nintendo-64.cpp which allowed this to work properly previously:

if(name.beginsWith("C-")) value = abs(value) > +16384;

It should have been carried over and inserted after line 193 in lucia/input/input.cpp (with minor modification) and it would have addressed this issue as well:

```
         if(auto button = input->cast<ares::Node::Input::Button>()) {
          auto value = inputNode.mapping->value();
          if(input->name().beginsWith("C-")) value = abs(value) > +16384;       <==missing line not carried over
          return button->setValue(value);
        }
```

This pull request looks to do the same thing, although there is a lot more code introduced.

invertego:
Good catch. I'd argue it's cleaner to differentiate by type than by name, but the end result is the same so long as the C-buttons are the only inputs that fall into this category.

PR #21, opened 2021-07-26
z80: correct the effect of IE on next instruction

invertego:
The implementation of IE in the Z80 processor was incorrectly causing
nonmaskable interrupts raised before the next instruction to be dropped.
A correct implementation should do two things:
- set IFF1 and IFF2 to 1
- disable maskable interrupts until after the following instruction

This fixes the issue of audio dropping out of Neo Geo Pocket games
seemingly at random.

References:
Z80 CPU User Manual (http://www.zilog.com/docs/z80/UM0080.pdf)
Page 6: "NMI is always recognized at the end of the current instruction,
independent of the status of the interrupt enable flip-flop..."
Page 183: "During the execution of this instruction [EI] and the
following instruction, maskable interrupts are disabled."

### DIFF
--- a/ares/component/processor/z80/instruction.cpp
+++ b/ares/component/processor/z80/instruction.cpp
@@ -1,11 +1,6 @@
 auto Z80::instruction() -> void {
+  EI = 0;
   P = 0;
-
-  if(EI) {
-    EI = 0;
-    IFF1 = 1;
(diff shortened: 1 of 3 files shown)


LukeUsher:
Excellent find!

Crazy how we're still finding Z80 bugs at this stage.

Issue #28, opened 2021-08-03
macOS Mojave support


spotanjo3:
What about people who prefer to stay with Mojave because of 32bit and 64bit ? 

And

> The entire Neo Geo Pocket/Color library is now thought to be playable, with minor graphical issues in four titles.

Which four titles ? 

By the way, keep up the good work!

spotanjo3:
Hello?

LukeUsher:
Did you test on Mojave? Your issue provides no useful information for us to work on.

spotanjo3:
> Did you test on Mojave? Your issue provides no useful information for us to work on.

Hello, LukeUsher. Thanks for answering my question.

Yes, because If I did and it works and I wouldn't reported it here. Here:

<img width="412" alt="Screen Shot 2021-08-05 at 12 12 16 PM" src="https://user-images.githubusercontent.com/16481778/128384302-00444bd7-9991-4d5b-84f6-3290d92e55c4.png">

<img width="421" alt="Screen Shot 2021-08-05 at 12 12 32 PM" src="https://user-images.githubusercontent.com/16481778/128384347-2effaade-2f50-4744-9eb4-672f135972b5.png">

It requires macOS Catalina (version 10.15). Some of us prefer to stay with Mojave due to 32bit and 64bit. Catalina removed 32bit permanently.

merryhime:
Please note that the codebase has not been tested on earlier versions of the macOS operating system, e.g. pre-github higan v110 was compiled with a deployment target of macOS 10.15.

merryhime:
Closed by #37. Please see latest nightly: https://github.com/higan-emu/ares/releases/tag/nightly

spotanjo3:
Thank you! Now, it's working for Mojave. Good work!

PR #37, opened 2021-08-05
nall: Add macOS deployment target flags

merryhime:
I do not have a macOS 10.14 system to test if it actually runs on this version, however it does compile with the appropriately versioned headers and run on more recent macos versions.

Related to #28.


LukeUsher (approved):
Aweaome, thanks!

PR #44, opened 2021-08-07
Gamepad updates: creating more inputs and updating to a more consistent layout

Shideravan:
This is part 1 of the updates I want to do and we're discussing them recently. We'll certainly discuss more this topic and I'm glad to help to improve the input system.
But I'm very cautious in my first code commit here.
In this first part, I'm just adding the missing input that is/will be needed for the controllers in systems already supported in Ares right now (but I'm also creating as a commented code right now the skeleton for the Sega Saturn control pad and Dual Shock controller for Playstation).
I'm disassociating the default layout from the one of XBox 360 + C/Z/L1/R2/whatever, as we don't know what controller the user is using (or even if he's only using the keyboard) with buttons such XYAB or even Z and C (that aren't even in the XBox 360 controller) and even if the user is using a controller we don't are sure if the letter position is the same across all controller models. But XBox 360 users don't be afraid: I'm thinking of implementing a button to auto-associate the XInput compatible controller with a default layout, to be even easier for the emulator can be to just "get and play" as following the goal to be the most user-friendly as possible.
EDIT: Now this layout is used internally only. As requested, the input system for the user sees doesn't contain a "global configurator that messes with everything" anymore.

~Also, is noteworthy to see that even the layout being thought as "universal as possible" there is a detailed description of each button name meaning, as a tooltip. It reads:~

EDIT: removed.

~It is also worth mentioning that I tested it extensively and because of that I also decided to do a more conservative commit for now. And that's why Shoulder Buttons are still called "Shoulder-L1" instead of "Shoulder Left (Top)" as I would prefer, to say in truth (but would require a few more changes to how these strings are handled in this part of the code to not break anything).~

EDIT: This part doesn't exist anymore. Was changed for an, even more, user-friendly approach.
The "virtual gamepad" can now be fully transparent for the user.
The tooltip now reads:

```
The buttons you set for a system will be translated for another.
eg.setting BAYX Super Famicom buttons will be translated respectively as Cross, Circle, Square and Triangle for Dual Shock controller.
```

Thanks everyone for the feedback.


FitzRoyX:
I understand wanting to remove "C" and "Z" because modern pads don't have that. But I don't see what's so confusing about picking a modern layout of some kind, whether it's MS or Nintendo. Everyone knows positionally that A is a face button and "L" is a shoulder button, but I have no idea what "B1" and "T1" and "asterisk" mean without reading your glossary. Even after reading it, I'm confused at your considering SNES X and Y "top row" buttons when they are in a diamond formation.

If you were going to use generic names, you would want something like:
topfacebutton
leftfacebutton
rightfacebutton
bottomfacebutton

But these are much longer strings than a , b , x, y, lb, lt, rb, rt and thus annoying to work with from a coding standpoint. The GUI is not a hard problem to overcome, it should look something like this:

cfg1

I'm missing the thumbsticks and triggers, and I didn't know how to give the companies their own columns to make it more readable, but you get the point.

Shideravan:
> I'm confused at your considering SNES X and Y "top row" buttons when they are in a diamond formation.

SNES controller even has two light gray separations for each row and in the American version, the top row and bottom row have even different colors and shapes.
SNES Controller

> The GUI is not a hard problem to overcome, it should look something like this:
> cfg1

There are no buttons such L1/R1 in any of the XBox controllers. Also, It's better to don't use "." here not only because it's much less clean but also as it may mess with the `settings.bml`.

> I'm missing the thumbsticks and triggers, and I didn't know how to give the companies their own columns to make it more readable, but you get the point.

I already considered this approach, but it may cause even more confusion as none of them have the third face row button and none of them have a keypad. And I don't think having a different approach for each button helps to. It's better to be consistent, as it avoids unnecessary confusion.
Also, this is the "default" pad. Users can change the buttons in the specific system they want to emulate anytime.

>I'm missing the thumbsticks and triggers.

Both are in this commit. Thumbsticks are called LAnalog-Thumb and RAnalog-Thumb and the triggers are called TriggerLeft and TriggerRight (EDIT: I changed the name to Trigger-Left and Trigger-Right for more consistency with the rest of the names). The Thumbsticks weren't even created before, but it's needed for proper Dual Shock support.
image

AbreuEfechou:
Nice!

FitzRoyX:
> SNES controller even has two light gray separations for each row and in the American version, the top row and bottom row have even different colors and shapes.

The PS1 controller doesn't have that grouping coloration, nor does the N64. The only systems with clear "rows" are old NEC and Sega pads. This part of the code is trying to describe how an old pad translates to a modern pad, and fighting/Saturn pads just aren't a current layout. Sega XYZ should be getting automapped to the shoulders, they are secondary buttons that didn't even exist on the stock Genesis.

> There are no buttons such L1/R1 in any of the XBox controllers. Also, It's better to don't use "." here not only because it's much less clean but also as it may mess with the `settings.bml`.

Yeah, it's LB and RB, I hastily put that together. BML parser does not like slashes and won't work with them, that's why I had to use periods.

> I already considered this approach, but it may cause even more confusion as none of them have the third face row button and none of them have a keypad. And I don't think having a different approach for each button helps to. It's better to be consistent, as it avoids unnecessary confusion.

The coleco is not going to be used as a global pad. It really doesn't matter what random buttons its numpad gets automapped to. It should probably just be hardcoded to the keyboard since that's basically what's going to happen if Ares ever supports "computer" systems like Amiga and FM Towns.

AbreuEfechou:
> The coleco is not going to be used as a global pad. It 
really doesn't matter what random buttons its numpad gets automapped to. It should probably just be hardcoded to the keyboard since that's basically what's going to happen if Ares ever supports "computer" systems like Amiga and FM Towns.

As an user I like being able to map the numpad or anything else where I want to, because I like the freedom and also because if my keyboard or usual controller is being a little elf I can find alternative ways to play or if my right hand is hurting I can use a setup that is comfortable for my left one.
I liked this commit both because things aren't hard coded in it and the naming system, the way it was before was too confusing. 
There's no sense to not change it. Colecovision's action buttons inputs was mapped as start and select before, there's no sense to not change it.
Also, @Shideravan already said that later he will work at in a auto mapper for xinput, so no need to keep pestering about things that help people with unconventional setups or that like colecovision.

AbreuEfechou:
Also you (@FitzRoyX ) can always make your own pull request if you don't.like one so much

Shideravan:
> > SNES controller even has two light gray separations for each row and in the American version, the top row and bottom row have even different colors and shapes.
> 
> The PS1 controller doesn't have that grouping coloration, nor does the N64.

N64 even has different colors for A and B (C buttons are mapped as the right joystick and it wasn't even changed yet)
image
Also, when someone with a XBox 360 will play Playstation 1, they'll map their buttons for each input they want and it'll translate to SNES very fine.
This complaining makes no sense.

> > This part of the code is trying to describe how an old pad translates to a modern pad, and fighting/Saturn pads just aren't a current layout. Sega XYZ should be getting automapped to the shoulders, they are secondary buttons that didn't even exist on the stock Genesis.

Say this to my current arcade controller. Your complaining makes it seem like the XBox 360 is the only controller in the universe. That's very far from reality.
image


> > I already considered this approach, but it may cause even more confusion as none of them have the third face row button and none of them have a keypad. And I don't think having a different approach for each button helps to. It's better to be consistent, as it avoids unnecessary confusion.
> 
> The coleco is not going to be used as a global pad. It really doesn't matter what random buttons its numpad gets automapped to. It should probably just be hardcoded to the keyboard since that's basically what's going to happen if Ares ever supports "computer" systems like Amiga and FM Towns.

Is not going just you don't wants to? Also, don't make sense with something like this to be hardcoded. Hard coding is not a viable solution, same with auto mapping sega xyz buttons. It becomes too rigid and it needs malleability as people uses in most various setups - some people even plays with the keyboard itself.

Shideravan:
> Also you (@FitzRoyX ) can always make your own pull request if you don't.like one so much

Yes, it seems that he just likes to complain about everything, but doesn't really want to help.
"The BIOS setup is 'tedious' ". (yes, unnecessarily rude like that)
"I don't like the improvement in the setup button because my 360 pad doesn't seem the center of the universe anymore" (even I didn't change a bit the way as a XBox 360 controller is mapped, just the button names in the default controller).
Man, you have to understand that this emulator is under construction (as the most open-source emulator project in history). And come just to complain about the other people's work just because "after me, the deluge" doesn't help. Have you ever thought every time someone is going to fix something they need to spend more energy arguing about the commit than coding? 
This discourages those who come willing to help with the project.

gm112:
Assuming all control pads are arranged in a top/bottom row setup isn't accurate, either. The Xbox pad and Playstation controllers are both arranged in a + shape. Which breaks the convention these changes are presuming to propose. So, as an outside commentator, I don't really see what benefit these changes make other than making another assumption in place of an existing one. There is likely a better solution that can be worked out with cooperation. Just my two cents.

Shideravan:
> Assuming all control pads are arranged in a top/bottom row setup isn't accurate, either. The Xbox pad and Playstation controllers are both arranged in a + shape. Which breaks the convention these changes are presuming to propose. So, as an outside commentator, I don't really see what benefit these changes make other than making another assumption in place of an existing one. There is likely a better solution that can be worked out with cooperation. Just my two cents.

Then I want to see some suggestions about a better way to naming them. 
Face Button 1-6?
If the problem is just with the default names, it may be easy to change.

AbreuEfechou:
> > Assuming all control pads are arranged in a top/bottom row setup isn't accurate, either. The Xbox pad and Playstation controllers are both arranged in a + shape. Which breaks the convention these changes are presuming to propose. So, as an outside commentator, I don't really see what benefit these changes make other than making another assumption in place of an existing one. There is likely a better solution that can be worked out with cooperation. Just my two cents.
> 
> Then I want to see some suggestions about a better way to naming them.
> Face Button 1-6?
> It's easy to change.

I really don't bother how will they be called, as long when configuring a Genesis pad doesn't mess with N64. Or when change the Colecovision numpad doesn't mess all the other system buttons. That's why I like this commit so much.
For me, if it's called foo1, foo2 or moon, sun, saturn, mercury... as long they're consistent (what it doesn't is in version 1.22), it iss fine for me and I'm sure most the users too.

Shideravan:
I'm just using right now a naming convention which does make the less deep configuration (and further testing) needed.
Button-1,2 and 3: are the bottom row of the face buttons. 1-3 goes from left to right
Button-4,5 and 6: are the top row of the face buttons. 4-6 goes from left to right.
What do you think? (if somebody doesn't like, please give us a better solution for the naming in the Virtual Pad).

Shideravan:
> I don't really see what benefit these changes make other than making another assumption in place of an existing one. 

Please, I just ask for you to read all the commit changes before say something so harsh like that next time.
This commit isn't just about the naming conventions.
It's about configuring something and don't more mess with the rest.
I'm even starting to implement the Dual Shock pad, which isn't just about "naming convention".
It's changing the Colecovision from using Start and Select (or Start and Back in XBox 360) as action buttons.
It's about being more consistent across the systems.
How we do name the default Virtual Pad is just a minor thing in the whole commit (EDIT: That even is completely invisible for the final user with the most recent updates).
That said, I changed the face buttons' names, as this seems to be so controversial to something more generic and neutral.
Button 1-6. No top row, no bottom row. EDIT: and now it's even fully transparent for the user.
Also, some users gave me nice feedback and ideas on how to avoid any misunderstanding in this part and further makes this more intuitive for everyone.

AbreuEfechou (approved):

AbreuEfechou (approved):

AbreuEfechou (approved):

Shideravan:
Further improving and giving more easiness for configuration. The concept of "virtual gamepad" is now transparent to the user.
The users don't need to worry anymore about the meaning of the inputs in the global configuration and if they configure something it is automatically reflected for the equivalent layouts. 
For example: if the user configures the Playstation gamepad it'll reflect in the SNES gamepad. Cross, Circle, Square, and Triangle buttons will be translated respectively to B, A, Y and X. This is clearly explained in the new tooltip info.
Thanks for the feedback everyone.

ghost:
I'm sorry for the inconvenience, I just want to say a suggestion so that it can be thought about in the future, I don't want to sound rushed or toxic.
My suggestion: implementation of the possibility to create and save profiles of controls. 👍 Thank you for all.

Shideravan:
I need to change some things and publish later in another PR

PR #49, opened 2021-08-11
tms9918: implement multicolor mode

invertego:
This mode is used by the ColecoVision games Cabbage Patch Kids Picture
Show and Smurf Paint 'n' Play Workshop.


LukeUsher (approved):
You’re a machine!

PR #54, opened 2021-08-16
cv, msx, sg: init VRAM to zero on load

invertego:
Initialize VRAM to zero on load for systems that use the TMS9918/V9938.
On real hardware this memory will be in a partially indeterminate state,
but as far as approximations go, 0x00 is much less problematic than
0xff. A sprite attribute table filled with 0xff is interpreted to have
valid entries for sprites in the top right corner of the screen. If the
pattern data is also filled with 0xff, a white square will be drawn in
the corner. On the other hand, a sprite entry of all 0x00 bytes will be
fully transparent (due to the color byte being zero) and therefore
invisible.

This is most noticeable in games that never get around to fully
initializing VRAM, such as the ColecoVision game Boulder Dash, which has
a permanent square in the corner.

SG-1000 games only seem to exhibit this problem briefly after booting,
and only for certain games, as they eventually get around to
initializing VRAM. Some homebrew is more seriously affected.

MSX compatibility is in too poor a state to test comprehensively, but a
survey of other emulators shows initializing VRAM to zero is the norm.

This change also allows VRAM to carry over across soft resets. This
should be closer to hardware behavior, though in reality it is possible
for some bits to be perturbed.

### DIFF
--- a/ares/cv/vdp/vdp.cpp
+++ b/ares/cv/vdp/vdp.cpp
@@ -7,6 +7,8 @@ VDP vdp;
 #include "serialization.cpp"
 
 auto VDP::load(Node::Object parent) -> void {
+  vram.allocate(16_KiB, 0x00);
+
   node = parent->append<Node::Object>("VDP");
 
   screen = node->append<Node::Video::Screen>("Screen", 256, 192);
@@ -45,7 +47,6 @@ auto VDP::power() -> void {
   TMS9918::power();
   Thread::create(system.colorburst() * 2, [&] { main(); });
   screen->power();
-  vram.allocate(0x4000);
 }
 
 }
--- a/ares/msx/vdp/vdp.cpp
+++ b/ares/msx/vdp/vdp.cpp
@@ -15,6 +15,7 @@ auto VDP::load(Node::Object parent) -> void {
     screen->setSize(256, 192);
     screen->setScale(1.0, 1.0);
     screen->setAspect(1.0, 1.0);
+    TMS9918::vram.allocate(16_KiB, 0x00);
     TMS9918::load(screen);
   }
 
@@ -24,6 +25,9 @@ auto VDP::load(Node::Object parent) -> void {
     screen->setSize(512, 424);
     screen->setScale(0.5, 0.5);
     screen->setAspect(1.0, 1.0);
+    V9938::vram.allocate(128_KiB, 0x00);
+    V9938::xram.allocate(64_KiB, 0x00);
+    V9938::pram.allocate(16);
     V9938::load(screen);
   }
 }
@@ -58,15 +62,11 @@ auto VDP::frame() -> void {
 
 auto VDP::power() -> void {
   if(Model::MSX()) {
-    TMS9918::vram.allocate(16_KiB);
     TMS9918::power();
     Thread::create(system.colorburst() * 2, [&] { TMS9918::main(); });
   }
 
   if(Model::MSX2()) {
-    V9938::vram.allocate(128_KiB);
-    V9938::xram.allocate(64_KiB);
-    V9938::pram.allocate(16);
     V9938::power();
     Thread::create(system.colorburst() * 2, [&] { V9938::main(); });
   }
--- a/ares/sg/vdp/vdp.cpp
+++ b/ares/sg/vdp/vdp.cpp
@@ -7,6 +7,8 @@ VDP vdp;
 #include "serialization.cpp"
 
 auto VDP::load(Node::Object parent) -> void {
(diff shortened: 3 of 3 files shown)


Screwtapello:
Shouldn't RAM be initialised randomly, like in the SFC core (and other cores)?

https://github.com/higan-emu/ares/blob/master/ares/sfc/cpu/cpu.cpp#L92

LukeUsher:
> Shouldn't RAM be initialised randomly, like in the SFC core (and other cores)?
> 
> https://github.com/higan-emu/ares/blob/master/ares/sfc/cpu/cpu.cpp#L92

For 100% accuracy, probably; but it does concern me that this does cause issues in games and may impact user experience.

invertego:
@Screwtapello I did consider random initialization, but the available evidence suggests the initial VRAM contents follow a highly regular pattern, at least for Sega consoles. This is demonstrated by games that briefly draw uninitialized memory to the screen:

[Sega Galaga on SC-3000](https://www.youtube.com/watch?v=v0sTiVAoOgc)
[Sega Galaga on SMS (?)](https://www.youtube.com/watch?v=XWdNOUGW3qw)
[Safari Hunting on SC-3000](https://www.youtube.com/watch?v=68qMznTbOik)
[Safari Hunting on SMS](https://www.youtube.com/watch?v=gEgZNA4OxeY)
[N-Sub on SG-1000](https://www.youtube.com/watch?v=mp57jq2Pc7M)
[N-Sub on SMS](https://www.youtube.com/watch?v=BJWpDOpKoes)

The SG-1000/SC-3000 pattern looks roughly like alternating runs of 64 bytes of zero-ish and nonzero-ish data. The SMS looks similar but for runs of 128 bytes in length.

I didn't quite feel comfortable trying to simulate this in ares without proper test ROMs and a variety of hardware models to test on.

Screwtapello:
The patterns in the power-on state of RAM are well-known. Near spent a long time analysing SNES RAM dumps trying to find a way to reproduce things, and the "low entropy" randomisation in ares is the closest he got. The actual physical process that causes this behaviour is described in [Initial SRAM State as a Fingerprint and Source
of True Random Numbers for RFID Tags](http://web.eecs.umich.edu/~kevinfu/papers/holcomb-FERNS-RFIDSec07.pdf) (Holcomb, Burleson, and Fu, 2009), but I'll try to summarise:

- RAM is made up of transistors, which are made of silicon with carefully-constructed patterns of impurities that slightly attract or repel electrons
- when power is connected, the attraction or repulsion is small enough to not make a difference
- when power is disconnected, electrons can flow around freely and clump up wherever they want to, and the slight attraction or repulsion influences where they go
- because adding impurities to silicon is a physical process, it's not 100% even across the entire chip, so some parts of the chip more strongly attract or repel electrons than others
- because RAM transistors are laid out in a predictable pattern across the chip, bits that are physically near each other are likely to have similar levels of impurity in the silicon, and therefore likely to have similar values, but ultimately they're still random

If I had to simulate random power-on state in an emulator, I'd do something like this:

- divide the RAM size up into chips
    - for example, a computer with 512KB RAM might have four 128KB chips
- for each chip, draw a grid with one grid-square for each bit the chip can store
    - for example, 128KB is 1024 Kbit, so it would make a 1024×1024 grid
- use the [diamond-square algorithm](https://en.wikipedia.org/wiki/Diamond-square_algorithm) on the grid to simulate the uneven distribution of impurities in the silicon, giving each bit a value between 0 and 1
- at power-on, look at the generated  value of each bit:
    - value <= 0.25: set bit to 0
    - 0.25 < value < 0.75: set bit randomly
    - value >= 0.75: set bit to 1

> The SG-1000/SC-3000 pattern looks roughly like alternating runs of 64 bytes of zero-ish and nonzero-ish data. The SMS looks similar but for runs of 128 bytes in length.

It's easy to recreate these kinds of patterns by changing how you convert a memory address to a grid location. If you look at an image of a [real RAM chip](https://wccftech.com/images/articles/History-of-RAM/1-Megabit-Chip.jpg) you'll see there's a lot more structure - the memory space is divided into lanes, and each lane has blocks down each side, and each block probably has smaller structures within it. If you divide the grid into 8 columns and take one bit from each column, you'll get a different pattern than if you divide the grid into quadrants and take two bits from each one, or some more complex scheme.

You can get as complicated as you like, but since each physical chip's power-on pattern is unique, there's not much point to doing so - it's random enough that games generally don't depend on any particular pattern (except maybe 'not all zeros' or 'not all ones'), so the only benefit is aesthetic, giving you "realistic" looking glitches at power-on.

PR #66, opened 2021-08-19
Fix halt bug

Meepster99:
This fixes the halt bug in gameboy emulators
A rom to test it can be found [here](https://github.com/retrio/gb-test-roms)

### DIFF
--- a/ares/component/processor/sm83/instructions.cpp
+++ b/ares/component/processor/sm83/instructions.cpp
@@ -98,9 +98,9 @@ auto SM83::instructionDAA() -> void {
   if(!NF) {
     if(HF || (n4)A > 0x09) a += 0x06;
     if(CF || (n8)A > 0x99) {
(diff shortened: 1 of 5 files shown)


Screwtapello:
If `haltBugTrigger()` is only called in one place, why is it in a separate function?

Meepster99:
This is due to my lack of experience in large scale c++ projects, but I couldn't find a way to access the "status" variable from inside the SM83 namespace. 
Any way of cleaning it up would be appreciated.

LukeUsher:
Could you please rebase this PR and then force-push? I ask because right now it duplicates the commits that were already merged from the previous PR

Meepster99:
Hi! I think I rebased and force pushed the PR, but I am very new to github and am not sure.
Does this look good?

LukeUsher (approved):

PR #67, opened 2021-08-19
gb: only raise stat interrupts if display enabled

invertego:
This emulation bug, combined with the intentional emulation of the
hardware bug that causes spurious stat interrupts, was causing stat
interrupts to be raised when the display was disabled.

This stops the GB game Cool Hand from freezing after selecting cribbage.


LukeUsher:
Thanks!

PR #71, opened 2021-08-20
Adding .unf file support

### LINKED ISSUE #70: [NES] Support for .unf/.unif file extensions
**unf** stands for "Universal NES image file format".
There are some NES homebrew demos distributed only in this format.
https://fileinfo.com/extension/unf

Shideravan:
Fix https://github.com/higan-emu/ares/issues/70


LukeUsher (approved):
Thanks; for reference this doesn't change any other code because unf and unif are identical; this just allows ares to recognise 'UNIF' files with the .unf extension.

Issue #60, opened 2021-08-16
[GBC] Graphical errors in multiple games


Shideravan:
Shantae: Game runs and sound playback is ok. But most of the cutscene screens are black. In the game itself, just background, but invisible character sprites
![image](https://user-images.githubusercontent.com/12163923/129640390-b6ac7e6b-77b3-4594-92e3-9fbabade895a.png)
![image](https://user-images.githubusercontent.com/12163923/129642243-4556c6ca-0edf-4a3e-9b04-e2e32b975faf.png)
Donkey Kong Country: Glitched graphics in most screens
![image](https://user-images.githubusercontent.com/12163923/129640463-5df25f15-dc32-4687-8f2a-b7b22e44cf41.png)
![image](https://user-images.githubusercontent.com/12163923/129640471-b0609d13-dad5-49fc-a1ea-7e9b6a4d157c.png)
![image](https://user-images.githubusercontent.com/12163923/129640481-12e03084-fa0f-447a-a931-8d066b370a76.png)
Alice in Wonderland: Messed graphics in cutscenes
![image](https://user-images.githubusercontent.com/12163923/129640717-0ede6c5f-f4f6-41cc-aa39-28453108aabd.png)
The Simpsons - Night of the living treehouse terror: The gameplay screen has no image.
![image](https://user-images.githubusercontent.com/12163923/129640849-c2a0a554-ae98-41ac-bec1-74ba3a3b9f5b.png)
Just to name a few examples.

And most Game Boy Color exclusive games have any kind of graphical errors.
I suspect that maybe this is caused by incorrect tileset indexing.

Version: cb4a79e
Windows 10 x64

Shideravan:
Alice in wonderland and Shantae has fixed with https://github.com/higan-emu/ares/commit/d50e9d2f5db0d3b7adfeed0a1331a0a479f55464
In the other games, the issue continues.
![image](https://user-images.githubusercontent.com/12163923/130329638-79347f3f-3341-4b2f-abda-9446000c47bd.png)
![image](https://user-images.githubusercontent.com/12163923/130329616-41211a46-7a12-469b-b914-3e4250ebfb12.png)

Shideravan:
All fixed with https://github.com/higan-emu/ares/pull/79
![image](https://user-images.githubusercontent.com/12163923/130352797-9b1345a4-2a43-4590-ae8f-b5af7e7ab4bb.png)
[...]

Issue #58, opened 2021-08-16
[GBC] Pocket Music: freezes in white screen just before gameplay starts


Shideravan:
Version tested: `cb4a79e`
Windows 10 x64

Shideravan:
Fixed the problem of the issue with https://github.com/higan-emu/ares/commit/c1d8f2d873ec911896f2f2da2af252eb34b98b07
(but we still have some other video problem with this game, which I'll open a new issue for it)
![image](https://user-images.githubusercontent.com/12163923/130353275-506a7f70-3ca8-40bb-a20a-ba04c83d16aa.png)

Issue #43, opened 2021-08-07
[GBC] Tomb Raider(US): Graphics corruption + crashes ingame


Shideravan:
The cutscenes are all messed up and just before the gameplay starts the game freezes in the following screen:

![image](https://user-images.githubusercontent.com/12163923/128611992-74125a0a-d0cc-4321-9ba1-6096b3088d36.png)

Shideravan:
Fixed with https://github.com/higan-emu/ares/pull/79

PR #79, opened 2021-08-22
gbc: invert HDMA5 bit 7 on read

invertego:
A refactor in v116r07 replaced the variable dmaCompleted with dmaActive
but failed to update its usage on HDMA5 register reads. Bit 7 should
read 0 if HDMA is active and 1 if completed.

This change will affect many (most?) GBC games.


invertego:
#43, #58, #60 are all fixed with this change (together with my other recent changes).

Shideravan:
Thank you for the fixes @invertego

PR #88, opened 2021-08-23
gb: bring fast boot closer to normal boot

invertego:
The previous fast boot patch skipped the entire Nintendo logo animation
and sound. Unfortunately, this left a number of registers with different
values than they would otherwise have. The European release of the game
Dr. Franken does not initialize SCY, and fast boot was leaving it set to
100 instead of zero, causing the first logo to be shifted offscreen.

It would be easy enough to amend the patch to initialize SCY to 0, but
for most games this will cause the Nintendo logo to appear for a single
frame, which doesn't look so great.

This new patch skips the first ~4 seconds of the boot animation and
disables scrolling but otherwise leaves things alone. The result is a
a stationary Nintendo logo that appears centered on the screen,
accompanied by the usual sound, for ~1.5 seconds.

### DIFF
--- a/ares/gb/system/system.cpp
+++ b/ares/gb/system/system.cpp
@@ -135,8 +135,12 @@ auto System::power(bool reset) -> void {
 
     if(fastBoot->latch()) {
       if(name == "boot.dmg-1.rom") {
-        bootROM.program(0x64, 0x18);
-        bootROM.program(0x65, 0x7a);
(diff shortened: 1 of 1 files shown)


invertego:
@LukeUsher You mentioned that you were considering removing the fast boot option, but I'd be sad to see it go. I find it very convenient when debugging and iterating on changes, which can involve booting the same game many times.

PR #89, opened 2021-08-23
gb: use stable sort for sprites

invertego:
On DMG, sprites are sorted by x value then by OAM address. As the sprite
array begins in OAM order, a stable sort on x can achieve the desired
result. Conveniently, nall already contains a stable sort implementation
that is optimized for small arrays.

This fixes ordering issues with speech bubbles in Crayon Shin-chan 3.


LukeUsher:
Nice job, thanks!

PR #96, opened 2021-08-23
Giving support for .bin and .rom formats used by PD homebrews

### LINKED ISSUE #91: [ColecoVision] Support for .rom format
AFAIK the only games that are distributed in that format are homebrews from Collectorvision Games.

Shideravan:
Fixes https://github.com/higan-emu/ares/issues/91

### DIFF
--- a/mia/medium/colecovision.cpp
+++ b/mia/medium/colecovision.cpp
@@ -1,6 +1,6 @@
 struct ColecoVision : Cartridge {
   auto name() -> string override { return "ColecoVision"; }
-  auto extensions() -> vector<string> override { return {"cv", "col", "bin, "rom"}; }
+  auto extensions() -> vector<string> override { return {"cv", "col", "bin", "rom"}; }
   auto load(string location) -> bool override;
   auto save(string location) -> bool override;
   auto analyze(vector<u8>& rom) -> string;
--- a/mia/medium/colecovision.cpp
+++ b/mia/medium/colecovision.cpp
@@ -1,6 +1,6 @@
 struct ColecoVision : Cartridge {
   auto name() -> string override { return "ColecoVision"; }
-  auto extensions() -> vector<string> override { return {"cv", "col"}; }
+  auto extensions() -> vector<string> override { return {"cv", "col", "bin, "rom"}; }
   auto load(string location) -> bool override;
   auto save(string location) -> bool override;
   auto analyze(vector<u8>& rom) -> string;


Shideravan:
Fixed missing quotes. Probably a mistake just before pushing.

Kawa-oneechan:
I wouldn't merge this if it were up to me, mainly because of the whole Mega Drive thing.

Shideravan:
> I wouldn't merge this if it were up to me, mainly because of the whole Mega Drive thing.

Please elaborate

Screwtapello:
Luke is free to make his own decisions, but for anyone interested in "what Near would have done" there's a clear precedent in the Mega Drive/Genesis core: `.BIN` is a popular and traditional extension for Mega Drive games, far more popular than .bin for ColecoVision homebrew, and neither higan nor ares have ever supported it. Politically, because Near found generic extensions annoying, and technically because each emulation core has a list of supported file extensions and they can't overlap - for example, that's why you can't play "black cartridge" Game Boy games (ones with both DMG and GBC compatibility) in DMG mode - the GBC core claims the `.gbb` extension, so the DMG core can't use it.

To summarise:

- ares probably shouldn't support generic file-extensions, even if other emulators do, because it makes one thing easy (loading this specific ROM in this specific emulator) at the expense of making everything else more complex (accidentally loading non-game `.bin` files in an emulator, loading game `.bin` files in non-emulators, or emulators for the wrong platform, etc.)
- if ares *does* add support for generic file-extensions, they should probably be added to the Mega Drive core and not the ColecoVision core

Shideravan:
Thanks for the clarification @Screwtapello 
But it's valid to say ares _does_ have support for Mega Drive .bin extension (and was Near himself who implemented).
Also, there's precedent for overlapping files in ares (eg.: CD-ROM image files).
Also, ares _does_ support for GBC games in GB driver, you just need to choose the gameboy system
image

Even GBC games without support for DMG game boy, showing the exact same message as the real hardware, as expected
image

That said, I'm closing this PR.

Screwtapello:
> But it's valid to say ares does have support for Mega Drive .bin extension in the present version.

Huh, so it does. I apologise, I should have checked.

Kawa-oneechan:
Not liking generic extensions and supporting them anyway are not mutually exclusive. If there's enough pressure to allow .bin files on the MD core, you can just say "okay, I'll do it but under protest."

Shideravan:
> Not liking generic extensions and supporting them anyway are not mutually exclusive. If there's enough pressure to allow .bin files on the MD core, you can just say "okay, I'll do it but under protest."

The .bin support for mega drive games exists from day 1 of ares release, and we don't need such protest as it was Near's own decision (and makes sense to be this way as most the mega drive games are mainly distributed in .bin format). But I'll take note from this tip for the next issue fulfillment that is against our project principles. Nice tip :)

PR #100, opened 2021-08-24
As requested: updating readme only in parts that aren't more suited for the website

Shideravan:
As most of the info from https://github.com/higan-emu/ares/pull/35 is better suited to our new website.


Shideravan:
I'll separate this

PR #106, opened 2021-08-26
gb: resume from halt whenever (IE & IF) != 0

invertego:
Normally, breaking out of the halt state is handled by CPU::raise(), but
there is an edge case:
    ; preconditions: IME = 0, IE = 1, IF.0 = 1, LCDC.7 = 0
    ei
    halt

In other words, a vblank interrupt was requested with IME disabled, then
the display was disabled (so no more vblank interrupts will follow),
then interrupts were enabled while entering the halt state. Once halted,
there is no logic in ares to service the requested vblank interrupt.

A better long term approach would be to refactor the SM83 implementation
of halt to be more like the Z80 core - eliminating the loop within the
halt instruction itself - but in the meantime this change offers an
incremental improvement.

This fix prevents Amazing Penguin from freezing immediately on load.

### DIFF
--- a/ares/gb/cpu/memory.cpp
+++ b/ares/gb/cpu/memory.cpp
@@ -7,6 +7,7 @@ auto CPU::stop() -> void {
 
 auto CPU::halt() -> void {
   idle();
+  if(status.interruptLatch) r.halt = 0;
   if(Model::SuperGameBoy()) {
     scheduler.exit(Event::Step);
   }


LukeUsher:
@invertego would you mind joining bcord at https://discord.gg/xdsQRXfuCh, this is a discord for the higan/bsnes/ares family of emulators: you’ve provided a lot of useful improvements recently, it would be good to chat.

invertego:
@LukeUsher Sure, I joined a few weeks ago but have been mostly lurking. I'm around if you want to chat.

Issue #104, opened 2021-08-26
[MD] Demons of Asteborg - Graphical errors in the opening text


Shideravan:
This is a freshly released Mega Drive/Genesis game.
The issue happens is reportedly to happens in both the Demo version and the complete version (but I just have the demo version to test it now).
You can get a free demo ROM from the developer's website.
https://www.demonsofasteborg.com/

The full game as a ROM it's also selling there too.

There's are some screens of the game in ares with corrupted graphics in most recent version of ares (v123 dev-https://github.com/higan-emu/ares/commit/2380e62c79708cc1e284fc010ed7e7cd94ef4983)
![image](https://user-images.githubusercontent.com/12163923/130889191-b4dd41d2-9174-4b7b-ad6a-47db6f006d35.png)
![image](https://user-images.githubusercontent.com/12163923/130891525-7a31f2af-9e5c-4ee8-b09d-0df5369341b7.png)
![image](https://user-images.githubusercontent.com/12163923/130889283-6429ccb8-1823-42d2-bd7f-17708a27f9d3.png)
![image](https://user-images.githubusercontent.com/12163923/130891543-51a4949a-43ab-422f-ac96-0450aeccaf6f.png)

EDIT: This is the same game tested in MAME as a reference.
[video](https://youtu.be/cfHLxtPSzsk)

Shideravan:
A similar graphical error also happens in Comix Zone
![image](https://user-images.githubusercontent.com/12163923/131066597-c06740fd-d8fb-4fab-82d3-112697b664aa.png)

Shideravan:
The same with Mega Man: The Wily Wars
![image](https://user-images.githubusercontent.com/12163923/131070584-61c8ec0f-042e-4f04-91fa-efa06a198144.png)
![image](https://user-images.githubusercontent.com/12163923/131070622-5fd74571-3d65-4e64-b4a4-f3c710df9e7a.png)
![image](https://user-images.githubusercontent.com/12163923/131070813-c8ad554d-7a1e-40da-b3f3-1246edd2f371.png)

TascoDLX:
Confirmed to be a bug in DMA fill op: first byte written is incorrect. I should be able to track it down.

Shideravan:
Fixed with https://github.com/higan-emu/ares/pull/110

PR #97, opened 2021-08-24
md: vdp dma & h-int fixes

TascoDLX:
1) Allow VDP DMA to run at full speed
2) Avoid erroneous cancellation of H-INT

Fixes graphical issues in:
- Another World (E) #74
- 3 Ninjas Kick Back (U) #76
- Virtua Racing (U)
- Burning Force (U)
- TiTAN Overdrive 2 (E)

PR #110, opened 2021-08-27
md: fix sprites and dma fills

TascoDLX:
1) Corrects sprite processing per the test cases of the Sprite Masking and Overflow Test ROM by Nemesis.
This fixes sprite rendering in Mickey Mania's Moose Chase level.

2) Adjusts the DMA fill / FIFO interaction (regression from #97).
This fixes graphical errors seen in #104 (multiple games).

### DIFF
--- a/ares/ares/ares.hpp
+++ b/ares/ares/ares.hpp
@@ -46,7 +46,7 @@ namespace ares {
 
   //incremented only when serialization format changes
   static const u32    SerializerSignature = 0x31545342;  //"BST1" (little-endian)
-  static const string SerializerVersion   = "121";
+  static const string SerializerVersion   = "123";
 
   namespace VFS {
     using Pak = shared_pointer<vfs::directory>;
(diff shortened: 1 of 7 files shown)


LukeUsher:
Nice work,

Since this changes the serialiser, it breaks save state compatibility; it might be worth incrementing the save state/serialiser version number here https://github.com/higan-emu/ares/blob/4e8015b6630903838f3c6923a564405f4a393321/ares/ares/ares.hpp#L49

TascoDLX:
Cool! I wasn't aware of that.

Also, I have a fix for the DMA fill bug I can fold in, which was artifact of the #97 DMA fix. There's at least one more affecting Overdrive 1 I still need to check out.

TascoDLX:
OK I figured out that Near was synching serializer version to the release version (using decimal point for minor releases). So, I just assumed keep doing that.

I added the fix for the DMA fill errors in #104. The Overdrive 1 bug I mentioned is actually a timing issue that I'm not ready to hack it out just yet. Accuracy is not to the point that we can be tweaking it out (e.g., RAM refresh timing is not quite right).

Issue #84, opened 2021-08-22
[ColecoVision] - Submerged turtles show a green box for a few frames in Frogger


remutro:
I know this is already known within the team, but it is probably best to create an issue to track it at this point. This is the only known corruption in ColecoVision games that I am aware of. Turtles on the bottom row will show green squares for a few frames while going down/and coming up. It doesn't seem to occur on the turtles on the top row. This occurs in Ares v123. Here is a screenshot:

![Frogger](https://user-images.githubusercontent.com/84990500/130367469-82782f6a-7615-4863-bf2a-fa2651b41e10.png)

invertego:
I took a look, and unfortunately this is a game bug. The game is inadvertently messing up its own copies to VRAM if a vblank interrupt comes in at just the wrong time. Here's an instruction trace that shows it happening:

```
1d08  ld   a,l
1d09  out  ($bf),a      ; write low address byte to port

CPU Interrupt: NMI
0066  jp   $8021
8021  push af
8022  call $1fdc
1fdc  jp   $1d57
1d57  in   a,($bf)      ; resets target of next port write
1d59  ret
8025  pop  af
8026  retn

1d0b  ld   a,h
1d0c  out  ($bf),a      ; attempt to write high address byte to port (writes low byte instead)
```

I also found [a report of similar issues](https://atariage.com/forums/topic/278887-does-anyone-notice-the-odd-graphical-glitch-on-frogger/) manifesting on real hardware, though that person was also using an SD cart adapter which could be responsible for disrupting the timing of the initial vblank. I don't know if this issue would reproduce on real hardware with a standard cart after enough attempts, or if the timing of the first vblank is consistent enough to avoid the bug entirely.

There are multiple ways this could be addressed in ares, but choosing one will require some thought.

LukeUsher:
Fixed by #113 , although as this is an issue with the game's logic, and can (sometimes) occur on real hardware, it's possible it may resurface in future.

PR #113, opened 2021-08-27
tms9918: widen vcounter to 9 bits

invertego:
By using an 8 bit counter, the last 6 lines of every frame were being
skipped and the VDP was firing vblank interrupts ~2.3% faster than
intended in the ColecoVision, SG-1000, and MSX.

This also fixes the corrupted turtle graphics in Frogger for
ColecoVision, though this is mostly by chance. The game is buggy and
susceptible to graphics corruption if a vblank is raised between these
instructions:

1d09  out  ($bf),a      ; write low VRAM address byte
1d0b  ld   a,h
1d0c  out  ($bf),a      ; write high VRAM address byte

It could easily be regressed by any future change that impacts timing.

### DIFF
--- a/ares/component/video/tms9918/tms9918.hpp
+++ b/ares/component/video/tms9918/tms9918.hpp
@@ -125,7 +125,7 @@ protected:
   } irqFrame;
 
   struct IO {
-    n8  vcounter;
+    n9  vcounter;
     n8  hcounter;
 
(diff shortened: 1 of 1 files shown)


invertego:
This fixes #84, though for the reasons given in the commit message, it won't be surprising if it comes back in the future.

PR #126, opened 2021-08-29
Fix window background with GTK3

ghisvail:
Same rationales as in https://github.com/bsnes-emu/bsnes/pull/127


ghisvail:
Without this patch the menubar is rendered in black when using GTK3.

With the patch:

Capture d’écran de 2021-08-29 12-05-57

Credits to @Screwtapello for the initial work in `bsnes`

PR #148, opened 2021-09-01
n64: Better Flashram support

jeltaqq:
Fix Majora's Mask, Paper Mario, Derby Stallion 64

### DIFF
--- a/ares/n64/cartridge/flash.cpp
+++ b/ares/n64/cartridge/flash.cpp
@@ -65,12 +65,12 @@ auto Cartridge::Flash::writeWord(u32 address, u64 data) -> void {
 
   case 0x78:  //erase
     mode = Mode::Erase;
-    status = 0x1111'8008'00c2'0000ull;
+    status = 0x1111'8008'00c2'001dull;
     return;
 
   case 0xa5:  //set write offset
     offset = u16(data) * 128;
-    status = 0x1111'8004'00c2'0000ull;
+    status = 0x1111'8004'00c2'001dull;
     return;
 
   case 0xb4:  //write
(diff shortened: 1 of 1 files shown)


LukeUsher:
As someone with very limited knowledge of n64 hardware, could you please explain what is going on here?

I would love to see a breakdown of what the status register values actually mean, and why changing the 0 -> 1 fixes these titles.

jeltaqq:
According to my limited understanding is that Derby Stallion 64 uses "11111800100C2001D" to check the status. So I guessed that "00C2001D" is good.

remutro:
I looked into this a little more. The second half of that value seems to represent the FlashRAM manufacturer and device code. You would think this wouldn't matter, but older Ids required special addressing. Ares was using addressing for new device Ids, while using an older device Id in its status messages. I believe this is why this change corrects the problems with saves. I really don't think Derby Stallion 64 specifically checks for an exact device Id. mupen64plus always uses 0x1E and doesn't have any problems with saves in that game. If you have more information regarding this game having a very specific check it would be great if you could post a link.

I've tested every game with FlashRAM with this change in place. It corrects Paper Mario, Majora's Mask, Derby Stallion 64, and also fixes Command & Conquer. Remaining FlashRAM games all still work (no regressions seen).

jeltaqq:
More information about Derby Stallion 64 was found https://github.com/project64/project64/issues/125. In conclusion, Project64 also uses 0x1E. I wonder if device Id 0x1D has been implemented, since using 0x1E in Ares did not solve the save.

remutro:
> More information about Derby Stallion 64 was found [project64/project64#125](https://github.com/project64/project64/issues/125). In conclusion, Project64 also uses 0x1E. I wonder if device Id 0x1D has been implemented, since using 0x1E in Ares did not solve the save.

Ares currently only uses one FlashRAM address lookup routine that should be compatible with device Ids 0x00c2001d (MX29L1101) & 0x003200f1 (MN63F8MPN).

PR #149, opened 2021-09-01
n64: Sets the initial values

jeltaqq:
I verified the initial values with test tools (COP0Cause, COP0Register, RDRAMTest).
https://github.com/PeterLemon/N64

### DIFF
--- a/ares/n64/cpu/cpu.hpp
+++ b/ares/n64/cpu/cpu.hpp
@@ -399,7 +399,7 @@ struct CPU : Thread {
   struct SCC {
     //0
     struct Index {
-      n6 tlbEntry;
+      n6 tlbEntry = 63;
       n1 probeFailure;
     } index;
 
@@ -417,7 +417,7 @@ struct CPU : Thread {
 
     //4
     struct Context {
-      n19 badVirtualAddress;
+      n19 badVirtualAddress = 0x7ffff;
       n41 pageTableEntryBase;
     } context;
 
@@ -428,7 +428,7 @@ struct CPU : Thread {
     } wired;
 
     //8
-    n64 badVirtualAddress;
+    n64 badVirtualAddress = 0xffffffff;
 
     //9
     n33 count;  //32-bit; +1 to count half-cycles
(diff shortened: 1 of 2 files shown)


jeltaqq:
Here are the test results.
COP0Cause
COP0Cause
COP0Register
COP0Register
RDRAMTest
RDRAMTest

LukeUsher (changes requested):
It would be better to set initial values in the power/reset functions, rather than in the class definition.

With this current approach, the initial values may not be restored when switching between games.

rasky:
The RDRAM registers are not built-in values. They get configures like this by IPL3 RDRAM calibration which is currently broken because the RI is not fully emulated. I would not hardcore them as boot values because it's not how the hardware works.

rasky:
The RDRAM registers are not built-in values. They get configured like this by IPL3 RDRAM calibration which is currently broken because the RI is not fully emulated. I would not hardcode them as boot values because it's not how the hardware works.

LukeUsher:
@rasky also raises a good point, so at the very least, please include a comment to explain that it's a hack/workaround, so that we can revisit it later.

jeltaqq:
Thanks for the good point. The RDRAM registers will be reverted as the less hacks the better. Also, COP0 sets the initial values in the power/reset function.

bigbass1997:
> the less hacks the better.

Agreed. I'd rather keep the number of hacks/workarounds to a minimum.

Shideravan:
I tested this commit with the latest code version, but didn't notice any emulation improvements on hard-setting these values. The compatibility with the latest ares dev build seems to be exactly the same as when these values aren't hardset for commercial games.
Probably because on booting N64 commercial games, the core quickly write over these values and the values hard-setted makes no difference.

LukeUsher:
> The RDRAM registers are not built-in values. They get configured like this by IPL3 RDRAM calibration which is currently broken because the RI is not fully emulated. I would not hardcode them as boot values because it's not how the hardware works.

@rasky Would you be able to create an issue to document this, and then I can close this PR

rasky:
@LukeUsher done: https://github.com/ares-emulator/ares/issues/350

PR #152, opened 2021-09-01
https://github.com/higan-emu/ares/issues/61 - Workaround

remutro:
This is a workaround to address issues with playing Battletanx (U) & Battletanx - Global Assault (U) in this emulator. This will allow everyone to play these two games until the boot process is better understood and we better understand what is missing that prevents these games from booting. Extensive testing has been performed against the entire N64 USA rom set to make sure this doesn't negatively impact any other games. 

Again, this is just a workaround, with comments to indicate this should be updated once we have a better understanding of the overall boot process and identify what is missing here.


remutro:
Hold on... need to rebase and resubmit. Sorry....

PR #153, opened 2021-09-01
Workaround for Github issue: https://github.com/higan-emu/ares/issues/61

remutro:
This is a workaround to address issues with playing Battletanx (U) & Battletanx - Global Assault (U) in this emulator. This will allow everyone to play these two games until the boot process is better understood and we better understand what is missing that prevents these games from booting. Extensive testing has been performed against the entire N64 USA rom set to make sure this doesn't negatively impact any other games.

Again, this is just a workaround which has been discussed on discord, with comments to indicate this should be updated once we have a better understanding of the overall boot process and identify what is missing here. For now, end users will just be able to enjoy these games.

### DIFF
--- a/ares/n64/pi/pi.cpp
+++ b/ares/n64/pi/pi.cpp
@@ -41,15 +41,22 @@ auto PI::power(bool reset) -> void {
   //d18    = osVersion
   //d19    = osRomType (0 = Gamepak; 1 = 64DD)
   string cic = cartridge.cic();
-  if(reset == 0) {
+
+  //FIXME: https://github.com/higan-emu/ares/issues/61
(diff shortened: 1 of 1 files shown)


bigbass1997:
I approve of this change. I've talked to remutro about it before on Discord. The existing implementation of this particular section of code, should be console accurate as far as I can tell. But this temporary change is necessary so people unfamiliar with this issue, can play the aforementioned games.

Shideravan:
I also approve. While we know that no workaround is really good in the long run, this commit provides a temporary fix and is very well aware of everybody. This may even help to have more use of the core in some scenarios which can help to have more feedback on the core which can even help in a future more permanent solution. I trust the extensive testing that @remutro has done to prevent this workaround to cause breaks in other situations, but of course, we will have to keep an eye on the core and continue to understand better about the boot process in order to later eliminate this shortcut ASAP.

Issue #125, opened 2021-08-28
[N64] Blast Corps (U): Image flickering in intro


Shideravan:
In the game Blast Corps, the image is flickering on the title screen. The game doesn't pass to enter in gameplay due the controller pak isn't fully implemented yet, but probably would happens in-game too.

remutro:
Game doesn't play at the correct speed because of issues documented in: https://github.com/higan-emu/ares/issues/14

**PLEASE be aware of this fact as we don't need an additional issue filed for every instance of the same problem in every game.** 

Saves/controller pak/rumble handling is currently insufficiently implemented in Ares, which is why you are seeing this controller pak error. Ares currently decides that if EEPROM, Flash, or SRAM is detected as a save option, then controller 1 will always be set to rumble. If EEPROM, Flash, or SRAM is not being used by the game, then controller 1 will be set to have a controller pak connected. Controllers 2 - 4 always have a rumble pak attached. And there currently isn't any way to change this behavior, or switch between rumble/controller pak in-game. Blast Corps is one game that has a 4Kb EEPROM as well as controller pak and doesn't have support for rumble. 

The list of games impacted is known in the team. Please make sure not to file additional issues like this for each additional game, which requires confirming what type of saves are supported when testing and understanding how controllers pak/rumble selection works (or you can, but they will likely get closed as duplicates :-) ).

Shideravan:
Ok, as the controller pak it's an already known issue, I'll removing this part and keep just the image flickering in the issue description. Thanks you for the heads up :)

remutro:
The controller warnings are fixed as of pull request 156. Game speed/flickering is believed to be caused by issues related to the RDP freeze flag not being properly implemented discussed in issue #14. Not sure if this ticket needs to remain open now if the second half is captured in another issue.

Shideravan:
Ok, issue closed

PR #156, opened 2021-09-05
Add mempak and rumble status to mia; eliminate most warnings in games

remutro:
This pull request adds controller pak and rumble support to mia's database which will now be checked when a game is loaded. This eliminates the majority of warnings/errors in games where Ares is assigning devices the game does not expect/support. Users will no longer need to configure buttons for a 2nd, 3rd, 4th controller pak just to bypass error screens in games. 

This is similar to other emulators, such as mupen64plus that store this information since it can't be determined based on the rom alone. Mupen's ini file was used as a basis for this information, but it was also cross-checked against multiple other rumble or controller pak lists so it should be fairly accurate. 

Logic is changed now so that if the cartridge has a save mechanism built-in, it doesn't mean we can't also support a controller pak as a secondary save mechanism which a number of games supported. Any cartridge based save mechanism will always be loaded (according to mia's database). If a game also supports a controller pak, then controller 1 will have a controller pak attached. If a game supports both controller pak and rumble, controller 1 will default to a controller pak for higher compatibility. Controller's 2-4 will never have a controller pak assigned (it will cause a file name collision based on Ares current implementation). If a game supports rumble, controller 1 will get rumble attached (if it doesn't support controller pak), and controllers 2-4 will also get rumble attached. If a game doesn't support controller pak or rumble, no controller will get these devices attached.

This still doesn't enable the ability to switch between controller pak/rumble in games that support it. Although it probably wouldn't be difficult now to add menu options based on the mia database that would allow a user to switch between the two. And if we can alter controller pak save file names (<game>.1.pak, <game>.2.pak, etc.), then this could be enabled for all controllers as well. 

This resolves issues raised in: https://github.com/higan-emu/ares/issues/138

It also addresses half of the issues identified in: https://github.com/higan-emu/ares/issues/125, with the other half of the issues in this ticket already being covered by: https://github.com/higan-emu/ares/issues/14, so issue #125 can likely be closed.

### DIFF
--- a/desktop-ui/emulator/nintendo-64.cpp
+++ b/desktop-ui/emulator/nintendo-64.cpp
@@ -60,20 +60,19 @@ auto Nintendo64::load() -> bool {
     port->connect();
   }
 
+  string rumble = game->pak->attribute("rumble"); 
   if(auto port = root->find<ares::Node::Port>("Controller Port 1")) {
     auto peripheral = port->allocate("Gamepad");
     port->connect();
-    if(auto port = peripheral->find<ares::Node::Port>("Pak")) {
-      if(!game->pak->read("save.ram")
-      && !game->pak->read("save.eeprom")
-      && !game->pak->read("save.flash")
-      ) {
+    if(auto port = peripheral->find<ares::Node::Port>("Pak")) { 
+      if(game->pak->attribute("mempak") == "yes")
+      {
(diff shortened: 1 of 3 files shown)


LukeUsher (approved):

Shideravan:
Nice!

Issue #142, opened 2021-08-30
[N64] Wave Race - Shindou Edition: Freezes in N64 logo


Shideravan:
This one is a very interesting issue as this affects only the Shindou edition, not the default version of Wave Race.
![image](https://user-images.githubusercontent.com/12163923/131363680-aab8ff85-bca7-4c5b-a6a1-12cd1b39cf92.png)

remutro:
@Shideravan - Now that invertigo's pull request has been merged, the Shindou edition of Wave Race 64 should play fine (and also enable rumble). Please test/close this issue when you get a chance. Thanks!

Shideravan:
Fixed with #161

PR #161, opened 2021-09-09
n64: consume cycles to compensate for skipped RDRAM initialization

invertego:
The RI_SELECT register is read by IPL3 to check if RDRAM initialization
should be skipped. If we are forcing it to be skipped, we should also
consume enough cycles to not inadvertently speed up the boot process.

Wave Race 64 Shindou Pak Taiou Version will freeze on the N64 logo if
the SCC count register, which increments at half the CPU clock rate, has
too small a value.

The idea of using RI_SELECT reads as a cue to consume extra cycles was
borrowed from MAME.

Thanks to @rasky for measuring the timing on real hardware.

### DIFF
--- a/ares/n64/ri/io.cpp
+++ b/ares/n64/ri/io.cpp
@@ -20,6 +20,18 @@ auto RI::readWord(u32 address) -> u32 {
   if(address == 3) {
     //RI_SELECT
     data = io.select;
+    if constexpr(!Accuracy::RDRAM::Broadcasting) {
+      //this register is read by IPL3 to check if RDRAM initialization should be
+      //skipped. if we are forcing it to be skipped, we should also consume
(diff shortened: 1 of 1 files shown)


invertego:
This fixes #142.

It may not be terribly obvious, but the reason this is behind the compile time Accuracy::RDRAM::Broadcasting check is because the code that forces RDRAM initialization to be skipped is also behind this check:

https://github.com/higan-emu/ares/blob/fe94e4046cc7e3be5480937eb14dfb8634be9f9b/ares/n64/ri/ri.cpp#L22

I think the intention was that because the RDRAM initialization code uses broadcast writes, it wouldn't make sense to run it until that feature had been implemented. It remains unimplemented in ares, for now.

PR #164, opened 2021-09-09
Only report N64 gamepads that are currently connected

remutro:
Only report N64 gamepads that are currently connected


remutro:
Canceling this pull request because this also got included in the other one and I didn't notice (not sure how that happened)

Issue #61, opened 2021-08-17
[N64] BattleTanx (U): doesn't boot/black screen


AbreuEfechou:
Version: latest
ROM CRC-32: 6C230765 (No-intro)

bigbass1997:
Once you load the ROM, go to the top menu, under Nintendo 64, click Reset. Wait a moment, and then it should work. It's a known issue.

Previously discussed on Discord, it has something to do with the reset state of the emulator. While there is a one bit difference on soft reset (or NMI), in PIF RAM at bootup, this is emulated already here: https://github.com/higan-emu/ares/blob/d653ca8f22a2dd029f54f9733e17599feb91d287/ares/n64/pi/pi.cpp#L43-L58

I suspect it might be something unrelated to the PIF RAM state at cold-boot/reset, as that does appear to be console accurate. But it's impossible to tell what the real root cause is without a thorough investigation into what the game gets hung up on, on cold boot, rather than a reset.

remutro:
The same issue impacts  Battletanx: Global Assault, but only the USA ROM. Curiously, the EU Battletanx: Global Assult ROM does not experience this issue. As bigbass1997 already pointed out, you can work around this issue for both USA ROMs by resetting these games after loading.

remutro:
Both games should now boot on first try now and can be played as of pull request: https://github.com/higan-emu/ares/pull/153

Can you close this issue please?

PR #165, opened 2021-09-10
n64: fix PIF RAM init on cold boot, preserve RDRAM on warm reset

invertego:
A recent workaround for BattleTanx caused a regression in Perfect Dark.
Essentially, it was always telling games that they had been started by a
warm reset rather than a cold boot. Perfect Dark expects data to be
preserved in RDRAM across resets and fails to get as far as drawing
something on the screen if it isn't there.

This change reverts the workaround so that Perfect Dark will perform the
full initialization required after a cold boot. Luckily, this will not
regress the BattleTanx games, as they were fixed by my last change
(making sure IPL3 consumes a sufficient number of cycles). This change
also skips clearing RDRAM on resets, which is again required by Perfect
Dark and possibly other games.

### DIFF
--- a/ares/n64/pi/pi.cpp
+++ b/ares/n64/pi/pi.cpp
@@ -35,34 +35,22 @@ auto PI::power(bool reset) -> void {
   bsd2 = {};
 
   //write CIC seeds into PIF RAM so that cartridge checksum function passes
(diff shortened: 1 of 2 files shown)


invertego:
The workaround referenced in the description is #153 and the BattleTanx issue is #61.

Issue #128, opened 2021-08-29
[N64] Jeopardy: Controllers not working


Shideravan:
The game seems to run normally, but none of the buttons in the controllers are working.

ares e6948d2

remutro:
Confirmed, known issue.

invertego:
As it turns out, this is a game bug. The game only supports up to 3 players, and if four controllers are plugged in when the game starts, it performs an out of bounds array access that breaks further input logic. The only workaround is to unplug a controller and reset the system. To make this game work well in ares out of the box, it will need to connect fewer than 4 controllers by default.

Note, this bug is also compounded by a couple of other bugs in ares. Currently, the frontend ignores attempts to disconnect controllers. Furthermore, the n64 core reports 4 controllers to the game no matter how many are connected according to the frontend.

remutro:
The last issue invertigo mentions here (n64 core reporting 4 controllers no matter what) will be addressed in pull request: https://github.com/higan-emu/ares/pull/163 

It's not enough to fix this issue though, as currently we do attach 4 controllers by default. Maybe we should attach 2 controllers by default, and users can add controller 3 & 4 if necessary through the menu which currently works. Hopefully there isn't too much of a timing issue the other way around with this game where if you can't disable a 4th controller in time this game will always remain broken in Ares due to the bug that's present with the game. Even if the "Nothing" option worked, you don't get a chance to select it until after you've loaded the game/core. 

With recent fixes in place, compiling Ares without having the 4th controller connected does in fact resolve controller issues in this game and it is playable.

Shideravan:
Fixed with #163 and #166

PR #166, opened 2021-09-11
n64: default to 3 controllers for Jeopardy!

invertego:
Jeopardy! only supports up to 3 players, and if four controllers are
plugged in when the game starts, it performs an out of bounds array
access that breaks further input logic. The only workaround is to unplug
a controller and reset the system. To make this game work well in ares
out of the box, we should connect no more than 3 controllers by default.

I opted to put a check specifically in desktop-ui because this is
fundamentally a mismatch between a specific game and a design choice (to
default to 4 controllers) in the desktop-ui frontend. It didn't feel
right to add a dedicated attribute to mia that did not describe a
feature of the cartridge itself and would not be generally applicable
other games (unlike rumble and memory pak support).

This change also brings the rumble and memory pak attributes in line
with the implementation of boolean attributes for other systems. There
is no functional change, though it will affect generated manifests if
they are/were persisted to disk.

### DIFF
--- a/desktop-ui/emulator/nintendo-64.cpp
+++ b/desktop-ui/emulator/nintendo-64.cpp
@@ -60,55 +60,25 @@ auto Nintendo64::load() -> bool {
     port->connect();
   }
 
-  string rumble = game->pak->attribute("rumble"); 
-  if(auto port = root->find<ares::Node::Port>("Controller Port 1")) {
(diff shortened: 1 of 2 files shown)


invertego:
Fixes #128 when combined with #163

LukeUsher (approved):

PR #188, opened 2021-09-18
enable clang builds for Windows

invertego:
clang/LLVM has a great tools ecosystem and supports things for Windows that GCC does not (e.g. targeting ARM64, generating PDBs). I ran into several compile and run time issues in the course of building with clang, addressed in this series of commits, broken out for clarity of purpose. I stopped short of changing the default compiler to clang on Windows, but we may want to consider doing so in the future.


ares/ares/node/object.hpp
```diff
@@ -1,14 +1,14 @@
-//identifier is static, allowing template<typename T> to access via T::identifier
+//identifier() is static, allowing template<typename T> to access via T::identifier()
 //identity() is virtual, allowing T* to access via T->identity()
 
 #define DeclareClass(Type, Name) \
-  static inline const string identifier = Name; \
+  static auto identifier() -> string { return Name; } \
```
Screwtapello, line 5: Does Clang not support constants in a class?

ares/component/processor/tlcs900h/instruction.cpp
```diff
@@ -630,172 +625,226 @@ auto TLCS900H::instructionRegister(R register) -> void {
...
   }
   case 0x06: {  //CPL r
-    if constexpr(Long) return undefined();
-    prefetch(4);
-    return instructionComplement(register);
+    if constexpr(!Long) {
+      prefetch(4);
+      return instructionComplement(register);
+    }
+    return undefined();
```
Screwtapello, line 632: Why does Clang require inverting the `if` test?

Screwtapello (commented):
Generally this looks good to me. Would it be difficult to add a Windows/Clang build to CI, so we can be sure this support doesn't regress in future? It doesn't have to publish a binary artefact, but just running the compiler and checking it doesn't explode would be a good start.

ares/ares/node/object.hpp
invertego: It does, but there is an issue with static initialization order. Please see the associated commit message.

ares/component/processor/tlcs900h/instruction.cpp
invertego: The early return wasn't short-circuiting anything at compile time. The compiler still wanted to follow the subsequent calls and instantiate templated functions with unsupported sizes. GCC was generating warnings here too; this cleanup applies to both it and clang.

invertego:
> Generally this looks good to me. Would it be difficult to add a Windows/Clang build to CI, so we can be sure this support doesn't regress in future? It doesn't have to publish a binary artefact, but just running the compiler and checking it doesn't explode would be a good start.

I tried setting this up, and it was all very fiddly until I switched to using packages from msys2. Here's what I came up with: https://github.com/invertego/ares/commit/e0ce9727fa593f3a3057a9d532dd41ad37539805
Sample run: https://github.com/invertego/ares/actions/runs/1251913893

I can make that a part of this PR if it looks reasonable. That commit also switches the existing GCC-based Windows build to use the msys2 packaged version. Our build is currently relying on a Chocolatey MinGW package containing GCC 8.1, which is at this point over 3 years old.

Screwtapello:
> I can make that a part of this PR if it looks reasonable. That commit also switches the existing GCC-based Windows build to use the msys2 packaged version. Our build is currently relying on a Chocolatey MinGW package containing GCC 8.1, which is at this point over 3 years old.

It looks reasonable!

I'm a bit disappointed in GitHub Actions. Looking at [the list of software on Windows images](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md), it looks like they've been keeping most things (Python, Perl, Rust, Go, Visual Studio, Firefox, Chrome, Git, etc.) quite up-to-date, so I'm not sure why their MinGW package is so stale.

It's a bit annoying that we need two separate copies of the "make" step, whose only difference is that the Windows one asks to be executed inside msys2. How difficult would it be to [install the latest version from Chocolatey](https://community.chocolatey.org/packages/mingw) instead? On the other hand, [the beta release of the 2022 Windows image](https://github.com/actions/virtual-environments/issues/3949) specifically encourages installing MSYS2 in the way you're doing, so it's probably not a bad idea to do it that way.

invertego:
> It's a bit annoying that we need two separate copies of the "make" step, whose only difference is that the Windows one asks to be executed inside msys2. How difficult would it be to [install the latest version from Chocolatey](https://community.chocolatey.org/packages/mingw) instead? On the other hand, [the beta release of the 2022 Windows image](https://github.com/actions/virtual-environments/issues/3949) specifically encourages installing MSYS2 in the way you're doing, so it's probably not a bad idea to do it that way.

I tried updating the MinGW Chocolatey package (and installing the LLVM one), and while this did work, it was very fiddly. The LLVM package is built with MSVC as the default target, so it must be overridden to use MinGW headers and libs. Normally it can discover these simply by looking up GCC in your path, but Chocolatey doesn't add the MinGW bin directory to your path. It puts [shim executables](https://docs.chocolatey.org/en-us/features/shim) for all of the individual binaries on your path. Now you have to work out the actual install directory and add that to your path.

Do all that, and there's still the fact that our Windows builds were using Powershell, and since the ares makefiles include random bash commands (e.g. cp, test), the only reason these worked is that the Windows build image puts the Git for Windows msys2 /usr/bin directory the path, and it contains standalone executables named after these commands. I found complaints about this in some GitHub Actions issues, as these tools are only meant for use from a GfW bash shell.

I believe using msys2 as a shell and installing a deliberately chosen set of packages for our build will create fewer headaches.

Screwtapello:
Yep, having a second "make" step does sound like less of a hassle than all that other stuff. Thanks for looking into it!

LukeUsher:
Nice work!

Just a heads up that I'm open to having the released builds be produced by clang on all platforms, for consistency. We should ensure that gcc can still build ares though, so that stage should remain in CI, but only clang artefacts should be published imo.

PR #189, opened 2021-09-19
Improvements the macOS UI and UX

LIJI32:
This PR introduces many massive improvements to the macOS UI and UX, as well as general performance improvements. It also adds several new hiro APIs that should allow several improvements to other frontends (e.g. High DPI icons, system native color translation). Some of these changes affect the other frontends, but only the macOS version was tested.


desktop-ui/presentation/presentation.cpp
```diff
@@ -2,6 +2,12 @@
 namespace Instances { Instance<Presentation> presentation; }
 Presentation& presentation = Instances::presentation();
 
+#if defined(PLATFORM_MACOS)
+#define ELLIPSIS "…"
```
Screwtapello: In Python, if you use non-ASCII characters you need an encoding declaration to specify what encoding to use. I'd like to think that all compilers on macOS would assume (or even require UTF-8), but because of C's history it wouldn't surprise me if there's some circumstance where it's actually MacRoman or Windows1252 or something. What are the risks here?

desktop-ui/presentation/presentation.hpp
```diff
@@ -53,7 +53,10 @@ struct Presentation : Window {
       MenuCheckItem pauseEmulation{&toolsMenu};
       MenuSeparator toolsMenuSeparatorB{&toolsMenu};
       MenuItem manifestViewerAction{&toolsMenu};
+      #if !defined(PLATFORM_MACOS)
+      // Cocoa hiro is missing the hex editor widget
       MenuItem memoryEditorAction{&toolsMenu};
+      #endif
```
Screwtapello, line 59: There's already a bunch of widgets that are platform specific:

https://github.com/higan-emu/ares/blob/c71f131b0cb2722c5c5d0591da87edeff816898d/hiro/components.hpp#L80-L85

Rather than adding a `PLATFORM_MACOS` check here, you should probably add a block like this in hiro to exclude the HexEdit widget on macOS, and add a check for `Hiro_HexEdit` here.

desktop-ui/program/utility.cpp
```diff
@@ -42,6 +42,8 @@ auto Program::openFile(BrowserDialog& dialog) -> string {
     window.setPath(dialog.path());
     window.setFilters(dialog.filters());
     window.setParent(dialog.alignmentWindow());
+    // Only affects macOS. TODO: are there scenarios where we want to forbid selecting folders?
+    window.setAllowsFolders(true);
```
Screwtapello, line 46: Ideally, we'd only allow selecting folders with extensions matching the given filters, not all folders.

nall/image.hpp
```diff
@@ -157,6 +157,30 @@ struct image {
   channel _blue {255u <<  0, 8,  0};
 };
 
+struct multiFactorImage : public image {
+    using image::image;
+
+    multiFactorImage(const multiFactorImage& source);
+    multiFactorImage(multiFactorImage&& source);
+    multiFactorImage(const image& lowDPI, const image& highDPI);
```
Screwtapello, line 165: If this is going to be a "multifactor" image, instead of just a "bifactor" image, shouldn't the API support an arbitrary number of factors?

desktop-ui/presentation/presentation.cpp
Kawa-oneechan: I strongly suggest using escape codes here instead of encoding declarations. `\u2026` for anything Unicode, which I would assume is the correct choice regardless and should automatically convert down to the correct format, `\x85` for Win-1252, `\xC9` for Mac Roman.

LukeUsher: Note that as long as the source file is stored with the correct encoding (recommended to be utf8) unicode should work everywhere.

Windows code should be doing utf8->utf16 conversion before calling WinAPIs with strings, when necessary

Screwtapello:
I built this PR for GTK+2 (Near's primary platform) and GTK+3 and took comparison screenshots so people can see the difference.

(ignore that the window close button switches sides occasionally, I happened to toggle that setting while doing something else)

<details>
<summary>Driver Settings in GTK+2 (current)</summary>

ares-gtk2-settings-drivers-master

</details>
<details>
<summary>Driver Settings in GTK+2 (this PR)</summary>

ares-gtk2-settings-drivers-liji

</details>
<details>
<summary>Video Settings in GTK+2 (current)</summary>

ares-gtk2-settings-video-master

</details>
<details>
<summary>Video Settings in GTK+2 (this PR)</summary>

ares-gtk2-settings-video-liji

</details>
<details>
<summary>Driver Settings in GTK+3 (current)</summary>

ares-gtk3-settings-drivers-master

</details>
<details>
<summary>Driver Settings in GTK+3 (this PR)</summary>

ares-gtk3-settings-drivers-liji

</details>
<details>
<summary>Video Settings in GTK+3 (current)</summary>

ares-gtk3-settings-video-master

</details>
<details>
<summary>Video Settings in GTK+3 (this PR)</summary>

ares-gtk3-settings-video-liji

</details>

As it happens, GTK+3 already nicely supports dark mode without any extra special effort, *except* the "hint" text on the video settings. It would be nice if this `SystemColor` abstraction could be more easily used by other backends - I don't know how difficult this would be in C++, but maybe if the conversion from `hiro::SystemColor` to `hiro::Color` was a method on `hiro::pWindow` so it could be inherited by all widgets, instead of having to add `setForegroundColor(SystemColor c)` and `setBackgroundColor(SystemColor c)` overloads to every single widget?

desktop-ui/presentation/presentation.cpp
LIJI32: C and Objective-C strings, file names, text files and so on are effectively guaranteed to be UTF-8 in macOS unless explicitly specified otherwise, in additional to what @LukeUsher already said. That said, I agree with Kawa's suggestion to use `\u2026` to avoid having non-ANSI characters in the file to begin with (so any accidental encoding change done on some other platform won't break macOS without anyone realizing).

desktop-ui/presentation/presentation.hpp
LIJI32: That's true, I haven't considered it. Good idea.

desktop-ui/program/utility.cpp
LIJI32: I was given the impression that game folders are not required to have an extension?

nall/image.hpp
LIJI32: Yes and no, I used multi-factor because some platforms might benefit from "mid DPI" (x1.5) or "very high DPI" (x3) images, but since I'm not very familiar with GUI development on these I didn't want to commit to a specific API. I went with something that is currently limited to two factors but can be later be extended with more methods as needed by other frontends.

LIJI32:
The reason I overload setForegroundColor is to support dynamic theme changes, so a widget knows a color should change in case a theme changes. This isn't something you can do with hiro::Color directly.

desktop-ui/presentation/presentation.hpp
LIJI32: Looks like Hiro_HexEdit is defined even for macOS, despite not being implemented :S

LukeUsher:
I went ahead and rebased this on master, to solve the merge conflicts

PR #190, opened 2021-09-19
n64: implement perfect PI DMA write

rasky:
This commit implement perfect PI DMA write. The PI DMA is normally used
for 8-byte aligned transfers, which is where it constantly works. The
hardware does contain a half-baked, buggy support for 2-byte aligned
transfers and odd-size lengths. I have extensively investigated the
behavior of every single quirk, and produced a test ROM that Ares
passes after this commit (timing is still off, but that's for another
day):

https://github.com/rasky/n64_pi_dma_test

The implementation is the same that I've previously submitted to cen64,
adapted for the ares codebase.


Shideravan:
Amazing thank you!

PR #191, opened 2021-09-19
n64: implement accurate timer interrupt in recompiler

rasky:
Previously, the timer interrupt was only being checked at the end of
each recompiler blocks, so each interrupt was potentially delayed until
next branch. Now, it is checked as part of CPU::step so it is
cycle-accurate. If the timer interrupt is triggered, the recompiler
is early-exited so that the interrupt can actually occur precisely
at the right moment.

On top of that, the previous code had also a overflow-checking bug,
so it failed to trigger an interrupt when executing a block in which
the counter overflown and compare was reached at the same time.

### DIFF
--- a/ares/n64/cpu/cpu.cpp
+++ b/ares/n64/cpu/cpu.cpp
@@ -35,6 +35,15 @@ auto CPU::main() -> void {
 
 auto CPU::step(u32 clocks) -> void {
   Thread::clock += clocks;
+
+  i33 dist = scc.compare - scc.count;
+  if(0 <= dist && dist < clocks){
+    scc.cause.interruptPending.bit(Interrupt::Timer) = 1;
+    if constexpr(Accuracy::CPU::Recompiler) {
+      recompiler.earlyExit = true;
+    }
+  }
+  scc.count += clocks;
 }
 
 auto CPU::synchronize() -> void {
@@ -59,12 +68,6 @@ auto CPU::synchronize() -> void {
     case Queue::SI_DMA_Write:  return si.dmaWrite();
     }
   });
-
-  clocks >>= 1;
-  if(scc.count < scc.compare && scc.count + clocks >= scc.compare) {
-    scc.cause.interruptPending.bit(Interrupt::Timer) = 1;
-  }
-  scc.count += clocks;
 }
 
 auto CPU::instruction() -> void {
@@ -79,6 +82,7 @@ auto CPU::instruction() -> void {
   if constexpr(Accuracy::CPU::Recompiler) {
     auto address = devirtualize(ipu.pc)(0);
     auto block = recompiler.block(address);
+    recompiler.earlyExit = false;
     block->execute();
   }
 
@@ -103,8 +107,8 @@ auto CPU::instructionEpilogue() -> bool {
   }
 
   switch(branch.state) {
-  case Branch::Step: ipu.pc += 4; return 0;
-  case Branch::Take: ipu.pc += 4; branch.delaySlot(); return 0;
+  case Branch::Step: ipu.pc += 4; return recompiler.earlyExit;
+  case Branch::Take: ipu.pc += 4; branch.delaySlot(); return recompiler.earlyExit;
   case Branch::DelaySlot: ipu.pc = branch.pc; branch.reset(); return 1;
   case Branch::Exception: branch.reset(); return 1;
   case Branch::Discard: ipu.pc += 8; branch.reset(); return 1;
--- a/ares/n64/cpu/cpu.cpp
+++ b/ares/n64/cpu/cpu.cpp
@@ -35,20 +35,12 @@ auto CPU::main() -> void {
 
 auto CPU::step(u32 clocks) -> void {
   Thread::clock += clocks;
-
-  i33 dist = scc.compare - scc.count;
-  if(0 <= dist && dist < clocks){
-    scc.cause.interruptPending.bit(Interrupt::Timer) = 1;
-    if constexpr(Accuracy::CPU::Recompiler) {
-      recompiler.earlyExit = true;
-    }
-  }
-  scc.count += clocks;
 }
 
 auto CPU::synchronize() -> void {
   auto clocks = Thread::clock * 2;
   Thread::clock = 0;
+  prevClock = 0;
 
    vi.clock -= clocks;
    ai.clock -= clocks;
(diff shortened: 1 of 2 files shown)


ares/n64/cpu/cpu.cpp
```diff
@@ -102,9 +96,15 @@ auto CPU::instructionEpilogue() -> bool {
     scc.random.index = 31;
   }
 
+  bool exc = 0;
+  if(scc.count++ == scc.compare){
```
invertego: Previously, scc.count was derived from the CPU clock. Now, it is counting instructions retired. This is ignoring the timing effects of (for example) icache/dcache misses. Was this intended?

Shideravan:
Nice one! Thank you 😊

rasky:
Yesterday I got confused about all the cycles shift. I revisited the code today and realized that it's still necessary to keep a 33-bit compare / count now that we increment count for each cycle (as count counts half-cycles).

rasky:
I changed the implementation to do the check in `CPU::step` so that it should be cycle-accurate in any case, including stalls.

ares/n64/cpu/cpu.cpp
```diff
@@ -35,6 +35,16 @@ auto CPU::main() -> void {
 
 auto CPU::step(u32 clocks) -> void {
   Thread::clock += clocks;
+
+  clocks >>= 1;
```
invertego: This is going to be rounding off the last bit more often than before (after every step rather than every instruction w/ the interpreter or every block w/ the recompiler). I don't know how significant that is likely to be.

ares/n64/cpu/cpu.cpp
```diff
@@ -35,6 +35,16 @@ auto CPU::main() -> void {
...
+  i33 dist = scc.compare - scc.count;
+  if(0 <= dist && dist < clocks){
```
invertego: Is there any correctness benefit to checking the compare condition here rather than instructionEpilog()? The interrupt won't be handled until the beginning of the next instruction. I'm concerned about the performance implications of potentially redundant work being performed on every call to step().

Performing the same check in instructionEpilog() would require knowing the cycle count at the start of the instruction, which could be stored in a field.

ares/n64/cpu/cpu.cpp
```diff
@@ -35,6 +35,16 @@ auto CPU::main() -> void {
...
+    scc.cause.interruptPending.bit(Interrupt::Timer) = 1;
+    if constexpr(Accuracy::CPU::Recompiler) {
+      recompiler.earlyExit = true;
+    }
+  }
+  scc.count += clocks;
```
invertego: An alternative approach would be to store scc.count as an offset relative to Thread::clock. The value of scc.count could be derived from these two values only as needed. This would avoid the need to increment both counters in parallel and would address the rounding issue mentioned in my other comment.

invertego:
This now looks broadly correct from a code inspection perspective. I think there is an opportunity to avoid special casing the timer interrupt and also improve recompiler timing across the board, but it doesn't have to be done in this change. What I had in mind:
- Add timer interrupts to the CPU scheduling queue like DMAs
- In instructionEpilog(), exit the block if the first event in the queue is ready to run
This would allow the existing scc.compare check to stay in CPU::synchronize().

invertego (commented):
~~This now looks broadly correct from a code inspection perspective.~~ At first glance I missed that this change was halving the counter speed. See my comment below.

I think there is an opportunity to avoid special casing the timer interrupt and also improve recompiler timing across the board, but it doesn't have to be done in this change. What I had in mind:

- Add timer interrupts to the CPU scheduling queue like DMAs
- In CPU::instructionEpilog(), exit the block if the first event in the queue is ready to run

This would allow the existing scc.compare check to stay in CPU::synchronize().

ares/n64/cpu/cpu.cpp
rasky: This is the previous version / outdated code. I’m closing this

ares/n64/cpu/cpu.cpp
rasky: I think CPU::step() should only be called with even numbers. That’s because it’s upscaling by 2 the actual clock cycles

invertego: There are a number of step(1) calls in cpu/memory.cpp. step() operates in terms of real cycles; the scaling by 2 happens at the top of CPU::synchronize().

Also, I just noticed that this change is halving the counter speed from what it was before. The shift by 1 is only necessary to cancel out the * 2 at the top of CPU::synchronize().

The SCC counter increments every other cycle, but that is already accounted for by a shift when it is read in CPU::getControlRegister().

invertego:
I happened across another bit of code that further complicates things. The recompiler can emit code that directly modifies the clock counter:
https://github.com/higan-emu/ares/blob/ed4d21d7a35ee1f1e3e4ba44f40781d2026bffb3/ares/n64/cpu/recompiler.cpp#L39-L44
The existing implementation handles these cycles in synchronize(), but the code in this PR currently does not.

ares/n64/cpu/cpu.cpp
rasky: It doesn't sound like there's a good place where to record the cycle count at the beginning of each instruction in the recompiler, isn't there?

ares/n64/cpu/cpu.cpp
rasky: The same comment applies: to make the check in the epilogue, we need to know `Thread::clock` at the beginning of each instruction, and I can't see a way to save that in a field in the recompiler.

ares/n64/cpu/cpu.cpp
rasky: Correct, I'll drop the shift here.

ares/n64/cpu/cpu.cpp
invertego: More precisely, the field would represent the cycle count after the last time this logic ran. The logical place to store that would be right after this code.

rasky:
I've pushed a commit that switches to doing the timer interrupt check only in `instructionEpilogue` as requested by the review. I can't personally vet it's correct. It does pass the libdragon testsuite which includes a stress test of the timer interrupt, but:

1) I can't personally vet that all cases where an instruction causes the timer interrupt to occur now triggers. For instance, I don't know if there are (or there will be in the future) calls to `Cpu::step` between `instructionEpilogue()` and `synchronize()`, which would make the thread clock advance "too late".  We could enforce this by making sure that `prevCount` is equal to `Thread::clock` when `synchronize` is called. I can add an assert there, though there are no other asserts in the code base (?).
2) There are surely race conditions if a timer interrupt occurs mid-instruction (eg: during RDRAM stalls on instruction fetch) on instructions that modify `count` or `compare` (`MTC0`). This could be fixed with a call to `checkTimerInterrupt` in the right places, but I don't have good tests for this.

In general, I feel this version is harder to verify that it is correct and possibly more fragile to future modifications of the code base. I can't deny that it must be faster than the previous one, though I don't know how much and I don't have means to do correct timings.

rasky:
> I happened across another bit of code that further complicates things. The recompiler can emit code that directly modifies the clock counter:
> 
> https://github.com/higan-emu/ares/blob/ed4d21d7a35ee1f1e3e4ba44f40781d2026bffb3/ares/n64/cpu/recompiler.cpp#L39-L44
> 
> 
> The existing implementation handles these cycles in synchronize(), but the code in this PR currently does not.

We can fix this by changing the idle skip to a call to `Cpu::step(64)`. I can do that if we decide to go with the first version of the PR

PR #198, opened 2021-09-22
recompiler: fix instructionEpilogue() test

invertego:
instructionEpilogue() returns a bool, but all recompilers were
performing a 64-bit test on the return value. If sizeof(bool) == 1, only
the bottom 8 bits are guaranteed to be valid.

### DIFF
--- a/ares/component/processor/sh2/cached.cpp
+++ b/ares/component/processor/sh2/cached.cpp
@@ -47,7 +47,7 @@ auto SH2::Recompiler::emit(u32 address) -> Block* {
     address += 2;
     if(hasBranched || (address & 0xfe) == 0) break;  //block boundary
     hasBranched = branched;
-    test(rax, rax);
+    test(al, al);
     jz(imm8(1));
     ret();
   }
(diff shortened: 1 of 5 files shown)


rasky (commented):
Is it guaranteed that the lowest 8 bits are nonzero for a true Boolean? Maybe you should generate the right test depending on bool sizeof?

invertego:
Regarding the bit representation of bools, the C++ standard just says "the value false is converted to zero and the value true is converted to one" [conv.integral]. In theory a compiler could pick any bit representation it wants until you cast to an integral type (including 0 for true and 1 for false), so checking the size alone wouldn't be enough to handle a sufficiently adversarial compiler. 😈

I would sooner change the return type to u8 for consistency of both size and bit representation.

merryhime (approved):

merryhime:
> Regarding the bit representation of bools, the C++ standard just says "the value false is converted to zero and the value true is converted to one" [conv.integral]. In theory a compiler could pick any bit representation it wants until you cast to an integral type (including 0 for true and 1 for false), so checking the size alone wouldn't be enough to handle a sufficiently adversarial compiler. 😈

The System V ABI defines that a value of type bool has the value in bit 0 and bits 1 to 7 shall be zero. Other bits are left unspecified. (Page 20 of AMD64 ABI Draft 0.99.7)

PR #213, opened 2021-10-02
recompiler: directly emit code for target ABI

invertego:
Previously, the recompilers in ares would assume the System V calling
convention, then emit extra moves on Windows to handle the differences
in register usage. Now, they directly populate the registers used by the
target system for passing parameters to functions. This is achieved
with register aliases set for the appropriate ABI at compile time.

Furthermore, registers that are nonvolatile only on Windows - rsi and
rdi - are no longer used as temporaries, removing the need to save and
restore them in recompiled code blocks.

This accounts for roughly a 10% reduction in recompiled code size on
Windows while making functionally no difference to other systems.

### DIFF
--- a/ares/component/processor/sh2/recompiler.cpp
+++ b/ares/component/processor/sh2/recompiler.cpp
@@ -44,8 +44,6 @@ auto SH2::Recompiler::emit(u32 address) -> Block* {
   push(rbp);
   push(r13);
   if constexpr(ABI::Windows) {
-    push(rsi);
-    push(rdi);
     sub(rsp, imm8(0x40));
   }
   mov(rbx, imm64(&self.R[0]));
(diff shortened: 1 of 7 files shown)


invertego:
I've checked this change for regressions on Windows with the CPU tests in the following repos:
https://github.com/PeterLemon/N64
https://github.com/PeterLemon/PSX

Note: For the ps1 core, I had to clamp the return value of GPU::vend() to vtotal() - 1 to get those tests to render properly. This is independent of my change.

I also spot checked a variety of games (particularly for the 32x) on Windows and on Ubuntu. The 32x core still defaults to the cached interpreter, so I switched it to the recompiler for my testing.

LukeUsher (approved):

PR #235, opened 2021-10-17
Fix linking on old macOS toolchains

rasky:
Ares JIT recently started to require the data segment to be executable (see  https://github.com/ares-emulator/ares/pull/200 for details on why this was done instead of allocating a new segment via mmap).

To do this, it uses `mprotect()` (`bump_allocator::resize`), but it also needs to tell the linker that the data segment can be upgraded to executable with `-Wl,-segprot,__DATA,rwx,rw` (nall/GNUmakefile). Unfortunately, for quite some time, the clang LD had a bug and didn't accept that option for the data segment.

The official CI has a newer linker without the bug, but for developers compiling ares with an old toolchain, this small tool will patch the binary to mark the data segment as potentially executable.

Notice that the tool does nothing if the binary is correct, so it is only used as fallback.

### DIFF
--- a/desktop-ui/GNUmakefile
+++ b/desktop-ui/GNUmakefile
@@ -68,6 +68,9 @@ all: $(all.objects)
 	$(info Linking $(output.path)/$(name) ...)
 	+@$(compiler) -o $(output.path)/$(name) $(all.objects) $(all.options)
 ifeq ($(platform),macos)
+# Apply workaround for buggy linker in Xcode < 11.4.1
+	@$(compiler) -o $(output.path)/macos-fix-jit macos-fix-jit.cpp
+	$(output.path)/macos-fix-jit $(output.path)/$(name)
 	rm -rf $(output.path)/$(name).app
 	mkdir -p $(output.path)/$(name).app/Contents/MacOS/
 	mkdir -p $(output.path)/$(name).app/Contents/Resources/
--- a/desktop-ui/macos-fix-jit.cpp
+++ b/desktop-ui/macos-fix-jit.cpp
@@ -1,0 +1,119 @@
+// macos-fix-jit - Fix Ares JIT support on macOS with old linkers
+//
+// Ares JIT requires the data segment to be executable (see 
+// https://github.com/ares-emulator/ares/pull/200 for details on why
+// this is done instead of allocating a new segment via mmap).
+//
+// To do this, it uses mprotect() (bump_allocator::resize), but it also
+// needs to tell the linker that the data segment can be upgraded to executable
+// with -Wl,-segprot,__DATA,rwx,rw (nall/GNUmakefile). Unfortunately, for quite
+// some time, the clang LD had a bug and didn't accept that option for the
+// data segment. The bug was fixed in Xcode 11.4.1.
+//
+// The official CI has a newer linker without the bug, but for developers compiling
+// ares with an old toolchain, this small tool will patch the binary to
+// mark the data segment as potentially executable.
+
+#include <stdio.h>
+#include <stdint.h>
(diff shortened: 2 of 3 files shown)


Screwtapello:
The nall build system supports cross-compilation, but it does not support differentiating between the application built for the target platform, and helper tools built for the host platform. That's why the output of the `sourcery` tool is committed to the repo instead of being built from scratch every time — if you tried to cross-compile, it would build `sourcery` for the target instead of the host and everything would break.

This is less of a concern for this tool specifically, because it's only needed for macOS and Apple doesn't let you build macOS binaries on other platforms, so the host and target are always going to be identical. That said, it would be an... untidiness. Near was the kind of person who would be much happier to drop support for old macOS toolchains than to risk untidiness of any form, but maybe that's not what we want to do anymore.

I'll leave the decision up to Luke, I just wanted to provide some background.

rasky:
I’m not sure what is the point. We do have a concrete problem here, and I wasted an afternoon debugging it. 

Do we want to help developers work on Ares or do we want to keep the build system tidy for cross compiling to a platform that does not support cross compilation?

desktop-ui/GNUmakefile
```diff
@@ -68,6 +68,9 @@ all: $(all.objects)
 	$(info Linking $(output.path)/$(name) ...)
 	+@$(compiler) -o $(output.path)/$(name) $(all.objects) $(all.options)
 ifeq ($(platform),macos)
+	# Apply workaround for buggy linker on old macOS toolchains
+	$(compiler) -o $(output.path)/macos-fix-jit macos-fix-jit.cpp
```
invertego: For consistency with the other rules, we may want to suppress the echoing of the compiler invocation using a '@' prefix.

desktop-ui/GNUmakefile
```diff
@@ -68,6 +68,9 @@ all: $(all.objects)
 	$(info Linking $(output.path)/$(name) ...)
 	+@$(compiler) -o $(output.path)/$(name) $(all.objects) $(all.options)
 ifeq ($(platform),macos)
+	# Apply workaround for buggy linker on old macOS toolchains
```
invertego: This comment will be passed to the shell unless the leading whitespace is removed. See https://www.gnu.org/software/make/manual/html_node/Recipe-Syntax.html#Recipe-Syntax

invertego: It may be helpful to clarify in the comment that "old" refers to Xcode < 11.4.1: https://developer.apple.com/documentation/xcode-release-notes/xcode-11_4_1-release-notes

desktop-ui/macos-fix-jit.cpp
```diff
@@ -0,0 +1,119 @@
+// macos-fix-jit - Fix Ares JIT support on macOS with old linkers
+//
+// Ares JIT requires the data segment to be executable (see 
+// https://github.com/ares-emulator/ares/pull/200 for details on why
+// this is done instead of allocating a new segment via mmap).
+//
+// To do this, it uses mprotect() (bump_allocator::resize), but it also
+// needs to tell the linker that the data segment can be upgraded to executable
+// with -Wl,-segprot,__DATA,rwx,rw (nall/GNUmakefile). Unfortunately, for quite
+// some time, the clang LD had a bug and didn't accept that option for the
```
invertego, line 10: As above, it may be helpful to clarify that this was broken until 11.4.1: https://developer.apple.com/documentation/xcode-release-notes/xcode-11_4_1-release-notes

invertego (changes requested):

rasky:
Done, thank you!

PR #245, opened 2021-10-17
desktop-ui: fix file/folder browser dialog alignment

invertego:
Align file/folder dialogs to the settings window when appropriate. The
absence of this alignment was triggering asserts in debug builds.


jbo-85:
Looking good, I fixed that in the old byuu-interface in higan: https://github.com/higan-emu/higan/pull/71, I didn't notice that it apparently didn't survive the switch to ares

invertego:
I wish I had known! I was also just looking at a libco bug, and your comment made me consider that I should check if it's been fixed in the upstream libco repository, and it turns out that it has.

PR #246, opened 2021-10-18
libco: cherry-pick two aarch64 fixes from higan-emu/libco

invertego:
Cherry-pick a correctness fix and an optimization for aarch64 from https://github.com/higan-emu/libco.


LukeUsher:
There might be licensing concerns here due to higan being GPL and ares being ISC; might need to speak to @Sintendo for clearance before we can merge this fix.

Screwtapello:
This is from libco, though, which is ISC like ares is.

Issue #250, opened 2021-10-19
[MD] Double Dragon: Doesn't launch


BiZkViT96:
I hope the developers don't mind that I'm spamming bug reports like this. 

Ares - Black screen
![2021-10-19 20_28_21-Double Dragon (U)  !](https://user-images.githubusercontent.com/88586431/137919659-8a8b42fd-7baf-4e7d-980a-c63ccae4a4a2.png)
Kega - in game
![2021-10-19 20_27_45-Fusion 3 64 - Genesis - DOUBLE DRAGON](https://user-images.githubusercontent.com/88586431/137919739-404c1b7e-37a8-421f-949d-f7f3d38a176c.png)

Double Dragon (U) [!]
SHA1 F054FB2C16A78CE5F5CE20A36BF0F2634F169969

Tested on ares nightly 2021-10-19

FitzRoyX:
Make sure you include the (Unlicensed) or (Proto) tags when a game is of that type. Sometimes that has a lot to do with why a game isn't working.

TascoDLX:
IMO you don't really need to post screenshots for simple issues like this. It's enough to say 'ares boots to black screen, but the rom boots fine in Fusion, BlastEm' or whatever, as that typically doesn't require illustration. If you're going to add anything, please include the full rom name and crc/sha hash if possible so we know exactly what you're testing.

Nevertheless, thank you for the report, this will be fixed shortly.

BiZkViT96:
@TascoDLX Okay, I'll keep that in mind. I hope Ares will replace Kega as a universal emulator for pre 32Bit Sega systems. Thanks.

@FitzRoyX In this case, it is retail.

BiZkViT96:
Fixed with https://github.com/ares-emulator/ares/pull/257

Issue #240, opened 2021-10-17
[MD] Star Cruiser: game freezes on SEGA logo


BiZkViT96:
Ares
![2021-10-18 02_28_27-Star Cruiser (J)  !](https://user-images.githubusercontent.com/88586431/137642228-4eb0cfa2-853b-429e-84bf-eb211fb8849b.png)
BlastEm
![2021-10-18 02_30_41-STAR CRUISER_MD - BlastEm - 58 8 fps](https://user-images.githubusercontent.com/88586431/137642291-613c75f1-c8b8-4f63-a169-c01593ad1242.png)

I tried the Japanese version and with the [translation](https://www.romhacking.net/translations/2776/) - the result is the same.

Star Cruiser (J) [!]
SHA1 CB099ECDE141BEFFDFED6BB7F1D3DC6340DA81D1

Tested on ares nightly 2021-10-17

BiZkViT96:
Fixed with https://github.com/ares-emulator/ares/pull/257

PR #257, opened 2021-10-21
M68000 & MegaDrive related fixes

### LINKED ISSUE #242: [MD] Triple Play Gold: Doesn't launch
Ares - Just black screen
2021-10-18 03_05_16-Triple Play Gold (U)  c !
BlastEm
2021-10-18 03_05_45-Triple Play Gold - BlastEm - 59 9 fps

Triple Play Gold (U) [c][!]
SHA1 007BEE242384DB1887C5831657470584FF77A163

Tested on ares nightly 2021-10-17

### LINKED ISSUE #248: [MD] Arkagis Revolution (Unlicensed): Doesn't launch
Ares - Black screen
2021-10-19 18_00_56-Arkagis Revolution - Trial Version (Rev 00)
Kega
2021-10-19 17_58_36-Fusion 3 64 - Genesis - ARKAGIS REVOLUTION
BlastEm
2021-10-19 18_01_40-ARKAGIS REVOLUTION - BlastEm - 59 9 fps

Kega and BlastEm run game with no problem.

Get trial rom from there website https://arkagis.com/revolution/

Arkagis Revolution - Trial Version (Rev 00)
SHA1 AD2E9E1DFC82F687B9BDB5B972A340990EA577F6

Tested on ares nightly 2021-10-17

### LINKED ISSUE #249: [MD] Danny Sullivan's Indy Heat (Proto): Doesn't launch
Ares - Black screen
2021-10-19 20_14_10-Danny Sullivan's Indy Heat
BlastEm - in game
2021-10-19 20_13_16-Game title here - BlastEm - 59 3 fps
Look for Danny Sullivan's Indy Heat on https://hiddenpalace.org/Main_Page

Danny Sullivan's Indy Heat (USA) (Beta)
SHA1 6E76FF6A74374CD9170A902C250A1674EF69FD56

Tested on ares nightly 2021-10-19

TascoDLX:
1. M68k modifies the interrupt mask based on active priority, but it doesn't do this for exceptions or traps. This likely fixes issues in a number of MD/M-CD games, including The Amazing Spider-man vs. The Kingpin [CD] which now boots.
2. The M68k TAS instruction was purposely borked for the MD cpu, but that doesn't apply to the Mega-CD subcpu (or other systems). Fixes severe graphical glitches in The Adventures of Batman & Robin (CD) and Batman Returns (CD).
3. Fixed M68k trace disassembly of PC-relative addresses to show proper sign-extension.
4. Fixed a masking issue that occurs when a MD cart has SRAM starting at $200000 but ROM extends past $300000. Fixes #242.
5. Tweaked the MD Z80 bus ownership flag so that bus switching appears to be instantaneous because _some games_ want to do very dumb things otherwise (ahem, fixes #248). This should be fairly safe since the bus is extremely likely to be available by the time the next instruction executes (unfortunately, at this time we can't break in the middle of a Z80 instruction based on bus cycle). Also, removed a kludge that artificially delays the bus switch (added to fix Fatal Rewind, but it was errant due to a different issue).
6. Implemented MD M68k open bus lines correctly, at least for shared Z80 & I/O areas. Several games incorrectly check the bus ownership flag, comparing against open bus lines instead. Fixes #249, #250 and others.

Also, #240 is fixed somewhere in there as well, I lost track.

### DIFF
--- a/ares/component/processor/m68000/disassembler.cpp
+++ b/ares/component/processor/m68000/disassembler.cpp
@@ -50,7 +50,7 @@ template<u32 Size> auto M68000::_address(EffectiveAddress& ea) -> string {
   if(ea.mode ==  2) return {_addressRegister(AddressRegister{ea.reg})};
   if(ea.mode ==  5) return {"$", hex(_readDisplacement(read(AddressRegister{ea.reg})), 6L)};
   if(ea.mode ==  6) return {"$", hex(_readIndex(read(AddressRegister{ea.reg})), 6L)};
-  if(ea.mode ==  7) return {"$", hex((i16)_readPC<Word>(), 6L)};
+  if(ea.mode ==  7) { auto imm = _readPC<Word>(); return {"$", imm >= 0x8000 ? hex((i16)imm, 6L, 'f') : hex((i16)imm, 6L, '0') }; }
   if(ea.mode ==  8) return {"$", hex(_readPC<Long>(), 6L)};
   if(ea.mode ==  9) return {"$", hex(_pc + (i16)_readPC(), 6L)};
   if(ea.mode == 10) return {"$", hex(_readIndex(_pc), 6L)};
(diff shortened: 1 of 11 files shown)


LukeUsher:
Looks good overall, just one comment:

in point 5 you state that it is not possible for the z80 to break within an instruction on any given bus cycle, but this should be automatically supported by the overall design of ares.

Currently, apu::step() in the MegaDrive core is implemented as

```c++
auto APU::step(u32 clocks) -> void {
  Thread::step(clocks);
  Thread::synchronize(cpu, vdp, vdp.psg, opn2);
}
```

Which will sync with other components every time step is called

but if instead, that gets refactored to a loop, something like
```c++
auto APU::step(u32 clocks) -> void {
  while(clocks--) {
    Thread::step(1);
    Thread::synchronize(cpu);
  }

  Thread::synchronize(vdp, vdp.psg, opn2);
}
```

This will allow for the m68k to run between every single z80 clock cycle, but still only synchronise the rest of the hardware each time step is called, since they don't necessarily need cycle-granularity against the z80.

If the performance overhead of syncing this often becomes a problem, it could be done conditionally based on the state of the z80.

TascoDLX:
> in point 5 you state that it is not possible for the z80 to break within an instruction on any given bus cycle, but this should be automatically supported by the overall design of ares.

I agree that it's technically possible in ares. However, I feel it would be no more correct given how the z80 is written now, which is not bus-cycle accurate. We can definitely preempt instructions in that regard, but timing would need to be worked out to determine the exact delays before the bus is granted. Best we could do currently is maybe a fixed delay based on an average? Frankly I doubt enough research has been done to implement this with any certainty.

We can definitely file it under 'Future Plans'. If it turns out there are issues with the current implementation (e.g., disproportionate switching delays), we could modify it in the manner you said. However, it would introduce quite a bit more complexity, probably not appropriate at this time. Plenty more bugs to fix before then.

LukeUsher:
Alright, noted

Shideravan:
Very nice update, this fixed a few other bugs in other games too. Thanks!

PR #266, opened 2021-10-24
fc: support for taito tc0190 / tc0690 boards

encoded-byte:
Support added for TC0190 and TC0690 chips found on some Taito boards. These chips are very similar to Nintendo MMC3.

Taito TC0190 (iNes mapper 33):
- Akira (Japan) - Hangs when trying to get ingame
- Bakushou!! Jinsei Gekijou (Japan)
- Bakushou!! Jinsei Gekijou 2 (Japan)
- Don Doko Don (Japan)
- Golfkko Open (Japan)
- Insector X (Japan)
- Operation Wolf (Japan)
- Power Blazer (Japan)
- Takeshi no Sengoku Fuuunji (Japan)

Taito TC0690 (iNes mapper 48):
- Bakushou!! Jinsei Gekijou 3 (Japan)
- Bubble Bobble 2 (Japan)
- Captain Saver (Japan)
- Don Doko Don 2 (Japan)
- Flintstones, The - The Rescue of Dino & Hoppy (Japan)
- Jetsons, The - Cogswell's Caper (Japan)

"Akira" hangs when trying to get ingame. This does not seem to be mapper related but a timing issue, but I need to debug it further.

Comparing to MMC3, TC0690 raises the irq line with some cycles of delay, the exact value is unknown. I'm delaying the irq 6 cpu cycles as it fix shaking on "The Jetsons" (Mesen is doing the same).

Because of its similitudes I was using the existing MMC3 implementation for irq timing, and that uncover a bug: when irqCounter and irqLatch are both zero the irq line won't raise and it should; this was causing hangs on "Captain Saver". It's fixed on this implementation, but the same fix is probably needed on txrom.

"The Flintstones" is known to have erratic irq behavior, and Mesen is applying special fixes for it. It's apparently working in this implementation, but more testing is needed.

### DIFF
--- a/ares/fc/cartridge/board/board.cpp
+++ b/ares/fc/cartridge/board/board.cpp
@@ -31,6 +31,7 @@ namespace Board {
 #include "sunsoft-4.cpp"
 #include "sunsoft-5b.cpp"
 #include "taito-tc0190.cpp"
+#include "taito-tc0690.cpp"
 
 auto Interface::create(string board) -> Interface* {
   Interface* p = nullptr;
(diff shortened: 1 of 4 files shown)


FitzRoyX:
Welcome to project, thanks for working on the NES core.

Issue #261, opened 2021-10-22
[MD] Resistance The Spiral (Demoscene): Doesn't launch


BiZkViT96:
Just black screen. Demo here https://www.pouet.net/prod.php?which=82607

Works fine on BlastEm and Kega.

The_Spiral_by_Resistance_2019_60Hz
SHA1 EBD51FD1FF293D3071D443D59E009F72DA0B4B14

Tested on ares nightly 2021-10-21

LukeUsher:
fixed by #267

Issue #221, opened 2021-10-04
[M-CD] A/X-101 (USA): The emulator crashes after select "start" in main menu


Shideravan:
The emulator crashes on choose the "start" option in main menu
![image](https://user-images.githubusercontent.com/12163923/135792576-926a79cb-a876-411f-b990-97c01c6342f1.png)

Using Windows 10 x64
ares 3ca1f9e
Sega CD BIOS: USA 2.0.0

TascoDLX:
Try a different v2 BIOS (recommend the one from June '93). There a CD h/w revision that isn't emulated correctly due to the way seek ops work. The game seems to work fine otherwise.

Issue #219, opened 2021-10-04
[M-CD] Robo Aleste: Graphics corrupted


Shideravan:
Some sprites are corrupted in Super Robo Aleste
![image](https://user-images.githubusercontent.com/12163923/135789232-41abf0fb-a668-48ef-aafd-9a8e68e89ed1.png)

LukeUsher:
fixed by #267

Issue #112, opened 2021-08-27
[MD] Super Hydlide (J): Doesn't registering the player


Shideravan:
You can normally start the game by selecting the following options (select every option with C button):
* 3rd option (create)
* write any name and choose the last option (end)
* In this screen you choose one of 4 classes. Choose any.
* Will show the status and confirm. Press C in the first option (yes)
* The first option is "start" to start the game. But will appear a message "not registered". Even after registering yourself.
![image](https://user-images.githubusercontent.com/12163923/131074108-9a9a95ae-26aa-4f30-8ae0-46d430a1e6e7.png)

The issue only happens in Japanese version. And it doesn't create a save file, then probably the issue is save related.

LukeUsher:
fixed by #267

PR #267, opened 2021-10-25
Mega-CD fixes + a couple MegaDrive quick fixes

TascoDLX:
1. Squashed a bug in M-CD timer control to eliminate Silpheed read errors.
2. Rewrote M-CD graphics render function, which fixes several issues including #219 & gfx glitches in Battlecorps, Batman Returns.
3. Changed behavior of CDD to allow continuous playback (BIOS handles pausing when applicable). Fixes delay issue in Willy Beamish intro (& general CDDA playback).
4. Report error on unhandled CDD command, which is stopgap to avoid app crashes when using unsupported BIOS version. This should resolve the crash in #221 (though this issue is fully resolved when using a compatible v2 BIOS).
5. Proper region autodetect on discs via identifying boot code, specifically for cases where the disc header is incorrect.
6. Added support for the 3-button pad on M-CD and set it to default as in #197. Some games have incompatibility (e.g., Battlecorps char select) or exhibit slowdown in ares due to improper polling patterns (Willy Beamish, Rebel Assault).
7. Fixed a minor issue in VDP interrupt ack which occurs when H-INT is enabled but V-INT is disabled. This fixes glitches in one scene of The Spiral demo (ref: #261, this is fully resolved now).
8. Specific fix for #112 which is missing SRAM info in its header.

### DIFF
--- a/ares/md/mcd/cdd.cpp
+++ b/ares/md/mcd/cdd.cpp
@@ -242,6 +242,8 @@ auto MCD::CDD::process() -> void {
     io.status = Status::Playing;
   } break;
 
+  default:
+    io.status = Status::CommandError;
   }
 
   status[0] = io.status;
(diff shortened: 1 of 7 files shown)


LukeUsher (approved):
Nice job, as always

PR #283, opened 2021-11-10
mia: prefer known rom extensions in zip archives

### LINKED ISSUE #260: [MD] Titan MD-NICCC (Demoscene): Doesn't launch from zip archive
Just black screen. Get demo here https://www.pouet.net/prod.php?which=81027

Works fine on BlastEm. Demo has accuracy check and Kega fails on it.
2021-10-22 19_02_49-Fusion 3 64 - MegaDrive - MD-NICCC

titan-md-niccc
SHA1 8A200F3EEEC641DFD378780F55CE523F190A7A1E

Tested on ares nightly 2021-10-21

invertego:
Instead of just reading the first file from a zip archive, prefer files
that match known rom extensions. Extra files such as readmes should not
interfere with the loading of the game.

Fixes #260

### DIFF
--- a/mia/pak/pak.cpp
+++ b/mia/pak/pak.cpp
@@ -10,6 +10,18 @@ auto Pak::name(string location) const -> string {
   return Location::prefix(location);
 }
 
+auto Pak::read(string location) -> vector<u8> {
+  //attempt to match known extensions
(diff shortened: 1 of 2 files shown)


LukeUsher:
nice fix, thanks!

PR #301, opened 2021-11-16
fc: support for bandai 74x161 boards

encoded-byte:
Suppor added for discrete logic Bandai boards.

Bandai 74x161 (iNES mappers 70/152):
- Arkanoid II (Japan)
- Family Trainer 4 - Jogging Race (Japan) *Required peripheral: Family Trainer Mat
- Family Trainer 5 - Meiro Daisakusen (Japan) *Required peripheral: Family Trainer Mat
- Family Trainer 6 - Manhattan Police (Japan) *Required peripheral: Family Trainer Mat
- Gegege no Kitarou 2 - Youkai Gundan no Chousen (Japan)
- Kamen Rider Club (Japan)
- Pocket Zaurus - Juu Ouken no Nazo (Japan)
- Saint Seiya - Ougon Densetsu (Japan)
- Space Shadow (Japan) *Required peripheral: Bandai Hyper Shot

In this pr I also split the implementation of Bandai FCG and LZ93D50; those chips are very similar, but I think they deserve different files for a more readable code and to describe its differences better.

Note that saving is broken on LZ93D50 games with X24C02 eeprom (about half the games on mapper 16). In its current form they are using a X24C01, this eeprom (besides of having half the size) is functionally different from X24C02. In fact "Dragon Ball Z III" is booting "by accident": the game sends commands on boot that the eeprom misunderstands, and its not ending in standby status by pure luck (that would produce an infinite loop in the game code). It's possible that these games don't work at all or hang on saving, "SD Gundam Gaiden 3" hangs on starting game probably related to this. Just using a M24C02 (already emulated) didn't work, I will look into this issue.

### DIFF
--- a/ares/fc/cartridge/board/bandai-74161.cpp
+++ b/ares/fc/cartridge/board/bandai-74161.cpp
@@ -1,0 +1,64 @@
+struct Bandai74161 : Interface {
+  static auto create(string id) -> Interface* {
+    if(id == "BANDAI-74161" ) return new Bandai74161(Revision::B74161);
+    if(id == "BANDAI-74161A") return new Bandai74161(Revision::B74161A);
+    return nullptr;
+  }
+
+  Memory::Readable<n8> programROM;
+  Memory::Readable<n8> characterROM;
+
+  enum class Revision : u32 {
+    B74161,
+    B74161A,
+  } revision;
+
+  Bandai74161(Revision revision) : revision(revision) {}
+
+  auto load() -> void override {
+    Interface::load(programROM, "program.rom");
+    Interface::load(characterROM, "character.rom");
+    mirror = pak->attribute("mirror") == "vertical";
+  }
+
+  auto readPRG(n32 address, n8 data) -> n8 override {
+    if(address < 0x8000) return data;
+    n3 bank = (address < 0xc000 ? programBank : (n3)0x7);
+    return programROM.read(bank << 14 | (n14)address);
+  }
+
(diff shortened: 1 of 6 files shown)


ares/fc/cartridge/board/bandai-74161.cpp
```diff
@@ -0,0 +1,64 @@
+struct Bandai74161 : Interface {
+  static auto create(string id) -> Interface* {
+    if(id == "BANDAI-74161" ) return new Bandai74161(Revision::B74161);
```
sp1187, line 3: NesCartDb calls this mapper `BANDAI-74*161/161/32`, would that perhaps be preferable?

encoded-byte: Maybe, but note that there are more boards in ares right now with names different than those on NesCartDb. If NesCartDb names are preferable I can change it.

sp1187: As this name system is often used by other emulators (MAME, Mesen, Nestopia being some of the examples), I personally think it would be good if this system could be followed as much as reasonably possible.

FitzRoyX: I wouldn't get too hung up on naming for now. I've been looking at nescartdb for years and it's not an easy thing to come up with a consistent solution for. Company+chips does seem the most straightforward... but there are a few boards with like 10 discrete logic chips on them and that would be a pretty long name. As long as boards are being appropriately split due to differences, we can always come back to names.

PR #328, opened 2022-01-03
n64: fix RSP halt condition to be more accurate

rasky:
* Fix usage of break opcode in a delay slot. After unhalt, the
   RSP continues execution, taking the jump if needed. So the
   branch state should not be affected by the break opcode.
 * Reading from SP_PC produces garbage (open bus?) when the RSP is
   running. This fix is unlikely to fix games but may help homebrew
   developers.
 * Writing to SP_PC resets the RSP MIPS pipeline, so any pending branch
   is canceled.


LukeUsher (approved):

Shideravan:
Nice! Thank you!

PR #332, opened 2022-01-13
Update terminal.hpp

thar0x29a:
Allow manual controll about terminal colors.

### DIFF
--- a/nall/terminal.hpp
+++ b/nall/terminal.hpp
@@ -4,59 +4,56 @@
 
 namespace nall::terminal {
 
-inline auto escapable() -> bool {
-  #if defined(PLATFORM_WINDOWS)
-  //todo: colors are supported by Windows 10+ and with alternate terminals (eg msys)
-  //disabled for now for compatibility with Windows 7 and 8.1's cmd.exe
-  return false;
-  #endif
-  return true;
-}
+#if defined(PLATFORM_WINDOWS)
+bool escape_colors = false;
+#else
+bool escape_colors = true;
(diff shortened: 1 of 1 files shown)


nall/terminal.hpp
```diff
@@ -4,59 +4,56 @@
 
 namespace nall::terminal {
 
-inline auto escapable() -> bool {
-  #if defined(PLATFORM_WINDOWS)
-  //todo: colors are supported by Windows 10+ and with alternate terminals (eg msys)
```
LukeUsher, line 9: This comment shouldn't be lost, but instead reworded into a note instead of a todo: explaining why colours are disabled by default on Windows

nall/terminal.hpp
```diff
@@ -4,59 +4,56 @@
...
-  //disabled for now for compatibility with Windows 7 and 8.1's cmd.exe
-  return false;
-  #endif
-  return true;
-}
+#if defined(PLATFORM_WINDOWS)
+bool escape_colors = false;
```
LukeUsher, line 8: Perhaps this should be renamed something like "enable_colors" or "use_color", making it's purpose more clear

LukeUsher (changes requested):

nall/terminal.hpp
thar0x29a: Thanks for the feedback. I think I will rework this even more and include true color support as well ..

LukeUsher:
This PR has been stale for long enough I think

PR #334, opened 2022-01-17
n64: fix mult/div opcode timings

rasky:
Taken from NEC VR4300 manual. These are easier than expected to
implement accurately as the CPU doesn't execute them in parallel but
stalls the entire pipeline while they are running.


ares/n64/cpu/interpreter-ipu.cpp
```diff
@@ -248,6 +248,7 @@ auto CPU::DDIV(cr64& rs, cr64& rt) -> void {
     LO.u64 = rs.s64 < 0 ? +1 : -1;
     HI.u64 = rs.s64;
   }
+  step(69);
```
Kawa-oneechan, line 251: Nice.

Shideravan:
Nice! Thank you!

PR #353, opened 2022-02-12
[N64] Change PI DMA to use 16 bit fetches

### LINKED ISSUE #351: [N64] Paper Mario (USA) - Creating a save file broken, skips new game cutscene
Tested on ares v126 and latest master

CasualPokePlayer:
Using the read<Byte> currently in use breaks DMAs from Flash (breaking Paper Mario, probably others), which always return 0. I assume 8 bit reads are just not allowed for Flash, although I can't verify this myself.

Checking https://github.com/n64dev/cen64/commit/8367698e20cca7eb262779f19c8c7847ae31bb29 comments, it seems that fetches should always be 16 bits anyways, so I presume this is the correct behavior, and fixes DMAs from Flash. Would like @rasky's opinion on this as this might be complete bullshit interpreting his work from me.

Noting no regressions with the original testrom have occurred.

Resolves https://github.com/ares-emulator/ares/issues/351

### DIFF
--- a/ares/n64/pi/dma.cpp
+++ b/ares/n64/pi/dma.cpp
@@ -27,8 +27,12 @@ auto PI::dmaWrite() -> void {
     if (length.bit(0)) length += 1;
 
     i32 rom_len = (cur_len + 1) & ~1;
-    for (u32 i = 0; i < rom_len; i++)
-      mem[i] = bus.read<Byte>(io.pbusAddress++);
+    for (u32 i = 0; i < rom_len; i += 2) {
+      u16 data = bus.read<Half>(io.pbusAddress);
+      mem[i + 0] = data >> 8;
+      mem[i + 1] = data & 0xFF;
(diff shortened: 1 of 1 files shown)


rasky:
The PR looks correct to me. The accesses made by the PI during DMA on the cartridge side are always 16-bit, so this PR matters if Ares emulates correct access size behaviors. I didn’t realize this does already happen for flashes.

I can’t easily test what a byte size access on a real flash chip does (I could do that through emulated flashes on development cartridges, but it might not identical in these details), so we’ll trust what makes games work for now.

Byte sized accesses to ROM has a weird address scrambling behavior (it works in the first two bytes in each 32-bit word, and does an address bitmask on the other two bytes) and I believe this is how PI behaves, before the address lines reach ROM. So I guess this pattern should be emulated first also for flash, and then we could check what a flash chip does with a 8bit access.

PR #361, opened 2022-02-17
md: fix Triple Play Gold

rasky:
This game is unusual because the SRAM is always mapped (by default, with
no way to turn it off), even though the ROM is bigger than 2Mb.
A portion of the ROM seems to be always shadowed, though it's unclear
whether the ROMs contain FF data in that area, or simply the ROM mapper
skips it. There is related confusion in the existing dumps, as most
dumps seem to contain SRAM contents in the 0x200000 area, as expected.

### DIFF
--- a/ares/md/cartridge/board/banked.cpp
+++ b/ares/md/cartridge/board/banked.cpp
@@ -10,7 +10,7 @@ struct Banked : Interface {
   auto load() -> void override {
     Interface::load(rom, "program.rom");
     if(auto fp = pak->read("save.ram")) {
-      Interface::load(sramAddr, sramSize, wram, uram, lram, "save.ram");
+      Interface::load(sramAddr, sramSize, ramAlwaysEnabled, wram, uram, lram, "save.ram");
     }
     if(auto fp = pak->read("save.eeprom")) {
       Interface::load(m24c, "save.eeprom");
@@ -107,7 +107,7 @@ struct Banked : Interface {
       return;
     }
 
-    if(address == 0xa130f0) {
+    if(address == 0xa130f0 && !ramAlwaysEnabled) {
       ramEnable   = data.bit(0);
       ramWritable = data.bit(1);
     }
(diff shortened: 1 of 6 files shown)


LukeUsher:
I'm holding back on merging this until we get clarification on how the hardware works, which rom dumps are good/bad, etc

I don't think there's enough information to make the correct implementation decision right now.

ekeeke:
See:
https://datomatic.no-intro.org/index.php?page=show_record&s=32&n=1457
https://datomatic.no-intro.org/index.php?page=show_record&s=32&n=1458

for the PCB layout of these two games (Triple Play Gold and Triple Play 96) and available ROM dumps.

As explained in above post, the games only have a 3MB ROM chip but it is mapped discontinuously in the CPU address space:
- lower 2MB is mapped to 0x000000-0×1FFFFF
- upper 1MB is mapped to 0x300000-0x3FFFFF

The games also embedd backup RAM which is mapped in between, in the usual 0x200000-0x2FFFFF region.

The result is that ROM dumps done from the cartridge port (with devices like Retrode or similar MD cartridge dumpers) are 4MB files with SRAM data included (usually 0xFF when coming out from factory but can be anything depending on backup memory state when dumped). These ROM dumps are playable in emulators without specific care but will only allow backup saves if you enable SRAM in the middle region as specified in ROM header.

The 3MB ROM dumps are coming straight from the Mask ROM and will need to be properly mapped by the emulator to be playable.

Both dumps are available in no-intro.

EDIT: To be fully accurate, the Mask ROM chip is actually a 4MB ROM chip (ROM chip with size not a power of 2 do not exist afaik) but the upper MB is empty (reads as 0xFF) and is not mapped in CPU address space.

LukeUsher:
> EDIT: To be fully accurate, the Mask ROM chip is actually a 4MB ROM chip (ROM chip with size not a power of 2 do not exist afaik) but the upper MB is empty (reads as 0xFF) and is not mapped in CPU address space.

This is what I expected, thanks for confirming; closing this PR for now, and I will implement support for this specific cart board. Given that ares usually has a separate board implementation for each type of cartridge, modifying the standard SEGA linear mapped board to support this doesn't feel like the right solution.

Issue #212, opened 2021-10-01
[MD] Titan Overdrive 1 graphical glitches


essasetic:
### Ares:
![Screenshot-20211001190409-1920x1160](https://user-images.githubusercontent.com/34970796/135667444-3edb0e12-8218-4af1-a95d-40933e936346.png)
![Screenshot-20211001190612-1920x1160](https://user-images.githubusercontent.com/34970796/135667443-8fb65921-d1d7-4005-9af7-e9af8dfdf745.png)
![Screenshot-20211001190645-1920x1160](https://user-images.githubusercontent.com/34970796/135667442-605da3fa-0786-4bab-91e0-16f58c727708.png)
![Screenshot-20211001190845-1920x1160](https://user-images.githubusercontent.com/34970796/135667439-c4dbac4c-f4b8-4d79-b0f2-c441ebf12e4b.png)

### Original hardware:

![Screenshot from 2021-10-01 19-12-26](https://user-images.githubusercontent.com/34970796/135667850-c2c6d46f-8b38-4cb6-8f49-6ec4437f108f.png)
![Screenshot from 2021-10-01 19-12-45](https://user-images.githubusercontent.com/34970796/135667852-79f048b3-57f9-4b22-ba49-6f65356f08eb.png)
![Screenshot from 2021-10-01 19-13-09](https://user-images.githubusercontent.com/34970796/135667856-cca0d34a-d343-463e-b09c-c4fc1f32ed8f.png)
![Screenshot from 2021-10-01 19-13-50](https://user-images.githubusercontent.com/34970796/135667858-6e4c4ae7-9081-425a-a4d1-82c29cbf29f9.png)

NOTE: I haven't tested the emulator specific versions on ares. Only the one meant for the original hardware as although the emulated versions will probably work perfectly. It isn't particularly useful for making the emulator more accurate.

essasetic:
UPDATE:
Did a re-test on Ares v126. Unfortunately nothing got fixed however there are updates on two of the bugs I brought up.

![Titan Overdrive 1](https://user-images.githubusercontent.com/34970796/144718557-9f9d98b0-44a5-4884-b37c-c01f3f14e3c0.png)

The text on the left doesn't even render garbled mess anymore but at least more of the robot is render (I guess)?


![Titan Overdrive 2](https://user-images.githubusercontent.com/34970796/144718586-830eca1d-0c30-435c-8d55-ccd25ef0e1dc.png)

[...]

PR #401, opened 2022-02-21
md: fix vblank bit toggling horizontal timing

rasky:
Currently, the vblank bit is turned on/off at the same time vcounter
is incremented, which is a reasonable assumption, missing any real
data on this.

But it turns out this is responsible for the breakage of Overdrive 1.
I did some tests and it looks like anything in the range >= 0xF0 (same
line), or <= 0x80 (on next line) fixes those bugs. I decided to
bind it to vedge(), which means that vblank bit now is triggered the
same instant that VINT is raised, which also makes sense when you
think of it.

Updates #212

### DIFF
--- a/ares/md/vdp/main.cpp
+++ b/ares/md/vdp/main.cpp
@@ -8,19 +8,30 @@ auto VDP::tick() -> void {
   cycles += 2;
   state.hcounter++;
   if(h32()) {
-    if(hcounter() == 0x00) hblank(0), vedge();
+    if(hcounter() == 0x00) hblank(0), vblankcheck(), vedge();
     if(hcounter() == 0x81) vtick();
     if(hcounter() == 0x93) hblank(1);
     if(hcounter() == 0x94) state.hcounter = 0xe9;
   }
   if(h40()) {
-    if(hcounter() == 0x00) hblank(0), vedge();
+    if(hcounter() == 0x00) hblank(0), vblankcheck(), vedge();
     if(hcounter() == 0xa1) vtick();
     if(hcounter() == 0xb3) hblank(1);
     if(hcounter() == 0xb6) state.hcounter = 0xe4;
   }
   irq.poll();
   vram.refreshing = 0;
+}
+
+auto VDP::vblankcheck() -> void {
+  if(v28()) {
+    if(vcounter() == 0x0e0) vblank(1);
+    if(vcounter() == 0x1ff) vblank(0);
+  }
+  if(v30()) {
+    if(vcounter() == 0x0f0) vblank(1);
+    if(vcounter() == 0x1ff) vblank(0);
+  }
 }
 
 auto VDP::vtick() -> void {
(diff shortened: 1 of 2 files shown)


LukeUsher:
This PR appears to cause a regression with some 32X titles; Knuckles Chaotix no longer boots up, instead, it hangs with a black screen.

LukeUsher:
> This PR appears to cause a regression with some 32X titles; Knuckles Chaotix no longer boots up, instead, it hangs with a black screen.

my bad; this was caused by the pr branch not being rebased on master; I'll make sure to rebase during testing in future

ekeeke:
For the record, all the data related to VDP events timings are summarized in this post from Nemesis (Exodus emulator author): http://gendev.spritesmind.net/forum/viewtopic.php?f=22&t=1291&start=17

As you can see, initial assumption was correct: VBLANK bit is set at the same time as Vcounter is incremented, which is also at the same time that Hint is triggered. Vint is triggered slightly later on the line.

rasky:
@ekeeke I don't see anywhere in any of the Nemesis tables posted over the years information about horizontal timing of VBLANK bit. If I set it at the same time of VCounter increment, many things break, and Blastem also agrees with this. It looks like there are still at least 2 non-vblank access slots performed after vcounter increment, so there must be a delay between vcounter increment and vblank bit.

In this PR, the delay is too much and breaks Gauntlet IV. In #404, I reduced it a bit and also made it continuous over the line so to allow for V30->V28 changes on line 224.

ekeeke:
It's indicated in the second table:

```
	//|VBlank set    |VCounter changes |                      |VCounter changes |                 |
	//|                     |from 0x0DF to    |     <Same>      |from 0x0EF to    |       <Same>      |
	//|                     |0x0E0 in [1].    |                             |0x0F0 in [1].    |                 |
```

There are a lot of things that can cause games to break unfortunately.

For example, the HCounter value used in ares to trigger VCounter increment seems a little off: it should be 0x85 in H32 and 0xA5 in H40 (not 0x81 and 0xA1). This is easily verified using a test ROM that read the HV Counters continuously.

rasky:
That is the *vertical* timing of when VBlank is set. It does happen on line 0xE0 / 0xF0. It does read as "vblank bit is set the instant the vcounter changes" (which is the assumption I mentioned), but I think this is an unfortunate copy-and-paste wording from other lines of the same table, or simply something that wasn't fully checked because it was being measured *vertically* not *horizontally*. We're speaking of 10-20 cycles of delay so unless Nemesis was looking exactly there, this is easy to miss.

Also, Kabuto found out that the vblank bit can be set *at any point in line 0xE0* if you switch from V30 to V28. If the check was bound to a specific instant (eg: at vcounter increment), then the vblank bit would not get set when switching later in the line. Notice that if you do switch from V30 to V28 during line 0xE1, vblank is never set, which means that the hardware does check for this condition over a period of multiple cycles.

ekeeke:
It was definitively measured "horizontally" from what I remember. He did continuous reads from VDP status to see when bits were changing using logical analyzer and HSYNC /IPL signals as  references for both Hcounter and VDP status reads.

>All the measurements about the alignment of the status register flags, sync lines, and interrupt events, have been sampled by me though, and have been confirmed to an exact SC cycle.


>Also, Kabuto found out that the vblank bit can be set at any point in line 0xE0 if you switch from V30 to V28. If the check was bound to a specific instant (eg: at vcounter increment), then the vblank bit would not get set when switching later in the line. Notice that if you do switch from V30 to V28 during line 0xE1, vblank is never set, which means that the hardware does check for this condition over a period of multiple cycles.

What Kabuto meant was that is is possible to enter VBLANK anytime in line 0xE0 **as soon as you switch back to V28** if you were in V30. This actually shows that the line counter is checked constantly (or every two cycles maybe) and compared against a given limit depending on the mode (0xE0 for V28, 0xF0 for V30) to enter VBLANK. So, when you are in V28, as soon as the VCounter increments to the limit (0xE0), the VDP detects it and sets VBLANK. Maybe it is a few cycles apart but this is definitively not when VINT occurs, which is later in that line (I am also pretty sure there are some games that expect a significant delay between VBLANK being set and VINT being triggered).

EDIT: from my own notes, the following games have some issues if there isn't enough delay between VBLANK flag being set in VDP status and VINT occurrence
Bram Stocker's Dracula
Out Runners
VR Troopers

PR #406, opened 2022-02-22
md: fix APU port in

### LINKED ISSUE #286: [MD] Thunder Force IV (Europe) Sounds worse compared to US version
The US version of this game seems to sound just fine, but the sound quality goes downhill as soon as I try the European version, tested it out in Blastem and it sounds fine there.  I also noticed that the sound quality immediately gets better as soon as you reset it once.

Tested on Version 125

rasky:
It's technically unused, but it must return 0xFF. Looks like some
game erroneously read it anyway.

Fixes #286

### DIFF
--- a/ares/md/apu/bus.cpp
+++ b/ares/md/apu/bus.cpp
@@ -89,7 +89,7 @@ auto APU::writeExternal(n24 address, n8 data) -> void {
 
 auto APU::in(n16 address) -> n8 {
   //unused on Mega Drive
-  return 0x00;
+  return 0xff;
 }
 
 auto APU::out(n16 address, n8 data) -> void {


ares/md/apu/bus.cpp
```diff
@@ -89,7 +89,7 @@ auto APU::writeExternal(n24 address, n8 data) -> void {
 
 auto APU::in(n16 address) -> n8 {
   //unused on Mega Drive
-  return 0x00;
+  return 0xFF;
```
LukeUsher: Minor nitpick: it is code convention in ares to use lower case for constants like this

LukeUsher (changes requested):

ares/md/apu/bus.cpp
rasky: fixed

PR #413, opened 2022-02-24
n64: invalidate recompiler on icache invalidation

rasky:
The test ROM https://github.com/Mr-Wiseguy/N64-Cache-Emulation-Tests
changes the body of an already-run function, without invalidating
the icache, and verifies that the previous body is run. This was
failing on Ares with the recompiler because we currently invalidate
the recompiled functions on memory writes.

This commit changes the behavior to invalidate recompiled functions
only on icache explicit invalidations (via opcode CACHE). It fixes
the test ROM, as the behavior is now consistent with the expectation
of the cache opcode to actually matter.

Notice that we are trading one not fully conformant behavior
with another one, because after this commit a game that loads code
into memory and forgets to invalidate the icache will always fail,
even though it might work by chance on a real system. On the other
hand, icache address conflics are pretty random especially during
development when linkers shuffle around functions continuously, so it
is really hard to get away with forgetting icache invalidation.
Relying on this bug is extremely unlikely, I think.

### DIFF
--- a/ares/n64/cpu/interpreter-ipu.cpp
+++ b/ares/n64/cpu/interpreter-ipu.cpp
@@ -116,6 +116,7 @@ auto CPU::CACHE(u8 operation, cr64& rs, s16 imm) -> void {
 
   case 0x00: {  //icache index invalidate
     auto& line = icache.line(address);
+    cpu.recompiler.invalidate(address);
     line.valid = 0;
     break;
   }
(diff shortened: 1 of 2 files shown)


ares/n64/cpu/interpreter-ipu.cpp
```diff
@@ -138,7 +139,13 @@ auto CPU::CACHE(u8 operation, cr64& rs, s16 imm) -> void {
 
   case 0x10: {  //icache hit invalidate
     auto& line = icache.line(address);
-    if(line.hit(address)) line.valid = 0;
+    // Invalidate recompiler also on non-hits. icache invalidate is a signal
+    // that something was changed in memory, even though the memory is not in
+    // the icache anymore.
+    cpu.recompiler.invalidate(address);
```
LukeUsher, line 145: Consider moving this recompiler invalidation block to the 'miss; case in https://github.com/ares-emulator/ares/blob/dc5ff13b5ca61e61a4057c0a4de143440181909d/ares/n64/cpu/icache.cpp#L47

This would ensure that the host recompiler cache is cleared when the n64 icache suffers a cache miss; so it would catch scenarios that just using the cache opcode would not.

LukeUsher (changes requested):

ares/n64/cpu/interpreter-ipu.cpp
LukeUsher: turns out I may have been incorrect here; The step function is called after the opcode is implemented, so it's too late to invalidate

PR #416, opened 2022-02-27
sfc: runtime pixel accurate PPU setting

invertego:


LukeUsher:
May I suggest inverting the switch? 

What if we default to the performance cores for pce and sfc, and then the option can then be called something like "Pixel Accuracy Mode", with a description of "Use pixel-accurate emulation where available"

This way we are fast out of the box, but the option is clear in intent and description; for those who really want perfect Air Strike Patrol.

In an ideal world, we wouldn't need to maintain two separate implementations of some components, but the fact of the matter is that the accuracy ppu for sfc just can't reach full-speed when enhancement chips are involved; meaning we need to keep both available. 

Since the MegaDrive performance vdp is currently rather broken, I wouldn't tie that one to this setting for now;

ares/sfc/ppu-performance/io.cpp
```diff
@@ -1,5 +1,5 @@
 auto PPU::latchCounters() -> void {
-  cpu.synchronize(ppu);
+  cpu.synchronize(*this);
```
LukeUsher: in all these cases that pass a PPU pointer to the member function, would it be possible to somehow do this via the constructor instead? 

That way we can avoid modifying all these callers and keep the code looking a little more neat/tidy

ares/sfc/ppu-performance/ppu.hpp
```diff
@@ -1,6 +1,7 @@
-struct PPU : Thread, PPUcounter {
+#define PPU PPUPerformance
+
+struct PPU : PPUBase::Implementation, PPUcounter {
   Node::Object node;
-  Node::Video::Screen screen;
```
LukeUsher, line 3: Is it necessary to move screen to the impl? Can't the impl access the screen member on the parent class; this prevents all the cases of having to replace screen with screen(), and gives consistency with other cores.

ares/sfc/ppu/ppu.hpp
```diff
@@ -1,9 +1,50 @@
-#if defined(PROFILE_PERFORMANCE)
-#include "../ppu-performance/ppu.hpp"
-#else
-struct PPU : Thread, PPUcounter {
+struct PPUBase {
+  struct Implementation : Thread {
```
LukeUsher, line 2: Is there a reason we need a separate Implementation type below PPUBase? Couldn't we just have a PPU class that PPUperformance and PPUaccuracy inherit from?

Or is that too-high performance overhead?

LukeUsher (commented):
Overall I like the idea, but I have a few questions and comments on the chosen implementation.

ares/sfc/ppu/ppu.hpp
invertego: The ppu global would need to become a pointer, putting it at odds with every other component in the system.

ares/sfc/ppu-performance/ppu.hpp
invertego: Are you asking if screen can be moved to PPUBase? You'd have the opposite problem in that case: instead of changing PPU-external references, you'd have to change PPU-internal references. I guess that could be alleviated by putting another member in PPU: `Screen& screen{ppu.screen};`

Functionally, that should behave the same as what is there now, though putting screen in PPUBase would obscure the fact that it is owned by the PPU implementation, which is responsible for its creation and destruction.

ares/sfc/ppu-performance/io.cpp
invertego: Is this comment intended for this line of code? This line is calling Thread::synchronize(), and I don't know how you could avoid passing arguments to it.

ares/sfc/ppu/ppu.hpp
LukeUsher: If there’s little-to-no performance impact, that may actually be a cleaner solution overall?

invertego: PPUBase is an abstraction that hides the fact that there are two PPU implementations from other parts of the system that don't need to be aware of this detail. I'd say it makes the rest of the code cleaner, but mechanically speaking it would not be difficult to replace all the instances of "ppu." with "ppu->" and "ppu" with "*ppu". It would just make code that performs the same action on multiple components (like System::load and unload) slightly irregular.

ares/sfc/ppu-performance/io.cpp
invertego: I rolled back the changes that removed references to globals.

ares/sfc/ppu/ppu.hpp
LukeUsher: I'm probably just nitpicking, but whatever we do won't be as clean as how everything works with the compile-time PPUs, and there's not really a way to solve that; I do think we need runtime switching as an option, it just makes the code ugly.

But we may have to live with that.

ares/sfc/ppu-performance/ppu.hpp
LukeUsher: disregard

invertego: As an aside, at one point I did look at moving screen (and all the other Node elements) to PPUBase and moving the load/unload implementations up into PPUBase as well. This would allow toggling between performance and accuracy between resets instead of requiring an unload/reload. I hit a snag with the Debugger load/unload, which was not trivial to reconcile between the two implementations.

ares/sfc/ppu/ppu.hpp
invertego: It's worth thinking critically about the design if we're going to repeat this concept with other components.

LukeUsher:
@invertego would you like to apply this change to PC-Engine also, or would you rather I take that on?

PR #420, opened 2022-03-04
Add Linux manual build steps

tnibert:
I noticed on the Ares website download section, under Linux it states "It is preferred that Linux users compile from source code" and does not provide an official build.  However there are no instructions readily available for how to compile on Linux.  I've added a basic compile script for Linux, as well as usage instructions in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,3 +16,9 @@ Nightly Builds
 
 Automated, untested builds of ares are available for Windows and macOS as a [pre-release](https://github.com/higan-emu/ares/releases/tag/nightly). 
 Only the latest nightly build is kept.
+
+Manual Build (Linux)
+--------------------
+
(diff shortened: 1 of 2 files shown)


LukeUsher:
I don’t really like the addition of a built script, makefile is already that; changing directory into desktop-ui and simply running “make” is enough, you can optionally add -j to your liking. Additionally defaulting to the number of professors with -j is potentially risky; I’ve brought down a machine due to ram exhaustion in the past, due to having many cores and little ram. 

Other than that there are a ton of build options; some of the software you listed as dependencies don’t have to be; you can enable/disable certain drivers, compile against qt, gtk2 or gtk3, etc, there is a lot of user choice. 

local=false is a bad idea for local building; as you're instructing ares to build for a generic target, rather than being optimised for your machine; negating a large benefit of self compilation. 

It would be better to write a proper build guide documenting all the different options available.

tnibert:
That's fair enough.  Thanks for the additional context.  When I have some free time I'll look through the makefiles and look to put together some more thorough documentation.

Issue #422, opened 2022-03-05
Request: Build Instructions


shinra-electric:
It took me quite a while to figure out how to build Ares. I think the project should post some build instructions somewhere. 

FWIW, here are the steps I took to build on Mac / macOS / M1 / Apple Silicon: 

Homebrew dependencies: `llvm` (There may be others, comment below if you came across some).

`git clone https://github.com/ares-emulator/ares`

`cd ares`

`make -C desktop-ui` 

After compiling, the app should be located in ares/desktop-ui/out/ and it needs to be signed. 

You can self-sign with `codesign --force --deep --sign - ` and then drop the app into terminal to get the path. (Note there is a space after the last `-`).

Screwtapello:
According to the comment in the source, `macos-fix-jit` is a workaround for people building ares with versions of XCode before 11.4.1:

https://github.com/ares-emulator/ares/blob/d5dcc645e42156cd753084f5c1fa16f1d45d8e7d/desktop-ui/macos-fix-jit.cpp#L3-L15

If you build on macOS, the fixer tool is automatically built and applied to the ares binary, you don't need to apply it manually:

https://github.com/ares-emulator/ares/blob/d5dcc645e42156cd753084f5c1fa16f1d45d8e7d/desktop-ui/GNUmakefile#L77-L80

Also, ares should be buildable just by having XCode installed, you shouldn't need Homebrew or any Homebrew packages.

shinra-electric:
Thanks for the info, I'll update the issue accordingly. 

Regarding homebrew, I had llvm removed as I was having issues with another project. I have Xcode 13.2.1 installed, but I was not able to compile ares. After re-installing llvm it worked, which is why I have it listed as a dependency. Maybe if you have llvm installed via homebrew you could try uninstalling and checking?

rasky:
@shinra-electric 

We added a top level Makefile, so now `make` is enough to build Ares. Also we added a codesigning step for app bundle which should fix the issue running the bundle on Apple Silicon. As explained by Screwtapello, macos-fix-jit is used internally by the build system and should not be manually run by users trying to compile Ares.

So now, as long as you have a reasonable C++ compiler installed on your system, running `make` after checkout should be enough to build Ares. I would say the concern of this issue should now be solved.

LukeUsher:
Closing this issue due to new top-level makefile;

shinra-electric:
I can confirm the fix worked for me, and also that it compiled without llvm installed. 

[...]

PR #426, opened 2022-03-05
Add top-level makefile and codesign binary on Mac

rasky:
The top-level Makefile should cover most concerns of #422 and in general removes confusion from people first building ares. New targets can be added if we need to build multiple UIs in the future.

Codesigning on Mac is required on Apple Silicon to run the app bundle

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,0 +1,6 @@
+all: desktop
+
+desktop:
+	$(MAKE) -C desktop-ui
+
+.PHONY: all desktop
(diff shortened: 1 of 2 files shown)


LukeUsher:
since this calls make a second time, will it still correctly pass any build options passed? eg `cores=md` or `hiro=gtk3`

GNUmakefile
```diff
@@ -0,0 +1,6 @@
+all: desktop
+
+desktop:
+	$(MAKE) -C desktop-ui
```
LukeUsher, line 4: nit: we tend to use two spaces rather than a tab for indentation

rasky:
@LukeUsher yes, that's standard GNU make behavior. Command line options are copied into `$MAKEFLAGS` so that they are implicitly forwarded while doing recursive builds.

PR #456, opened 2022-03-13
ruby: pulseaudio shouldn't block for events

AmatCoder:
Now it sets a timeout of one second


AmatCoder:
With some Linux distros (i.e. Arch) is possible to have installed libpulse but not have running pulseaudio daemon.

In those cases ares freezes at init waiting indefinitely for events.

PR #457, opened 2022-03-13
sfc: fix "Super FX voxel demo"

jbo-85:
sfc: Add database entry for the "Super FX voxel demo" to specify a RAM size of 64k instead of 32k
Support ROM/RAM addressing for Super FX ROMs that are not a power of 2 (i.e. Voxel)

should fix 436


LukeUsher:
Shouldn't all roms be a power of two? non-power of two chips don't exist? 

Wouldn't it would be more correct for mia to pad to the next power-of-two on load time, mirroring how a real rom chip would work, rather than adapting the sfc core to support non power of two?

PR #514, opened 2022-03-24
Allow to disconnect devices from ports + MCD Corpse Killer fix

TascoDLX:
Short version: There's a bug in retail MCD Corpse Killer (U) which breaks input polling when a gamepad is plugged into controller port 2. This fix allows for unplugging of an offending device.

Long version:
1. Adds functionality to disconnect from port when swapping devices. Note, as with MD gamepads, disconnect() may not be implemented in some cases, so deallocation strategy is used (see following commit). Of course, this change will need to be reviewed for devices across the board to ensure safety.
2. A plugged MD gamepad device is deallocated before allocating a new device, since the disconnect action is not handled. In case a valid device name is not given, the device will remain unallocated (null). An alternative approach would be to implement disconnect() to do the same, but this is provided to be the simplest solution. Either way, **similar fixes will need to be done for devices in other cores**, otherwise this remains incomplete.
3. Extracts an identifiable text field -- product version/serial number -- from the MCD disc header. Otherwise, only the file name is made available (as Title/Name).
4. Skip plugging of control pad into port 2 when Corpse Killer CD is loaded. This is handled on first init of the controller ports.

Please take as much time as you need on this PR, and feel free to suggest any alternative approaches if deemed necessary. I can make no guarantees that these changes are safe for all devices outside of the MD core.

### DIFF
--- a/ares/md/controller/port.cpp
+++ b/ares/md/controller/port.cpp
@@ -11,6 +11,8 @@ auto ControllerPort::load(Node::Object parent) -> void {
   port->setType("Controller");
   port->setHotSwappable(true);
   port->setAllocate([&](auto name) { return allocate(name); });
+  port->setDisconnect([&] { return disconnect(); });
+
   port->setSupported({"Control Pad", "Fighting Pad"});
 }
 
@@ -20,7 +22,6 @@ auto ControllerPort::unload() -> void {
 }
 
 auto ControllerPort::allocate(string name) -> Node::Peripheral {
-  device = nullptr;
   if(name == "Control Pad" ) device = new ControlPad(port);
   if(name == "Fighting Pad") device = new FightingPad(port);
   if(device) return device->node;
--- a/ares/md/controller/port.cpp
+++ b/ares/md/controller/port.cpp
@@ -20,6 +20,7 @@ auto ControllerPort::unload() -> void {
 }
 
 auto ControllerPort::allocate(string name) -> Node::Peripheral {
+  device = nullptr;
   if(name == "Control Pad" ) device = new ControlPad(port);
   if(name == "Fighting Pad") device = new FightingPad(port);
   if(device) return device->node;
--- a/desktop-ui/presentation/presentation.cpp
+++ b/desktop-ui/presentation/presentation.cpp
@@ -417,7 +417,13 @@ auto Presentation::loadEmulator() -> void {
 
     Group peripheralGroup;
     { MenuRadioItem peripheralItem{&portMenu};
+      peripheralItem.setAttribute<ares::Node::Port>("port", port);
       peripheralItem.setText("Nothing");
+      peripheralItem.onActivate([=] {
+        auto port = peripheralItem.attribute<ares::Node::Port>("port");
+        port->disconnect();
+        port->allocate(); // deallocation for ports not implementing disconnect()
+      });
       peripheralGroup.append(peripheralItem);
     }
     for(auto peripheral : port->supported()) {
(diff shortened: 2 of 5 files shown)


desktop-ui/presentation/presentation.cpp
```diff
@@ -417,7 +417,13 @@ auto Presentation::loadEmulator() -> void {
 
     Group peripheralGroup;
     { MenuRadioItem peripheralItem{&portMenu};
+      peripheralItem.setAttribute<ares::Node::Port>("port", port);
       peripheralItem.setText("Nothing");
+      peripheralItem.onActivate([=] {
+        auto port = peripheralItem.attribute<ares::Node::Port>("port");
+        port->disconnect();
+        port->allocate(); // deallocation for ports not implementing disconnect()
```
LukeUsher: Is this required? The default disconnect implementation removes the connected peripheral node see https://github.com/ares-emulator/ares/blob/3ca1f9ebb4ae3f472f7fba661746058f84126536/ares/ares/node/port.hpp#L32

ares/md/controller/port.cpp
```diff
@@ -20,6 +20,7 @@ auto ControllerPort::unload() -> void {
 }
 
 auto ControllerPort::allocate(string name) -> Node::Peripheral {
+  device = nullptr;
```
LukeUsher: This shouldn't be required; the call to disconnect() in the parent Node class should clean this up.

TascoDLX: At this time, it is required b/c disconnect() isn't hooked in, and the code is not resetting device except on port unload(). This looks to be the case for many other controller ports as they were written the same way. In this case, allocate() is working as a makeshift connect() since configuration is done automatically by the device.

TascoDLX:
The changes I have queued up for this PR:
- Remove empty allocate() call from Nothing item -- rely on disconnect() only.
- Implement disconnect() in md/controller/port to handle deallocation via device.reset().
- Hook in disconnect() call to md/controller/port.

I think that should cover it for the MD, but let me know if I'm missing anything.

LukeUsher:
> The changes I have queued up for this PR:
> 
> * Remove empty allocate() call from Nothing item -- rely on disconnect() only.
> * Implement disconnect() in md/controller/port to handle deallocation via device.reset().
> * Hook in disconnect() call to md/controller/port.
> 
> I think that should cover it for the MD, but let me know if I'm missing anything.

I’m still not sold on these changes being required, but I’ll do some testing later. 


Higan (and my PolyBlast fork) have supported hot plugging devices, even setting to “nothing” during runtime and it has always worked correctly, with these same emulator cores; unless there’s an underlying issue that has never been picked up on in the entire history of higan? It’s not impossible but unlikely imo

LukeUsher:
I see what's happening; the device does get disconnected from the port and stops responding to input, but the underlying ControllerPort::device pointer doesn't get freed until a new device is connected.

It's not a memory leak because it is freed automatically when a new device is connected, even if a new device is never allocated, it will be freed on unload, so it doesn't leak.

This is why it has never been an issue: calling node->disconnect() is enough to break the connection between ares and the port; it doesn't need to be deallocated.

TascoDLX:
There's an outstanding issue when disconnecting a Fighting Pad that may cause ares to crash. (*tested in Windows only)

Fighting Pad has a thread used for timing, and it seems deleting it can trigger the crash. Note, this seems to be a preexisting issue that occurs in the latest release (v127), in which the deletion is triggered in allocate() rather than disconnect().

Currently seeking resolution to this issue in the Discord server.

LukeUsher:
Sorry it took so long; I've addressed the crash issues in master

PR #519, opened 2022-04-09
n64: fix SIGILL at startup when SSE4.2 is not supported

### LINKED ISSUE #193: [crash] Some core makes ares crash on Linux as soon as it runs
I initially built ares with just `make` and when I tried to run it it'd crash with `illegal instruction (core dumped)` as soon as it attempted to load. To make it clear, I mean the UI itself crashed, before even trying to load any game. I had to build ares using `make clean` then `make cores=md` (I only needed that core anyway) and now it runs properly.

In other words, support for some core (not Mega Drive) is causing it to crash on Linux. No idea which one.

PC specs:
* Ubuntu 18.04
* x86-64
* 4GB RAM

rasky:
The N64 core was forcing -msse4.2 for compilation of the whole Ares.
On host PCs without SSE 4.2, Ares would crash at startup with SIGILL
because GCC used SSE 4.2 instructions even in hiro, while building up
the UI.

RSP SIMD emulation requires SSE 4.2, but the solution here is to
fallback to SISD if SSE 4.2 is not supported. This can be checked
with a predefined macro, that will be set by GCC. We do compile with
-march=native, so this will reflect the host PC of the build process.

Since this is yet another compile-time flag that affects the N64 core
performance, add a status message warning if SIMD is disabled.

Fixes #193


nall/intrinsics.hpp
```diff
@@ -173,16 +173,21 @@ namespace nall {
...
   };
 #elif defined(__amd64__) || defined(_M_AMD64)
   #define ARCHITECTURE_AMD64
+  #ifndef __SSE4_1__
+  #define __SSE4_1__ 0
+  #endif
   struct Architecture {
     static constexpr bool x86   = 0;
     static constexpr bool amd64 = 1;
+    static constexpr bool sse41 = __SSE4_1__;
```
Screwtapello: This is `sse41` while all the other `Architecture` structs have `sse42`?

rasky: Right I need to fix it

zoncabe: i already tried compiling whit the ss41 and the ss42 lines but its crashing both ways

rasky: Is it crashing or not compiling?

There is a problem in the patch as shown by the CI. I'll fix it. Afterwards, when it compiles for you, it should work

rasky: The problem should be fixed now. Please try again.

zoncabe (approved):
this was made based on a debugging from a computer of my ownership and i can confirm that ares was compiled succsesfully and is running now

PR #526, opened 2022-04-16
Nall: Fix typo in nall/decode/chd.hpp:    type --> typeStr

shinra-electric:
Using `type` results in compiler warnings: `warning: result of comparison against a string literal is unspecified (use an explicit string comparison function instead)` 

Was informed by merryhime that `typeStr` was probably intended.

Changing to `typeStr` fixes the warnings. 

Should probably be tested to ensure the comparison still works as intended.

This is following up from #525


LukeUsher:
nice spot, thanks

PR #528, opened 2022-04-23
build: fix building from Windows command prompt

invertego:


LukeUsher:
This currently fails to build on macOS because vulkan (and therefore parallel-rdp) are not available there

PR #529, opened 2022-04-23
n64: implement proper PI bus writes with IO busy bit

rasky:
All writes through the PI bus are actually performed asynchronous by
the hardware, and the status is reflected in the PI IO busy bit.

This commit emulates this properly, and also emulates what happens while
the busy bit is set: further writes are ignored, and reads at any
address return the word being written.

This fixes A Bug's Life and Toy Story 2.

### DIFF
--- a/ares/n64/pi/bus.hpp
+++ b/ares/n64/pi/bus.hpp
@@ -1,0 +1,76 @@
+inline auto PI::readWord(u32 address) -> u32 {
+  if(address <= 0x046f'ffff) return ioRead(address);
+
+  if (unlikely(io.ioBusy)) {
+    writeForceFinish(); //technically, we should wait until Queue::PI_BUS_Write
+    return io.busLatch;
+  }
+  return busRead<Word>(address);
+}
+
+template <u32 Size>
+inline auto PI::busRead(u32 address) -> u32 {
+  static_assert(Size == Half || Size == Word);  //PI bus will do 32-bit (CPU) or 16-bit (DMA) only
+  static constexpr u32 unmapped = 0;
+
+  if(address <= 0x04ff'ffff) return unmapped; //Address range not memory mapped, only accessible via DMA
+  if(address <= 0x0500'03ff) return dd.c2s.read<Size>(address);
+  if(address <= 0x0500'04ff) return dd.ds.read<Size>(address);
+  if(address <= 0x0500'057f) return dd.read<Size>(address);
+  if(address <= 0x0500'05bf) return dd.ms.read<Size>(address);
+  if(address <= 0x05ff'ffff) return unmapped;
+  if(address <= 0x063f'ffff) return dd.iplrom.read<Size>(address);
+  if(address <= 0x07ff'ffff) return unmapped;
+  if(address <= 0x0fff'ffff) {
+    if(cartridge.ram  ) return cartridge.ram.read<Size>(address);
+    if(cartridge.flash) return cartridge.flash.read<Size>(address);
+    return unmapped;
+  }
+  if(address <= 0x13fe'ffff) return cartridge.rom.read<Size>(address);
+  if(address <= 0x13ff'ffff) return cartridge.isviewer.read<Size>(address);
+  if(address <= 0x7fff'ffff) return unmapped;
+  return unmapped; //accesses here actually lock out the RCP
+}
+
+inline auto PI::writeWord(u32 address, u32 data) -> void {
+  if(address <= 0x046f'ffff) return ioWrite(address, data);
+
+  if(io.ioBusy) return;
+  io.ioBusy = 1;
+  io.busLatch = data;
+  queue.insert(Queue::PI_BUS_Write, 300);
+  return busWrite<Word>(address, data);
+}
+
+template <u32 Size>
+inline auto PI::busWrite(u32 address, u32 data) -> void {
+  static_assert(Size == Half || Size == Word);  //PI bus will do 32-bit (CPU) or 16-bit (DMA) only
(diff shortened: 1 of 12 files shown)


ares/n64/pi/pi.hpp
```diff
@@ -45,6 +52,7 @@ struct PI : Memory::IO<PI> {
     n32 readLength;
     n32 writeLength;
     n1  romLockout;
+    n32 busLatch;
```
invertego, line 58: Needs serialization

ares/n64/pi/bus.hpp
```diff
@@ -0,0 +1,79 @@
+template<u32 Size>
+inline auto PI::busRead(u32 address) -> u64 {
+  static constexpr u64 unmapped = 0;
+
+  if (unlikely(io.ioBusy)) {
+    io.ioBusy = 0; //technically, we should wait until Queue::PI_BUS_Write
```
invertego: Can you elaborate on why this early reset is needed?

ares/n64/pi/bus.hpp
```diff
@@ -0,0 +1,79 @@
...
+  }
+  if(address >= 0x13ff'0000 && address <= 0x13ff'ffff) {
+    return cartridge.isviewer.read<Size>(address);
+  }
+  return cartridge.rom.read<Size>(address);
+}
+
+template<u32 Size>
+inline auto PI::busWrite(u32 address, u64 data) -> void {
+  if(io.ioBusy) return;
```
invertego, line 38: How will this interact with PI::dmaRead()? It makes back-to-back writes in a loop.

ares/n64/pi/bus.hpp
```diff
@@ -0,0 +1,79 @@
...
+    io.busLatch = data << ((~address & 1) * 16);
+  }
+  if constexpr(Size == Word) {
+    io.busLatch = data;
+  }
+  if constexpr(Size == Dual) {
+    io.busLatch = data >> 32;
+  }
+
+  queue.insert(Queue::PI_BUS_Write, 300);
```
invertego, line 41: Are we at risk of overflowing the queue? I'm concerned about DMA in particular, as it goes through this code path.

ares/n64/pi/bus.hpp
rasky: As tried to describe in the comment, all reads are blocking. In case of a read while a write is ongoing, the VR4300 is stalled until the write is finished, and then the last written value is returned.

To be more accurate, we should in general emulate stalls on reads, and in a case a write is ongoing, stall the remaining amount of cycles until that write is finished.

This is a quick workaround. If you want I can call queue.remove here so that we’re more safe wrt queue overflows

ares/n64/pi/bus.hpp
rasky: Right, that is broken now. Suggestions?

We could have an internal function that does performs a word write (eg busWriteWord) without any PI busy concern, and call it from both dmaWrite and busWrite.

ares/n64/pi/bus.hpp
rasky: In general, only one write can be ongoing at the same time. There are two shortcuts right now (reading after writing, and isviewer) so I guess I could refactor those into a write abort functions that also removes the queue entry, so we avoid the problem altogether

ares/n64/pi/pi.hpp
rasky: Done

ares/n64/pi/bus.hpp
rasky: I went ahead and did this. Now there is a lower-level layer of functions called busWrite/busRead that only can do 16-bit and 32-bit accesses (mimicking what happens on the PI bus: DMA can do both 16-bit and 32-bit accesses while CPU will always do 32-bit accesses).

These are now called from readWord / writeWord so that the special word mangling performed by Memory::IO get applied also to bus reads/writes and the code doesn't need to be duplicated like in the previous version of this PR.

Given this new approach, the quirky behavior of cartridge 8-bit/16-bit reads is now more straightforward and simply happens because ROM is a physical 16-bit device so the 32-bit access from PI is correctly aligned to 16-bit. This is emulated via a new class called Memory::Readable16.

ares/n64/pi/bus.hpp
```diff
@@ -0,0 +1,76 @@
...
+  if(address <= 0x7fff'ffff) return;
+}
+
+inline auto PI::writeFinished() -> void {
+  io.ioBusy = 0;
+}
+
+inline auto PI::writeForceFinish() -> void {
+  io.ioBusy = 0;
+  queue.remove(Queue::PI_BUS_Write);
```
invertego, line 75: I noticed that priority_queue::remove(...) loops over the entire (currently 512 element) array, which is not ideal. That said, it looks like it could be safely modified to only loop over priority_queue::size elements.

rasky: Done

ares/n64/pi/dma.cpp
```diff
@@ -40,8 +39,9 @@ auto PI::dmaWrite() -> void {
       cur_len = max(cur_len-misalign, 0);
     }
 
+    cpu.recompiler.invalidateRange(io.dramAddress, io.dramAddress + cur_len);
```
invertego: `if constexpr(Accuracy::CPU::Recompiler) {`

As an aside, the existing invalidate() calls in Bus::write() should be guarded by this check, too. All other references to the recompilers in the n64 core are guarded this way.

rasky: Done

PR #535, opened 2022-04-26
n64: add separate CIC 7102 detection

bigbass1997:
The IPL3 bootcode associated with CIC-NUS-7102 (used in Lylat Wars), is not the same as the IPL3 used with CIC-NUS-6101 (used in Star Fox 64). Ares doesn't make this distinction.

Without knowing more about how each IPL3 differs in a functional sense, it's tough to say how exactly they are different; but at the least, the CRC calculated over this section of the rom, is different. Thus, Ares incorrectly detects Lylat Wars as using the 7101 (the PAL version of the fallback case).

While the game still appears to work on Ares despite this incorrect detection, I believe that may be due to some other inaccuracy in Ares. (Using the 64Drive) When Lylat Wars is played on an NTSC console, using CIC 7101, the game just shows a black screen. However, when using CIC 7102, as the official cartridge would, the game works correctly. Note, I don't have a PAL console to test this with.

### DIFF
--- a/mia/medium/nintendo-64.cpp
+++ b/mia/medium/nintendo-64.cpp
@@ -116,12 +116,15 @@ auto Nintendo64::analyze(vector<u8>& data) -> string {
   u8 revision = data[0x3f];
 
   //detect the CIC used for a given gamepak based on checksumming its bootcode
-  //note: NTSC 6101 = PAL 7102 and NTSC 6102 = PAL 7101 (IDs are swapped)
+  //note: NTSC 6101 != PAL 7102; they use different bootcodes
+  //note: NTSC 6102 == PAL 7101
   //note: NTSC 6104 / PAL 7104 was never officially used
+  //note: Except for above cases, NTSC 610x == PAL 710x
   bool ntsc = region == "NTSC";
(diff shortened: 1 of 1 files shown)


bigbass1997:
Also, I didn't think to include it, but technically, the CRC calculations are not using the entire IPL3 code segment. Ares calculates it over `[0x40, 0x09C0)` of the rom, whereas the IPL3 actually spans `[0x40,0x1000)`. Ares covers enough of the IPL3, that the differences are still apparent in the CRC values, so I don't know if it's worth changing.

rasky:
> While the game still appears to work on Ares despite this incorrect detection, I believe that may be due to some other inaccuracy in Ares

I think there are a couple of reasons:

* Ares detects the region using the header information. It then uses the region to select the correct PIF ROM (IPL1/2) for the region, which then proceeds to put the correct TV type in 0x80000300, which then is used by Lylat Wars. Notice that this by itself is not necessarily enough to break the game: Lylat Wars could be solid and works on both PAL and NTSC displays (though it's been officially released only in one region). This could be tested for curiosity.
* (The real reason). The region protection check is made before VR3000 even boots, by the PIF core communicating with CIC. Ares does not emulate the PIF and CIC SM5 cores yet, so the protocol where the protection check happens is basically non existent.

PR #542, opened 2022-05-01
n64: make sure TLB entries are correctly initialized at power

rasky:
This bug has always been there, but was hidden by the incorrect TLB
matching logic. After fixing the logic in 0bf24b3, this bug was
uncovered.

Fixes Goldeneye.


Shideravan:
Nice find!
It's awesome that we now have proper TLB logic emulated and the "regression" (or better saying, the uncovered issue) is no more

PR #545, opened 2022-05-04
n64: fix cache to use physical addresses

rasky:
There is a bit confusion of virtual vs physical addresses in Ares,
so dcache/icache were sometimes using a virtual address, sometimes
a physical one. This is wrong: cache is based on physical addresses
only.

Moreover, the CACHE opcode was not even going through TLBs, while it
should. So a CACHE opcode on a TLB-mapped address wasn't working before.


rasky:
This PR makes Indiana Jones and the Infernal Machine and Star Wars Episode 1: Battle for Naboo playable.

Shideravan:
Awesome!!!!
Thanks for the update @rasky

PR #565, opened 2022-05-24
n64: rework RSP DMA double buffering

rasky:
This change implements RSP DMA double buffering in a more faithful
way, so that reads from the registers always return the current values
(ongoing transfer or finished transfer) rather than the pending ones.

Moreover, this also makes Ares transfer bytes in background a row at
a time, rather than all of them at the end. It is a bit more accurate,
and it was also easier to implement in the new architecture.

This also fixes several tests in n64-systemtest.

### DIFF
--- a/ares/n64/cpu/cpu.cpp
+++ b/ares/n64/cpu/cpu.cpp
@@ -52,7 +52,7 @@ auto CPU::synchronize() -> void {
 
   queue.step(clocks, [](u32 event) {
     switch(event) {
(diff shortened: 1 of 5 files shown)


LukeUsher:
This PR seems to significantly change the serialization function; perhaps serializer version should be updated in ares.hpp?

LukeUsher (approved):

LukeUsher:
re: serialization: we will move to per-core versonining in the next release

PR #566, opened 2022-05-24
n64: reactivate 64-bit memory map

rasky:
Ares used to have full 64-bit TLB support but it was removed at some
point because of some wrong testing that apparently made us believe
that the CPU didn't support that mode. It will have to be restored from
the git history in the future, though it's not very urgent because
nobody is using 64-bit pointers on N64 (there is no toolchain with
64-bit pointers, nor it seems useful).

Luckily, at least the segment parsing was left in and just commented out.
This allows to pass the last two remaining tests in n64-systemtest.

### DIFF
--- a/ares/n64/cpu/memory.cpp
+++ b/ares/n64/cpu/memory.cpp
@@ -75,8 +75,8 @@ auto CPU::userSegment64(u64 address) const -> Context::Segment {
 
 auto CPU::segment(u64 address) -> Context::Segment {
   auto segment = context.segment[address >> 29 & 7];
-//if(likely(context.bits == 32))
-  return (Context::Segment)segment;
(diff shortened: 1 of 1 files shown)


Shideravan:
@rasky maybe this is the commit where the changes was made, don't is?
https://github.com/ares-emulator/ares/commit/54159cf95e63f6864323718571cc4d389a5044b4

rasky:
> @rasky maybe this is the commit where the changes was made, don't is?
> [54159cf](https://github.com/ares-emulator/ares/commit/54159cf95e63f6864323718571cc4d389a5044b4)

Yes it’s that one, thanks for finding it. As I wrote, I’m not in a rush right now to reverse it, but I’ll keep it handy.

PR #586, opened 2022-05-28
debugger: fix address masking in instruction tracer (for 64 bit addresses)

CasualPokePlayer:
This is similar to https://github.com/ares-emulator/ares/commit/822186413c9274295e49b922297f0cc744921eda, but fixes masking for N64 as it has 64 bit addresses.

### DIFF
--- a/ares/ares/node/debugger/tracer/instruction.hpp
+++ b/ares/ares/node/debugger/tracer/instruction.hpp
@@ -28,7 +28,7 @@ struct Instruction : Tracer {
   }
 
   auto address(u64 address) -> bool {
-    address &= (1ull << _addressBits) - 1;  //mask upper bits of address
+    address &= (u128(1) << _addressBits) - 1;  //mask upper bits of address
     _address = address;
     address >>= _addressMask;  //clip unneeded alignment bits (to reduce _masks size)
 
@@ -57,7 +57,7 @@ struct Instruction : Tracer {
   //call when writing to executable RAM to support self-modifying code.
   auto invalidate(u64 address) -> void {
     if(unlikely(_mask && updateMasks())) {
-      address &= (1ull << _addressBits) - 1;
+      address &= (u128(1) << _addressBits) - 1;
       address >>= _addressMask;
       _masks[address >> 3] &= ~(1 << (address & 7));
     }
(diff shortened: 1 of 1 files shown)


rasky:
We can’t use u128 if we want to compile Ares on 32-bit platforms. I suggest we either prepare the mask once by shifting right maxuint and then use it, or special case with an if the 64-bit case where no masking is required

ares/ares/node/debugger/tracer/instruction.hpp
```diff
@@ -28,7 +28,7 @@ struct Instruction : Tracer {
   }
 
   auto address(u64 address) -> bool {
-    address &= (1ull << _addressBits) - 1;  //mask upper bits of address
+    address &= (u128(1) << _addressBits) - 1;  //mask upper bits of address
```
invertego: Let's avoid 128-bit math if it's not strictly needed.
```suggestion
    address &= ~0ull >> (64 - _addressBits);  //mask upper bits of address
```

ares/ares/node/debugger/tracer/instruction.hpp
```diff
@@ -57,7 +57,7 @@ struct Instruction : Tracer {
   //call when writing to executable RAM to support self-modifying code.
   auto invalidate(u64 address) -> void {
     if(unlikely(_mask && updateMasks())) {
-      address &= (1ull << _addressBits) - 1;
+      address &= (u128(1) << _addressBits) - 1;
       address >>= _addressMask;
       _masks[address >> 3] &= ~(1 << (address & 7));
     }
```
invertego, line 63: GitHub apparently doesn't allow comments on lines not included in a change, but line 107 appears broken as well. It should be replaced with something like this:
```
auto size = 1ull << (_addressBits - _addressMask - 3);
```

invertego (changes requested):
Thank you for fixing this again! I have just a couple requests.

CasualPokePlayer:
@invertego I already pushed in a commit to avoid the 128 bit use :P (and did catch that other broken case)

Although your masking code seems to be more clear here, I'll go replace it. Your size code would work fine if we can assume that the resulting shift is not less than 0 (or really an absurdly large number).

invertego:
Yes, it seems I must have started my review just before you pushed your commit. I don't believe your commit actually addresses the _addressBits == 64 case (64 % 64 == 0, so (1ull << 0) - 1 == 0), but if you're going to change it again that shouldn't matter.

CasualPokePlayer:
> I don't believe your commit actually addresses the _addressBits == 64 case (64 % 64 == 0, so (1ull << 0) - 1 == 0)

Yes, I realized that as I was adding in your code kek

invertego (approved):

PR #641, opened 2022-07-21
Fix fds loading

remutro:
Auto detection depends on emulator name matching mia media name. From mia's perspective, it is returning a famicom disk, so it makes sense to leave this as is (and would have required a lot more changes). Although "Famicom Disk System" sounds a little better for an emulator, "Famicom Disk" should be good enough for now. Otherwise I guess we would need to maintain some type of map between media type and emulator name. I looked at other emulator names and media and it looks like all match up, so this was the path of least resistance for getting *.fds games to auto load.


LukeUsher:
Something weird is going on with your branch here; the number of commits seems excessive?

PR #642, opened 2022-07-21
match emulator name with expected mia media definition to fix auto-detection of fds

remutro:
Auto detection depends on emulator name matching mia media name. From mia's perspective, it is returning a famicom disk, so it makes sense to leave this as is (and would have required a lot more changes). Although "Famicom Disk System" sounds a little better for an emulator, "Famicom Disk" should be good enough for now. Otherwise I guess we would need to maintain some type of map between media type and emulator name. I looked at other emulator names and media and it looks like all match up, so this was the path of least resistance for getting *.fds games to auto load.

### DIFF
--- a/ares/fc/fds/fds.cpp
+++ b/ares/fc/fds/fds.cpp
@@ -10,7 +10,7 @@ FDS fds;
 
 auto FDS::load(Node::Object parent) -> void {
   port = parent->append<Node::Port>("Disk Slot");
-  port->setFamily("Famicom Disk");
+  port->setFamily("Famicom Disk System");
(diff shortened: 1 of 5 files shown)


FitzRoyX:
The other entries in mia.cpp don't say things like "Super Famicom Cartridge". So I think mia.cpp needs to be changed to "Famicom Disk System".

LukeUsher:
I'll agree with @FitzRoyX here;

remutro:
Removed all references to just "Famicom Disk" and now everything reflects as (some form of) "Famicom Disk System".

PR #655, opened 2022-07-31
alphabetically order systems in makefile

FitzRoyX:


LukeUsher:
I'm not sure about this; basis that the current order is geared around keeping related consoles by the same vendor/in the same family together + in order of generation

FitzRoyX:
The only purpose of this list seems to be to skip unready cores like saturn, so this is a nitpick PR tbh. But if anything were to be moved down, I'd think "system starts with" would be the fastest way to find something. Like why would I expect to find coleco near the end? I guess I don't get all the hidden rules in place.

LukeUsher:
Closing as we use cmake now; making this no longer relevant

PR #656, opened 2022-08-01
md: add sram entry for psy-o-blade

FitzRoyX:
This game's internal bit for sram amount is malformed or something. Ares is unable to handle the value it sees and a massive runaway ram file starts being written to the user's hard disk. Although this PR stops that from happening in this case, it won't stop other cases should a user try to load a rom with the same malformed data. As a result, I recommend not merging this PR until the root problem is discovered and safeguarded against so that it is not forgotten. Could be as simple as a missing clamp?

### DIFF
--- a/mia/medium/mega-drive.cpp
+++ b/mia/medium/mega-drive.cpp
@@ -309,16 +309,10 @@ auto MegaDrive::analyzeStorage(vector<u8>& rom, string hash) -> void {
     ram.size = 32768;
   }
 
-  //Might and Magic III: Isles of Terra (USA)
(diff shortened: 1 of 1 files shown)


LukeUsher:
The issue with psy-o-blade is that the header for sram size specifies the memory range as  0x20000020-0x00010020

This start address is invalid, and the end address is significantly lower than start; which is obviously incorrect. 
We will need a database entry to enable working sram, but we could potentially add a safeguard that disables sram if the memory address range is obviously wrong, to prevent runaway files?

Issue #521, opened 2022-04-11
Add basic build instructions?


mariomadproductions:
Even if its just "cd into desktop-ui and run make", it'd probably be a good idea to put that somewhere.

rasky:
Given that it's just "make", do we really need to write that?

Shideravan:
> Given that it's just "make", do we really need to write that?

I really think so

Kawa-oneechan:
Is it really just "make" though? I had to jump through some pretty specific hoops to get a build environment set up from which I could just "make".

rasky:
Which hoops?

bigbass1997:
It's not nearly as simple as _just_ running make. Although it's easier than it used to be, it's not necessarily that simple. Either way, is there a negative to putting some instructions in the readme?

For people who have never needed to build something manually on their system before, and potentially even if they have, they may not have all the necessary packages. It's not that uncommon to find readme's that list what packages are necessary. However, to test this, I spun up a fresh VM (Ubuntu 22.04 LTS Beta Server with KDE Plasma).

Ran into a series of failed build, each more confusing than the last. With some googling and searching through `apt` and `apt-file`, I managed to find all the right packages. `make g++ pkg-config libgtk2.0-dev libasound2-dev`. Also needed something that provided OpenGL files, presumably mesa or some equivalent would work, for this I used `libgl-dev` (vendor neutral package). After that, I managed to successfully build and run ares.

Additionally, it'd be nice if the different configurations were mentioned in the readme too. Like how to change between accuracy and performance modes, and how to specify which cores to build. Also it'd be helpful to say how to start ares after building, because that's not obvious to people who haven't run it before.

Kawa-oneechan:
> Which hoops?

Compiling in Windows, I needed a _very specific_ compiler distribution. You'll find that [higan.readthedocs.io](https://higan.readthedocs.io/en/stable/install/windows/#compiling-from-source-on-windows) recommends one right there, but when I tried that it didn't work as required things were straight-up missing.

(I know this is *higan* readthedocs but I was in fact trying to compile bsnes, bear with me here.)

[...]

PR #671, opened 2022-08-14
Issue 521: Basic Build Instructions

remutro:
Issue #521 - Add basic build instructions

Tested on a fresh Ubuntu 22.04 installation, as well as a fresh Windows 11 installation (no Macs to test with). Hopefully this is a good compromise between just enough and being too verbose (assumes some compiling knowledge). Also hoping this keeps these same questions showing up in the forums.


rasky:
The build instructions mention to use `make -C desktop-ui` but we have a top-level Makefile. Can we simplify them to remove the directory mention?

PR #679, opened 2022-08-21
md/ym2612: do not apply SSG 4x speedup to attack phase

invertego:
Fixes the buzzing noise at the start of Yogi Bear


invertego:
Never mind, this needs more investigation.

PR #692, opened 2022-08-29
ms: Add support for the Sports Pad

raphnet:
This adds support for the Sports Pad, a trackball controller for Master System. Input
is done using a Mouse or a trackball (best).

Tested with Sports Pad Soccer and a homebrew test ROM.

### DIFF
--- a/ares/ms/controller/controller.cpp
+++ b/ares/ms/controller/controller.cpp
@@ -6,5 +6,6 @@ namespace ares::MasterSystem {
 #include "gamepad/gamepad.cpp"
 #include "light-phaser/light-phaser.cpp"
 #include "paddle/paddle.cpp"
+#include "sports-pad/sports-pad.cpp"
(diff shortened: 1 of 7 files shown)


raphnet:
It should be noted that under Linux at least (the only platform I can test on) the cursor does not move predictably without this also: #693

FitzRoyX:
I forgot to ask you this on your last PR, but is the taito paddle controller for MSX similar to the sega one you already implemented for master system?

LukeUsher (approved):

raphnet:
I just had a look, and it is quite different. The Sega Paddle sends 4 bit nibbles and is self clocked. The taito paddle controller on the other hand is clocked from outputs controlled by the MSX, and data is transmitted one bit at a time serially...

But this sounds fun, I will try implementing this peripheral for for MSX.

PR #693, opened 2022-08-29
ruby: xlib mouse double event workaround

raphnet:
When the mouse is grabbed, the current implementation hides the
pointer and centers it on the screen.

When the mouse is polled, its relative motion is determined by how far it
has moved from the center of the screen, then it is moved back to the
center by calling XWarpPointer().

XWarpPointer generates an event resulting in an additional poll, which then
sees no movement at all (since the pointer is now centered) and therefore
resets the relative motion values to zero.

If the above reset occurs before the controller code has had the chance to
access the relative motion values, the motion never reach the game. This
works around this issue using a flag to ignore the next event if the
mouse is still centered on the screen.

### DIFF
--- a/ruby/input/mouse/xlib.cpp
+++ b/ruby/input/mouse/xlib.cpp
@@ -21,6 +21,7 @@ struct InputMouseXlib {
     s32 threshold = 0;
     u32 relativeX = 0;
     u32 relativeY = 0;
+    bool ignoreNextCenter = false;
   } ms;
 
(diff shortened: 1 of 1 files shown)


raphnet:
For the record, without this I could not get my Master System Sports Pad code (#692) working correctly. The pointer would only move correctly for a few seconds or unpredictably.

For what it's worth, I tested and as far as I can tell, this change does not break SFC mouse emulation.

Issue #204, opened 2021-09-25
[PCE] Valkyrie no Desentsu: Major graphical issues


Shideravan:
On start of gameplay section, the camera doesn't follow the playable character and the hud follows him.

invertego:
This should be fixed now, even when "Pixel Accuracy Mode" is disabled.

PR #699, opened 2022-09-01
pce: performance vdp timing overrides

invertego:
Following the precedent set by the SFC performance PPU, override the
timing of the PCE performance VDP for a handful of problematic games.

All of these games already work fine with the accuracy VDP (enabled with
the "pixel accuracy mode" video setting) which does not and never will
contain contain game specific logic.

Affected games:
- Kore ga Pro Yakyuu '89
- Kore ga Pro Yakyuu '90
- TV Sports Baseball
- TV Sports Football
- TV Sports Hockey
- Valkyrie no Densetsu (#204)
- Victory Run

### DIFF
--- a/ares/pce/vdp-performance/vdp.cpp
+++ b/ares/pce/vdp-performance/vdp.cpp
@@ -53,7 +53,7 @@ auto VDP::main() -> void {
     vdc1.vsync();
   }
 
-  step(512);
+  step(renderingCycle);
 
   vdc0.hclock(); if(Model::SuperGrafx())
   vdc1.hclock();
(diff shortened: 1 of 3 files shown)


rasky:
Do we want to add game hacks in performance cores? I thought performance cores were a way to investigate a more performant implementation that might have some limits; if there are limits, I think they should stay.

LukeUsher:
> Do we want to add game hacks in performance cores? I thought performance cores were a way to investigate a more performant implementation that might have some limits; if there are limits, I think they should stay.

If we follow Near’s other projects, there are game specific timing values in bsnes for the few problematic games, but also it forces the pixel ppu for others. 

We could either do the same, or we could force the pixel accurate ppu for these games.

invertego:
I patterned this change directly after the performance ppu as it exists today in ares:
https://github.com/ares-emulator/ares/blob/f9eb9304f0ae46c939bbea8d1273711a8d289392/ares/sfc/ppu-performance/ppu.cpp#L154-L159

FitzRoyX:
> If we follow Near’s other projects, there are game specific timing values in bsnes for the few problematic games, but also it forces the pixel ppu for others.

I've always thought that overriding to a particular renderer is a bad idea. Because if someone is using the scanline renderer, it's because they can't get 60fps in the pixel one. Dropping below 60 is worse than i.e. not having a plane shadow drawn.

PR #703, opened 2022-09-03
Correct types for screen.hpp

CasualPokePlayer:
Fixes some memory leaks due to delete being used instead of delete[]


kev4cards:
I tried a local build of this yesterday. I saw about 100 MB less memory being used upon initial load of an N64 game as opposed to yesterday's latest master.

CasualPokePlayer:
I'm realizing I accidentally PR'd this to the n64dd branch... oops.

PR #704, opened 2022-09-03
Correct types for screen.hpp

CasualPokePlayer:
Fixes some memory leaks due to delete being used instead of delete[]


CasualPokePlayer:
https://github.com/ares-emulator/ares/pull/703 was intended to do this but I accidentially had the n64dd branch as the target, probably want to reset that branch before the PR.

LukeUsher:
Fwiw the 64dd branch was temporary; it's being worked on over at https://github.com/luigiblood/ares and will be PR'd into here when complete

PR #706, opened 2022-09-04
ms: Fix writes to controller port 2

raphnet:
Corrects an error in the original pull request which sent incorrect data to port 2. The Sports Pad now also works in port 2.


raphnet:
Thanks to @jbo-85 for pointing this out!

PR #711, opened 2022-09-05
desktop-ui: Document command-line options and add --help

raphnet:
Ideally I think one should not need to read the source code to learn that ares accepts command-line options and what those options are.

I wish I had known earlier that I could pass --system MSX when loading a ROM directly, it would have saved me a lot of time. I had tried --help, but it just opened the application window...

### DIFF
--- a/desktop-ui/desktop-ui.cpp
+++ b/desktop-ui/desktop-ui.cpp
@@ -67,18 +67,18 @@ auto nall::main(Arguments arguments) -> void {
   Emulator::construct();
 
   if(arguments.take("--help")) {
-	print("Usage: ares [OPTIONS]... game\n\n");
-	print("Options:\n");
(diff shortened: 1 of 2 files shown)


desktop-ui/desktop-ui.cpp
```diff
@@ -65,6 +65,23 @@ auto nall::main(Arguments arguments) -> void {
 
   inputManager.create();
   Emulator::construct();
+
+  if(arguments.take("--help")) {
+	print("Usage: ares [OPTIONS]... game\n\n");
```
LukeUsher: nit: indent too large

LukeUsher (changes requested):
Just a very minor code style thing; everything else looks good

raphnet:
Thank you for reviewing, I just corrected this.

PR #712, opened 2022-09-05
build: Add a clean target to the top level Makefile

raphnet:
Pull request #690 simplified the build instructions, however typing `make clean` in the top level directory as instructed does not work.

This adds a clean target to the top makefile so that this also works.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -3,4 +3,7 @@ all: desktop
 desktop:
 	$(MAKE) -C desktop-ui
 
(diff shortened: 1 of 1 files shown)


invertego:
I didn't see this before putting out #713, but it looks like we had the same idea. My PR adds all the targets supported by desktop-ui, not just clean.

raphnet:
Ok, what you did is much better! This is not needed then.

PR #713, opened 2022-09-05
build: pass through target from top-level makefile

invertego:


LukeUsher:
In the future it may become necessary to specify the target/ui to build, but since desktop-ui is the only option at present, this is fine for now

PR #719, opened 2022-09-06
n64: improve generation of FPU exceptions.

rasky:
Hook at least standard IEEE-754 exceptions on basic operations. This
should already catch most mistakes done while developing home-brew
games.

### DIFF
--- a/ares/n64/cpu/interpreter-fpu.cpp
+++ b/ares/n64/cpu/interpreter-fpu.cpp
@@ -115,10 +115,10 @@ auto CPU::setControlRegisterFPU(n5 index, n32 data) -> void {
 
     if(fpu.csr.roundMode != roundModePrevious) {
       switch(fpu.csr.roundMode) {
-      case 0: fesetround(FE_TONEAREST);  break;
-      case 1: fesetround(FE_TOWARDZERO); break;
-      case 2: fesetround(FE_UPWARD);     break;
-      case 3: fesetround(FE_DOWNWARD);   break;
+      case 0: fenv.setRound(float_env::toNearest);  break;
+      case 1: fenv.setRound(float_env::towardZero); break;
+      case 2: fenv.setRound(float_env::upward);     break;
+      case 3: fenv.setRound(float_env::downward);   break;
       }
     }
   } break;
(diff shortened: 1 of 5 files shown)


invertego:
This is a good candidate for a new flag under Accuracy::CPU. Messing with the host floating point control/status register is absolute death for performance. I observed a drop in the Mario 64 start screen from ~95 FPS -> 50 FPS. I've literally never heard of a game (on any platform) that cares about over/underflow and inexact flags, so it should be fine if we default to not emulating them.

It's already a bit annoying that you have to recompile ares to switch between the interpreter and recompiler, so we should probably take a look at making that dynamic along with the other accuracy flags. That would allow homebrew devs who care about these things to opt-in without having to build ares from source.

rasky:
Another option would be to activate FPU exceptions on the host, install a signal handler which simply sets a global flag, let the code run and check the global flag. This implies that we need to compile with `-ftrapping-math` but maybe that's acceptable, and we could even limit that flag to the n64 core (assuming we activate the FPU exceptions in the n64 core initialization).

rasky:
After several back and forth with @invertego, we realized that the PR is slow because of the use of the standard library routines which do RMW cycles and also modify the x87 state. By adding a custom implementation in nall/float-env.hpp, the performance regression is solved.

ares/n64/cpu/interpreter-fpu.cpp
```diff
@@ -115,16 +115,81 @@ auto CPU::setControlRegisterFPU(n5 index, n32 data) -> void {
...
+      case 2: fenv.setRound(float_env::upward);     break;
+      case 3: fenv.setRound(float_env::downward);   break;
       }
     }
   } break;
   }
 }
 
+auto CPU::fpeDivisionByZero() -> bool {
+    fpu.csr.cause.divisionByZero = 1;
```
invertego: Nit: 2 space indentation

rasky: Done

invertego (approved):
LGTM. Thank you for working to optimize this and especially for your work on the trapping implementation even though we ended up not needing to use it.

ares/n64/cpu/interpreter-fpu.cpp
```diff
@@ -525,32 +578,56 @@ auto CPU::FSQRT_D(u8 fd, u8 fs) -> void {
...
   if(!scc.status.enable.coprocessor1) return exception.coprocessor1();
-  FD(f64) = FS(f64) - FT(f64);
+  FD(f64) = CHECK_FPE(f64, FS(f64) - FT(f64));
 }
 
 auto CPU::FTRUNC_L_S(u8 fd, u8 fs) -> void {
   if(!scc.status.enable.coprocessor1) return exception.coprocessor1();
-  FD(s64) = FS(f32) < 0 ? ceil(FS(f32)) : floor(FS(f32));
+  auto f = FS(f32);
+  if (isinf(f) || isnan(f) || f >= 0x1p+63 || f < -0x1p-63) {
```
invertego:
```suggestion
  if (isinf(f) || isnan(f) || f >= 0x1p+63 || f < -0x1p+63) {
```

ares/n64/cpu/interpreter-fpu.cpp
```diff
@@ -525,32 +578,56 @@ auto CPU::FSQRT_D(u8 fd, u8 fs) -> void {
...
+  } else {
+    FD(s64) = f < 0 ? ceil(f) : floor(f);
+  }
 }
 
 auto CPU::FTRUNC_L_D(u8 fd, u8 fs) -> void {
   if(!scc.status.enable.coprocessor1) return exception.coprocessor1();
-  FD(s64) = FS(f64) < 0 ? ceil(FS(f64)) : floor(FS(f64));
+  auto f = FS(f64);
+  if (isinf(f) || isnan(f) || f >= 0x1p+63 || f < -0x1p-63) {
```
invertego:
```suggestion
  if (isinf(f) || isnan(f) || f >= 0x1p+63 || f < -0x1p+63) {
```

PR #729, opened 2022-09-10
ym2612: fix pm vibrato effect

TascoDLX:
This fixes phase modulation (vibrato) effects, like the whistle sound in MD Striker. Both pitch and timbre should be correct now.


BiZkViT96:
Yeah, I was confused why the bug report was closed despite the fact that the effect was still incorrect. But I thought - probably if dev said that it was fixed then its fixed.

PR #767, opened 2022-09-16
n64: fix interlaced display

rasky:
It was completely wrong in my last patch, plus it was rendering at 30fps for unknown reasons.

I also fixed a buffer overflow in mame RDP.

### DIFF
--- a/ares/n64/vi/vi.cpp
+++ b/ares/n64/vi/vi.cpp
@@ -69,17 +69,14 @@ auto VI::main() -> void {
   if(++io.vcounter >= (Region::NTSC() ? 262 : 312) + io.field) {
     io.vcounter = 0;
     io.field = io.field + 1 & io.serrate;
-    if(!io.field) {
-      #if defined(VULKAN)
-      if (vulkan.enable) {
-        gpuOutputValid = vulkan.scanoutAsync(io.field);
-        vulkan.frame();
-      }
-      #endif
-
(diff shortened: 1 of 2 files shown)


DerekTurtleRoe:
I tested every applicable game, and all regions of said games. Everything except for Donkey Kong 64, All Star Baseball 99, 2000, and 2001, and NBA Jam 99 and 2000 all work flawlessly! The only reason I didn't test those other games is because they hang Ares at boot, but that is an unrelated issue. Nice work @rasky ! 😄 

Edit: No version of any of those broken games works, except the kiosk demo of Donkey Kong 64, I need to investigate this more.

ghost:
Ganbare Goemon - Neo Momoyama Bakufu no Odori (Japan) has some issues now, this game uses `OS_VI_NTSC_LAF1`. 

The game can also vary between running at a 60FPS cap and a 30FPS cap depending on if you are on menus/special levels, the screenshots I took (using the Ares screenshot function) are at the 30FPS cap areas.

The image after is also being displayed at a "higher" resolution, hence the 640x480 image.

Note that the international releases use `OS_VI_NTSC_LAN1` thereby fixing this issue.

**Before:**
Before

**After:**
After

rasky:
@Fluvian looks like I was checking the current field register wrong, since when I get to render, it's been already initialized for the new frame. Fixed in the latest version, thanks!

PR #789, opened 2022-09-23
Fix z80 timings

raddad772:
Recently, a contributor tried to fix some errant Z80 timings by readjusting, for example, 'RLC_irr" to use "RLC_irr_r" with a blank final argument. This is incorrect, as RLC_irr takes 21 cycles, but RLC_irr_r takes 23 cycles. This also applies to RRC_, RL_, RR_ SLA_, SRA_ SLL_, SRL_, RES_, and SET_.

Also, BIT_o_irr_r is not an instruction on a Z80. Despite having the same encoding as RLC_irr, etc. above, it just acts like normal BIT_o_irr. So I removed BIT_o_irr_r by renaming to BIT_o_irr, and corrected cycle timings to 14 instead of 15.

I re-instituted the old RLC_irr etc. functions, and modified the _irr_r versions to take the correct amount of cycles.
I investigated where the extra cycles come from using VisualZ80.

I am coding my own suite of unit tests for Z80 based on Ares, and also my own Z80 core; these discrepancies were found in collaboration with @TomHarte .

I do not have a build setup for Ares, so a typo may have made it in that would be found on hitting compile. Please do test before merge


raddad772:
Sorry, didn't get timings 100% right

Issue #680, opened 2022-08-22
[MEGADRIVE] Freeze on the press start screen - Wonder Boy V - Monster World III (USA, Europe) ... 


BenjaminSiskoo:
Emulator : Ares 1.29
Nightly build 2022-08-20 21:03:27 UTC
Options : Default
Game : Wonder Boy V - Monster World III (Japan, Korea), Wonder Boy in Monster World (USA, Europe)
System : Megadrive / Genesis

Freeze on the press start screen. the game works with : Wonder Boy V - Monster World III (Japan, Korea) (Beta) (1991-07-25).

TascoDLX:
Freeze after pressing start likely means eeprom failure caused by incorrect header (misidentified as sram). The UE version is already fixed up in Mia, but the J version needs to be added w/ same params as UE.

Wonder Boy in Monster World (USA, Europe)
6b2ac36f624f914ad26e32baa87d1253aea9dcfc13d2a5842ecdd2bd4a7a43b9

Wonder Boy V - Monster World III (Japan, Korea)
d758864efa18da7a8a87dce32a5b61a8067fb60846696a9588956bc316e480f9

Shideravan:
Fixed in #800 
![image](https://user-images.githubusercontent.com/12163923/192912091-b3974524-d946-4bd5-839f-d65b9e262777.png)
![image](https://user-images.githubusercontent.com/12163923/192912147-5133758c-22d5-40b4-b739-59b5977c6292.png)

PR #800, opened 2022-09-28
MD/32X fixes

TascoDLX:
1. Fixes the bits read from the 32X adapter control register (external), confirmed by the 32X Mars Check Program.
2. Serial communication between paired SH-2's is fixed, confirmed by Mars Check. This also fixes the crash when starting the X-Men 32X prototype.
3. Added eeprom fix for the J version of Wonder Boy V to satisfy #680. The UE version was already covered.
4. Added controller workaround for Corpse Killer 32X CD, same that was added for the plain CD version previously (ref: 
commit 0049c89).

Ares now passes 32X Mars Check Program with no explicit errors. Note, there are still some PWM issues that will need to be tackled.

### DIFF
--- a/ares/component/processor/sh2/sh2.cpp
+++ b/ares/component/processor/sh2/sh2.cpp
@@ -50,7 +50,7 @@ auto SH2::power(bool reset) -> void {
   cache = {*this};
   intc = {*this};
   dmac = {*this};
-  sci = {*this};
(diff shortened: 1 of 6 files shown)


fredditch:
Hello, I still have the same freeze problem on [MD] Wonder Boy V (USA Europe), was there some sort of regression or something ?
I don't have this problem on other emulators, but I like Ares very much...

PR #820, opened 2022-10-01
n64: fix 1bit precision error in mame RDP

rasky:
mame RDP initializes the "ONE" combiner slot to 0xFF. Angrylion upstream and parallel RDP both initialize it to 0x100 instead, and tests on real hardware show that they are right.

Test ROM:
[testrom_emu.zip](https://github.com/ares-emulator/ares/files/9690266/testrom_emu.zip)


rasky:
Retracted, I'll do further tests

PR #822, opened 2022-10-03
Aspect ratio correction for MSX, MSX2, Game Gear, and ColecoVision

PeterSamuels:
Change setAspect for MSX, MSX2, Game Gear, and ColecoVision.

Note: Game Gear has conflicting sources for pixel aspect ratio (8:7 or 6:5) but I believe 6:5 is correct judging from photographs.

### DIFF
--- a/ares/cv/vdp/vdp.cpp
+++ b/ares/cv/vdp/vdp.cpp
@@ -15,7 +15,7 @@ auto VDP::load(Node::Object parent) -> void {
   screen->colors(1 << 4, {&VDP::color, this});
   screen->setSize(256, 192);
   screen->setScale(1.0, 1.0);
-  screen->setAspect(1.0, 1.0);
+  screen->setAspect(8.0, 7.0);
   screen->setViewport(0, 0, 256, 192);
 
   TMS9918::load(screen);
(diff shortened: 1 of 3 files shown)


LukeUsher:
Further confusing the game-gear situation is how common LCD replacements are; and the replacement LCDs have different pixel density, resolution and aspect from the original, so it is important to reference footage of the original screens here.

According to this thread (https://misterfpga.org/viewtopic.php?t=799), the screen is a perfect 4:3 aspect ratio with the 160x144 resolution completely filling the display

> Measuring the screen with a ruler, I get 66 mm × 50 mm—an exact 4:3 ratio.

image

LukeUsher:
Another point of note is that ares's MSX and Master System video emulation does not output or render the horizontal (or in some cases, vertical) border region of the signal; I am in the process of reworking emulation of these systems to output the full frame, so we can implement overscan more accurately.

Would these values still be correct if we did properly output the full visible frame? (Approx 280x240)

There are approx 24-28px of horizontal borders.

Unfortunately; this footage was incorrectly stretched but you can get an idea of just how large the borders were on real hardware from this video taken from a PAL Master System: https://www.youtube.com/watch?v=ZKxLR3Wu1XI

The NTSC system was almost identical, except there is smaller bottom border.

I do have a pal master system and an ntsc master system I can capture frames from if necessary 

As of last night I also managed to get my ntsc-j msx2 working too. So we can get frame captures from there also

PeterSamuels:
RE: Game Gear, yeah, I believe the 6:5 aspect would change it to 4:3.

RE: MSX/SMS, I don't think it would affect anything if I understand how setAspect works.

PR #856, opened 2022-10-21
build: fix windows arm64 build, add to CI

invertego:


nall/GNUmakefile
```diff
@@ -189,7 +189,9 @@ ifeq ($(platform),windows)
   options += $(if $(findstring clang++,$(compiler)),-fuse-ld=lld)
   options += $(if $(findstring g++,$(compiler)),-static -static-libgcc -static-libstdc++)
   options += $(if $(findstring true,$(console)),-mconsole,-mwindows)
-  windres := windres
+  ifeq ($(windres),)
+    windres := windres
+  endif
```
Screwtapello, line 194: Is this necessary? My understanding was that `make windres=blah` will always override whatever value the makefile tries to set it to; you don't need to test for an unset variable.

invertego: You are correct. This change should only affect whether the value of an environment variable named `windres` will be picked up. I found it convenient to just set `compiler` and `windres` in my cross-compiler shell and not have to think about it afterward.

PR #896, opened 2022-12-14
6502: fix adc/sbc flags in decimal mode

invertego:
Validated against https://github.com/TomHarte/ProcessorTests/tree/main/6502/v1


LukeUsher:
The same fix is most likely required in the HuC6280 also?

invertego:
No, it calculates these flags differently.

PR #898, opened 2022-12-17
nall: split implementation into .cpp files

invertego:
Begin the process of splitting nall into separate translation units. It can now be used in two ways:

1) as a header-only library by defining NALL_HEADER_ONLY
2) by compiling nall/nall.cpp

sourcery has been updated to demonstrate the header-only use and everything else now builds nall.cpp.

So far the only things moved into .cpp files are implementations that call Windows APIs. This allows most of ares to build without including windows.h and enables the removal of the abomination that was windows/guard.hpp. The things that actually need windows.h (hiro, ruby) must avoid using names reserved by Windows headers, but the rest of ares is free to use names like "far", "boolean", and "interface" with reckless abandon.

This work is done in preparation for building against MSVC and Windows SDK headers/libraries an an alternative to MinGW-w64.


Kawa-oneechan:
> This work is done in preparation for building against MSVC and Windows SDK headers/libraries an an alternative to MinGW-w64.

Oh heck yeah.

PR #899, opened 2022-12-18
 build: support Clang + MSVC/WinSDK libraries

invertego:
This change makes it possible to build ares using Clang without MinGW-w64. All that is required is a copy of Visual Studio (or Build Tools for Visual Studio) and Clang (also available through the VS installer) along with GNU Make.

To build, just open an x64 Native Tools command prompt and run `make compiler=clang++` as usual.

New CI builds are added to test this configuration for both x64 and ARM64.

### DIFF
--- a/ares/GNUmakefile
+++ b/ares/GNUmakefile
@@ -1,7 +1,5 @@
 ifeq ($(platform),windows)
   options += -luuid -lkernel32 -luser32 -lgdi32 -lcomctl32 -lcomdlg32 -lshell32
-  options += -Wl,-enable-auto-import
-  options += -Wl,-enable-runtime-pseudo-reloc
 else ifeq ($(platform),macos)
 else ifneq ($(filter $(platform),linux bsd),)
   options += -Wl,-export-dynamic
--- a/ares/n64/cpu/interpreter-ipu.cpp
+++ b/ares/n64/cpu/interpreter-ipu.cpp
@@ -252,9 +252,19 @@ auto CPU::DADDU(r64& rd, cr64& rs, cr64& rt) -> void {
 auto CPU::DDIV(cr64& rs, cr64& rt) -> void {
   if(!context.kernelMode() && context.bits == 32) return exception.reservedInstruction();
   if(rt.s64) {
+    #if defined(_MSC_VER)
+    if(rs.s64 != (-1LL << 63) || rt.s64 != -1LL) {
+      LO.u64 = rs.s64 / rt.s64;
+      HI.u64 = rs.s64 % rt.s64;
+    } else {
+      LO.u64 = rs.s64;
+      HI.u64 = 0;
+    }
+    #else
     //cast to i128 to prevent exception on INT64_MIN / -1
     LO.u64 = s128(rs.s64) / s128(rt.s64);
     HI.u64 = s128(rs.s64) % s128(rt.s64);
+    #endif
   } else {
     LO.u64 = rs.s64 < 0 ? +1 : -1;
     HI.u64 = rs.s64;
--- a/ares/sfc/expansion/21fx/21fx.cpp
+++ b/ares/sfc/expansion/21fx/21fx.cpp
@@ -30,8 +30,8 @@ S21FX::S21FX(Node::Port parent) {
 
   string filename{Path::temporary(), "21fx.so"};
   if(link.openAbsolute(filename)) {
-    linkInit = link.sym("fx_init");
-    linkMain = link.sym("fx_main");
+    linkInit = (decltype(linkInit)::cast)link.sym("fx_init");
+    linkMain = (decltype(linkMain)::cast)link.sym("fx_main");
   }
 }
 
--- a/hiro/GNUmakefile
+++ b/hiro/GNUmakefile
@@ -74,9 +74,19 @@ ifeq ($(hiro.resource),)
   hiro.resource := $(hiro.path)/windows/hiro.rc
 endif
 
+ifneq ($(findstring windres,$(windres)),)
+  # windres
+  hiro.resource.extension := .o
+  hiro.resource.command = $1 $2
+else
+  # rc
+  hiro.resource.extension := .res
+  hiro.resource.command = /FO $2 $1
+endif
+
 hiro.objects := \
   $(object.path)/hiro-$(hiro).o \
-  $(if $(filter windows,$(hiro)),$(object.path)/hiro-resource.o)
+  $(if $(filter windows,$(hiro)),$(object.path)/hiro-resource$(hiro.resource.extension))
 
 $(object.path)/hiro-$(hiro).o: $(hiro.path)/hiro.cpp
 	$(if $(filter qt%,$(hiro)),$(info Compiling $(hiro.path)/qt/qt.moc ...))
@@ -84,9 +94,9 @@ $(object.path)/hiro-$(hiro).o: $(hiro.path)/hiro.cpp
 	$(info Compiling $(subst ../,,$<) ...)
 	@$(compiler) $(hiro.flags) $(flags) $(flags.deps) -c $< -o $@
 
-$(object.path)/hiro-resource.o: $(hiro.resource)
+$(object.path)/hiro-resource$(hiro.resource.extension): $(hiro.resource)
 	$(info Compiling $(subst ../,,$<) ...)
-	@$(windres) $< $@
+	@$(windres) $(call hiro.resource.command,$<,$@)
 
 hiro.verbose:
 	$(info hiro Target:)
--- a/hiro/windows/utility.cpp
+++ b/hiro/windows/utility.cpp
@@ -7,12 +7,6 @@ static constexpr u32 Windows7     = 0x0601;
 
 static auto Button_CustomDraw(HWND, PAINTSTRUCT&, bool, bool, bool, u32, const Font&, const image&, Orientation, const string&) -> void;
 
-static auto OsVersion() -> u32 {
-  OSVERSIONINFO versionInfo{0};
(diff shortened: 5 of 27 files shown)


ares/GNUmakefile
```diff
@@ -1,7 +1,5 @@
 ifeq ($(platform),windows)
   options += -luuid -lkernel32 -luser32 -lgdi32 -lcomctl32 -lcomdlg32 -lshell32
-  options += -Wl,-enable-auto-import
```
invertego, line 3: These seem to be very old MinGW options that no longer serve a purpose. They've been untouched for at least 12 years.

ares/n64/cpu/interpreter-ipu.cpp
```diff
@@ -252,9 +252,19 @@ auto CPU::DADDU(r64& rd, cr64& rs, cr64& rt) -> void {
...
+    if(rs.s64 != (-1LL << 63) || rt.s64 != -1LL) {
+      LO.u64 = rs.s64 / rt.s64;
+      HI.u64 = rs.s64 % rt.s64;
+    } else {
+      LO.u64 = rs.s64;
+      HI.u64 = 0;
+    }
+    #else
     //cast to i128 to prevent exception on INT64_MIN / -1
     LO.u64 = s128(rs.s64) / s128(rt.s64);
```
invertego, line 265: Rather than emitting extended division instructions, Clang was emitting calls to a builtins library that isn't linked in automatically when using the msvc target. It's debatable whether this optimization is really significant in the first place.

ares/sfc/expansion/21fx/21fx.cpp
```diff
@@ -30,8 +30,8 @@ S21FX::S21FX(Node::Port parent) {
 
   string filename{Path::temporary(), "21fx.so"};
   if(link.openAbsolute(filename)) {
-    linkInit = link.sym("fx_init");
-    linkMain = link.sym("fx_main");
+    linkInit = (decltype(linkInit)::cast)link.sym("fx_init");
```
invertego, line 33: Casts are now required because I removed the void* assignment operator.

hiro/GNUmakefile
```diff
@@ -74,19 +74,29 @@ ifeq ($(hiro.resource),)
   hiro.resource := $(hiro.path)/windows/hiro.rc
 endif
 
+ifneq ($(findstring windres,$(windres)),)
```
invertego, line 77: You don't get (the LLVM version of) windres if you just install Clang by itself. Well, technically you do, because the same llvm-rc executable is capable of operating as a replacement for windres (the GCC resource compiler) or rc (the MSVC resource compiler) based on the name of the executable, but there's no way to override which one it masquerades as - it looks at argv[0] to decide.

Using llvm-rc as an rc replacement is actually preferable to windres in some ways, though. It outputs an architecture agnostic .res file instead of an .o file, and the latter just adds an extra wrinkle when cross compiling.

hiro/windows/utility.cpp
```diff
@@ -179,7 +173,7 @@ static auto CALLBACK Shared_windowProc(WindowProc windowProc, HWND hwnd, UINT ms
   if(!window) return DefWindowProc(hwnd, msg, wparam, lparam);
   auto pWindow = window->self();
   if(!pWindow) return DefWindowProc(hwnd, msg, wparam, lparam);
-
+
```
invertego, line 176: This file had mixed LF and CRLF line endings that my editor cleaned up.

hiro/windows/utility.cpp
```diff
@@ -7,12 +7,6 @@ static constexpr u32 Windows7     = 0x0601;
 
 static auto Button_CustomDraw(HWND, PAINTSTRUCT&, bool, bool, bool, u32, const Font&, const image&, Orientation, const string&) -> void;
 
-static auto OsVersion() -> u32 {
-  OSVERSIONINFO versionInfo{0};
-  versionInfo.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
-  GetVersionEx(&versionInfo);
```
invertego, line 13: The call to GetVersionEx was causing deprecation warnings. I starting looking into how to silence those when I realized this code is entirely unused, so I just removed it.

nall/GNUmakefile
```diff
@@ -187,12 +198,20 @@ endif
...
 # windows settings
 ifeq ($(platform),windows)
-  options += -mthreads -lpthread -lws2_32 -lole32
+  extension := .exe
+  ifeq ($(msvc),true)
+    flags += -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_WARNINGS -DNOMINMAX
+  endif
+  options += -mthreads -lws2_32 -lole32 -lshell32
   options += $(if $(findstring clang++,$(compiler)),-fuse-ld=lld)
-  options += $(if $(findstring g++,$(compiler)),-static -static-libgcc -static-libstdc++)
```
invertego, line 192: A couple of things to note:
- Due to "g++" being a substring of "clang++", this was being applied to both compilers.
- On my machine, -static-libgcc and -static-libstdc++ weren't having any effect on the build output with either GCC or Clang. I removed them because they caused unused option warnings in the new build configuration.

nall/chrono.hpp
```diff
@@ -3,14 +3,15 @@
 #include <nall/function.hpp>
 #include <nall/string.hpp>
 
+#include <chrono>
+
 namespace nall::chrono {
 
 //passage of time functions (from unknown epoch)
 
 inline auto nanosecond() -> u64 {
```
invertego, line 12: MSVC doesn't implement this POSIX API. I initially had a _MSC_VER check here but I couldn't see any reason not to just use std::chrono everywhere. This also allowed me to stop linking against winpthreads in the existing Windows build configuration (which is where the MinGW implementation of this API resides) as it was our only dependency on that library.

nall/function.hpp
```diff
@@ -37,12 +36,6 @@ template<typename R, typename... P> struct function<auto (P...) -> R> {
     return *this;
   }
 
-  auto operator=(void* source) -> function& {
```
invertego, line 40: Clang couldn't decide which assignment operator to use when this one was an option. Casting function pointers to void* is implementation-defined and kind of frowned upon anyway, so I decided to just drop it.

nall/platform.cpp
```diff
@@ -20,6 +20,14 @@ NALL_HEADER_INLINE auto setsockopt(int socket, int level, int option_name, const
   return ::setsockopt(socket, level, option_name, (const char*)option_value, option_len);
 }
 
+NALL_HEADER_INLINE auto usleep(unsigned int us) -> int {
```
invertego, line 23: This is just what the MinGW-w64 implementation of usleep() happens to do; I didn't want to change behavior here. At some point, it would probably be wise to evaluate everywhere ares is doing sleeps and seek better alternatives.

nall/serializer.hpp
```diff
@@ -76,10 +76,11 @@ struct serializer {
   }
 
   template<typename T> auto operator()(T& value) -> serializer& {
-    static_assert(has_serialize_v<T> || is_integral_v<T> || is_floating_point_v<T>);
+    constexpr bool integral = is_integral_v<T> || is_same_v<T, u128>;
```
invertego, line 79: Yet another way __int128 is sketchy. I'd like to get rid of the remaining uses of 128-bit types.

https://quuxplusone.github.io/blog/2019/02/28/is-int128-integral/

LukeUsher:
> This change makes it possible to build ares using Clang without MinGW-w64. All that is required is a copy of Visual Studio (or Build Tools for Visual Studio) and Clang (also available through the VS installer) along with GNU Make.
> 
> To build, just open an x64 Native Tools command prompt and run `make compiler=clang++` as usual.
> 
> New CI builds are added to test this configuration for both x64 and ARM64.

In this scenario where does make come from? using Visual Studio native tools there is no make in PATH; and nmake is not compatible with gnumake.

nall/platform.cpp
LukeUsher: consider std::this_thread::sleep_for() ?

LukeUsher:
I tested this locally by using mingw32-make from my msys installation, but we can't assume that to be present for most people; since MSVC doesn't provide a gnumake, people wishing to compile will still require their own; it doesn't come with clang.

I ran this from the visual studio command prompt, without msys in my path
`C:\msys64\clang64\bin\mingw32-make.exe compiler=clang++ -j14`

invertego:
Yeah, I'm aware. Ive been pondering where to advise people to acquire make. There are plenty of prebuilt copies floating around the internet. One option would be to direct people to Chocolatey (`choco install make`) but not everyone has that installed.

Long term I'd prefer to replace the GNU Make dependency with CMake, or at least offer CMake as an alternative.

invertego:
choco offers make 4.3.
scoop offers make 4.4.
winget offers make 3.81 - the GnuWin32 release that dates to 2006. 🙃

invertego:
The copy from the msys2 package works standalone, but it's a bit of a struggle to extract thanks to the zstandard compression.
https://mirror.msys2.org/mingw/mingw64/mingw-w64-x86_64-make-4.4-2-any.pkg.tar.zst

The copy provided by scoop is hosted in a zip on SourceForge which is pretty easy to link users to directly.
https://sourceforge.net/projects/ezwinports/files/make-4.4-without-guile-w32-bin.zip/download

nall/platform.cpp
invertego: Yeah, that seems like a good candidate, but I'd leave it for a separate change.

LukeUsher:
I do like the idea of moving to cmake, it has been discussed in the ares discord a couple of times, and I did get it working on Windows already in an experimental fork, it might be time to revisit it now.

PR #906, opened 2022-12-26
ws: debugger/UI additions + emulation bugfixes batch

asiekierka:
tl;dr:

### V30MZ

* Fixed low-hanging CPU timing fruit, based on [WSTimingTest](https://github.com/FluBBaOfWard/WSTimingTest). There's still some tests not passing, most notably jumps.
* Fixed IN/OUT opcode decoding in the CPU tracer.
* Fixed emulation of undocumented opcode 0xD6 (while on NEC V20/V30 CPUs it is a mirror for XLAT, on NEC V30MZ it is actually the undocumented Intel 8086/80186 opcode SALC) as well as 0xF7 subop 1 (V20/V30 has it as a mirror for TEST, but on V30MZ it is simply a NOP).

### WonderSwan (emulation)

* Tweaked APU code to properly treat it as a 24000 Hz digital chip, as opposed to a 3072000 Hz analog chip - this fixes my own [Headphones' Warming Eve](https://demozoo.org/productions/316722/) demo and should overall bring audio accuracy closer to the device's own. It is theoretically possible to get the APU code to give *exact* matching input, as the output procedure is well-understood and the EXT port provides digital PCM samples - however, more research is required to get there. (This improves performance by a solid ~10% due to the significant amount of calculations that can be skipped.)
* Fixed APU channel 2 "half-volume" bits being ignored.
* Tentatively tweaked low-pass filter value based on [GuyPerfect's research](http://perfectkiosk.net/stsws.html#sound_output_procedure).
* Implemented a less stubby serial I/O stub. This properly emits TX/RX interrupts. (This probably also hurts performance a little... you gain some, you lose some.)
* Added support for Bandai 2003's 16-bit banking ports - this is used by exactly zero commercial games, but reverse-engineering has shown that the chip supports this.
* Added stubby support for NMI register 0xB7, as found and documented by FluBBa in the NitroSwan emulator.
* Appropriately gate WSC-mode I/O ports from Mono mode.
* Fixed behavior of the HBlank/VBlank PPU timers when the reload value is 1, but the timer is not enabled. Existing reverse-engineered documentation implies that this won't fire an interrupt, but hardware testing proves that it indeed does.
* Fixed APU internal speaker output logic.
* Added support for "pixel-inaccurate" mode; this makes the PPU step at a per-scanline level, and the APU step at a per-sample level. Improves performance by ~65% on my machine due to avoiding all the context switches.
* Fixed keypad interrupt emulation to match hardware quirks.
* Fixed handling the interrupt base port on WSC.

### WonderSwan (UI)

* Added PPU SCREEN1, SCREEN2 and tiles to the Graphics debug view.
* Added PPU, APU, cartridge, serial and SoC I/O ports to the Properties debug view. This covers most, but not all, of the device's I/O ports...
* Exposed the ability to use internal speaker output in the UI.

The APU changes in particular could use more verification. In total, in "pixel-inaccurate" mode, this should double the core's performance over ares v130.

PS. Can I just say that the Ares codebase is an absolute joy to work with? Seriously, good job to everyone involved.

### DIFF
--- a/ares/component/processor/v30mz/instructions-misc.cpp
+++ b/ares/component/processor/v30mz/instructions-misc.cpp
@@ -21,7 +21,7 @@ auto V30MZ::instructionLock() -> void {
 }
 
 auto V30MZ::instructionWait() -> void {
-  wait(1 + 9);
+  wait(10);
 }
 
 auto V30MZ::instructionHalt() -> void {
--- a/ares/component/processor/v30mz/instructions-misc.cpp
+++ b/ares/component/processor/v30mz/instructions-misc.cpp
@@ -21,7 +21,7 @@ auto V30MZ::instructionLock() -> void {
 }
 
 auto V30MZ::instructionWait() -> void {
-  wait(1);
+  wait(1 + 9);
 }
 
 auto V30MZ::instructionHalt() -> void {
@@ -30,6 +30,10 @@ auto V30MZ::instructionHalt() -> void {
 }
 
 auto V30MZ::instructionNop() -> void {
+  wait(1);
+}
+
+auto V30MZ::instructionUndefined() -> void {
   wait(1);
 }
 
--- a/ares/component/processor/v30mz/instructions-misc.cpp
+++ b/ares/component/processor/v30mz/instructions-misc.cpp
@@ -57,8 +57,13 @@ template<u32 size> auto V30MZ::instructionOutDW() -> void {
   out<size>(DW, getAccumulator<size>());
 }
 
-auto V30MZ::instructionTranslate(u8 clocks) -> void {
-  wait(clocks);
+auto V30MZ::instructionSetALCarry() -> void {
+  wait(8);
+  AL = PSW.CY ? 0xFF : 0x00;
+}
+
+auto V30MZ::instructionTranslate() -> void {
+  wait(4);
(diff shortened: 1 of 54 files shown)


asiekierka:
(Don't worry about the additional commits; merge whenever you feel comfortable with the contents thus far, or when there's a new version soon to be released, or...)

ares/component/processor/v30mz/instructions-misc.cpp
```diff
@@ -21,7 +21,7 @@ auto V30MZ::instructionLock() -> void {
 }
 
 auto V30MZ::instructionWait() -> void {
-  wait(1);
+  wait(1 + 9);
```
LukeUsher: Any reason why 1 +9 rather than 10?

asiekierka: It's not strictly necessary, but the V30MZ datasheet lists the timing for this opcode as the rather cryptic `Number of 9 + 9 × POLLB pin samplings`, which I *assume* means "1 + 1 * 9" (this makes the timing match hardware, anyway), so I wanted to reflect that here.

asiekierka: But I changed it, as I don't strongly feel about it either way.

PR #907, opened 2022-12-27
gb: fix MBC1 mapper implementation for 8/16Mbit cartridges

asiekierka:
This change can be validated with [the GBEmulatorShootout's test suite compilation](https://daid.github.io/GBEmulatorShootout/) ([MBC1 test ROMs](https://github.com/daid/GBEmulatorShootout/tree/main/testroms/mooneye/emulator-only/mbc1)) and [the Pan Docs](https://gbdev.io/pandocs/MBC1.html).

### DIFF
--- a/ares/gb/cartridge/board/mbc1.cpp
+++ b/ares/gb/cartridge/board/mbc1.cpp
@@ -17,23 +17,23 @@ struct MBC1 : Interface {
 
   auto read(n16 address, n8 data) -> n8 override {
     if(address >= 0x0000 && address <= 0x3fff) {
-      return rom.read((n14)address);
+      if(io.mode) {
+        return rom.read(io.ram.bank << 19 | (n14)address);
+      } else {
+        return rom.read((n14)address);
+      }
     }
 
     if(address >= 0x4000 && address <= 0x7fff) {
-      if(io.mode == 0) {
-        return rom.read(io.ram.bank << 19 | io.rom.bank << 14 | (n14)address);
-      } else {
-        return rom.read(io.rom.bank << 14 | (n14)address);
-      }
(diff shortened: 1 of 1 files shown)


ares/gb/cartridge/board/mbc1.cpp
```diff
@@ -17,23 +17,23 @@ struct MBC1 : Interface {
 
   auto read(n16 address, n8 data) -> n8 override {
     if(address >= 0x0000 && address <= 0x3fff) {
-      return rom.read((n14)address);
+      if(io.mode) {
+        return rom.read(io.ram.bank << 19 | (n14)address);
+      } else {
```
LukeUsher: since both of these paths return, consider removing the else case for slightly cleaner code, this better matches coding style in the rest of ares

```c++
if(io.mode) return rom.read(io.ram.bank << 19 | (n14)address);
return rom.read((n14)address);
```

ares/gb/cartridge/board/mbc1.cpp
```diff
@@ -17,23 +17,23 @@ struct MBC1 : Interface {
...
     }
 
     if(address >= 0xa000 && address <= 0xbfff) {
       if(!ram || !io.ram.enable) return 0xff;
-      if(io.mode == 0) {
-        return ram.read((n13)address);
-      } else {
+      if(io.mode) {
         return ram.read(io.ram.bank << 13 | (n13)address);
+      } else {
```
LukeUsher: Same here too

ares/gb/cartridge/board/mbc1.cpp
asiekierka: Done.

PR #912, opened 2022-12-28
v30mz, ws: fix bugs and improve emulation accuracy

### LINKED ISSUE #879: [WonderSwan] Mahjong Touryuumon (Japan): Runs faster than it should, missing music/audio
**Describe the bug**
Mahjong Touryuumon runs noticeably far faster than it should and is missing audio. Occurs on both Rev 1 and Rev 3.

**To Reproduce**
Start the game and mash through the first few screens. Most obvious with the portrait bobbing on the character select.

**Expected behavior**
I expected the game to run at the proper speed and play audio.

Tested on Ares v130.1 built November 23, 2022
Tested on Linux Mint 21

asiekierka:
## V30MZ

* **Fix JL/JGE opcodes acting like JLE/JG opcodes**
* Fix flag calculation in opcodes to match hardware: ADC, SBB, MUL, IMUL, AAD, RCL, RCR, ROL, ROR, SHL, SHR, SHRA, DAA, DAS, AAA, AAS
* Fix opcode INTO implementation
* Fix undefined opcodes 0x0F, 0xCxF0, 0xD8-0xDF, 0xFFF8 implementation
* Improve cycle timing for software-caused interrupts
* Improve cycle timing for certain effective address calculations
* Implement BRK flag handling

## WonderSwan

* Implement SRAM and cartridge ROM wait states. This fixes #879 
* Prevent access to Color-only RAM on Mono-mode WSC.
* Minor debugger fixes.

Special thanks to FluBBa, without whose work many of these bugs and inaccuracies wouldn't have been spotted.

With this, the following areas of inaccuracy remain:

* Division flags and exception handling. There are tests for these, but they will take some effort and time to implement correctly.
* Small timing issues around interrupts, calls, jumps and returns; cause unknown, more research is required.
* Interrupt priorities aren't taken into account when multiple interrupt sources happen on the same cycle.
* Anything we don't know about yet as a community.


asiekierka:
This one has gotten large enough.

PR #918, opened 2022-12-31
fc: use database for unheadered roms

invertego:


mia/medium/famicom.cpp
```diff
@@ -38,8 +38,11 @@ auto Famicom::load(string location) -> bool {
 
   array_view<u8> view{rom};
   if(auto node = document["game/board/memory(type=ROM,content=iNES)"]) {
-    pak->append("ines.rom", {view.data(), node["size"].natural()});
-    view += node["size"].natural();
+    //check if header is present
```
LukeUsher: I don't think this is the correct place for this code; see how it's within a "content=iNES" check? If the nes is unheadered, iNES shouldn't be set as a content type at all.

invertego: This change is working as intended. That code is there because all the entries in Famicom.bml contain an entry for the iNES header. A database match can occur with a headered or unheadered rom. The alternative to having this check here would be to strip the iNES header out of the database entries and add it back dynamically only when the iNES header is detected.

I guess the other option would be to strip out the iNES header entirely (if present) if there is a database match.

LukeUsher: Perhaps add a code comment to clarify this because it is confusing to see a check for a header after a check for Ines

invertego: I opted to remove the iNES entries from the database and add back the entry only when it is present. It makes this change a bigger but now the manifests always match the ROM contents, which is a bit cleaner.

mia/medium/famicom.cpp
```diff
@@ -132,7 +134,13 @@ auto Famicom::analyzeFDS(vector<u8>& data) -> string {
 auto Famicom::analyzeINES(vector<u8>& data) -> string {
   string hash = Hash::SHA256({data.data() + 16, data.size() - 16}).digest();
   string manifest = Medium::manifestDatabase(hash);
-  if(manifest) return manifest;
+  if(manifest) {
+    manifest += "    memory\n";
```
invertego, line 138: This will add an entry to the end, obviously. Order doesn't matter to the code that uses the manifest. I don't think this is a big deal, but I wanted to call it out.

Issue #427, opened 2022-03-06
[SNES] CPUMSC reset test - doesn't work [ares-v127]


Max833:
> https://github.com/PeterLemon/SNES/blob/master/CPUTest/CPU/MSC/CPUMSC.sfc
> 
> This rom has a test for the snes reset feature which does not work, instead of resetting the system and showing that the test has passed it will get stuck displaying garbage graphics.
> 
> It works on the sd2snes, has been reported as working in bsnes sometime in the past and it still works in the mednafen bsnes fork.
> 
> **[...]**
> 
> Its fixed in the jgemu bsnes fork.
> 
> https://gitlab.com/jgemu/bsnes/-/commit/3227e0102f1453b8ff2e9217c7a826d7fb38874a

Source: https://github.com/bsnes-emu/bsnes/issues/235

remutro:
@Max833 - What are the expected results of this test? Do you have a screenshot? 

I tried using the latest version of BSNES, and after loading the ROM and choosing the reset option as instructed it just gives a solid blue screen. 

Same result in Ares. The code is very much the same here in Ares still, and applying the changes that were made in jgemu's fork results in getting a green screen after resetting, instead of blue. Does green mean good/pass?

orbea:
This is the pass screen as shown in bsnes-jg.
![cpumsc](https://user-images.githubusercontent.com/4204285/210254442-b7847574-5ee0-4bcd-af70-357dd17af5ae.png)

remutro:
Thanks. Applying the same changes does allow the test to pass....eventually. But there seems to be another difference here. The colors end up changing, and depending on the what colors are in use you may not see any results. But resetting keeps rotating through colors and eventually you can see the results. So there is something else interfering with this test.

Also worth noting that nothing will display after resetting without "pixel accuracy" being used (black screen using scanline renderer).

![cpu](https://user-images.githubusercontent.com/84990500/210263110-b46aef22-b895-4f53-96ee-2644a2fb3aa0.png)

orbea:
The bsnes-jg fork only has the accurate ppu. I would guess something is being reset that should not be, I had very similar nondeterministic results when troubleshooting the patch for the jg fork.

remutro:
@orbea @Max833
Aha! Found the color changing issue. Should also not be randomizing CGRAM either when reset, along with the other !reset checks needed:

``` 
if(!reset) {
   random.array({cgram, sizeof(cgram)});
[...]

PR #928, opened 2023-01-03
Fix for Peter Lemon's SNES CPU reset test

remutro:
This is a fix for a reset test in Peter Lemon's CPU test ROM as defined in Github issue: https://github.com/ares-emulator/ares/issues/427 It addresses both pixel renderer and scanline renderer.

This has been discussed as a BNES issue previously, there has been a fix to a fork of bsnes that addresses this, and a similar fix was committed to BizHawk. The only problem is I can't to attempt to validate this behavior on real hardware (although the test itself _does_ work on real hardware). 

I'm going to open a discussion about this in our Discord to get others input, so it may be best to hold off on merging until others get a chance to chime in on this.

### DIFF
--- a/ares/sfc/ppu-performance/ppu.cpp
+++ b/ares/sfc/ppu-performance/ppu.cpp
@@ -118,7 +118,7 @@ auto PPU::map() -> void {
 auto PPU::power(bool reset) -> void {
   Thread::create(system.cpuFrequency(), {&PPU::main, this});
   PPUcounter::reset();
-  screen->power();
+  if(!reset) screen->power();
 
   ppu1.version = 1, ppu1.mdr = 0x00;
   ppu2.version = 3, ppu2.mdr = 0x00;
@@ -131,15 +131,18 @@ auto PPU::power(bool reset) -> void {
   latch = {};
   io = {};
   mode7 = {};
-  window.power();
-  mosaic.power();
-  bg1.power();
-  bg2.power();
-  bg3.power();
-  bg4.power();
-  obj.power();
-  dac.power();
-
+  
+  if(!reset) {
+    window.power();
+    mosaic.power();
+    bg1.power();
+    bg2.power();
+    bg3.power();
+    bg4.power();
+    obj.power();
+    dac.power();
+  }
+  
   updateVideoMode();
 
   string title;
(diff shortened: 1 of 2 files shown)


Max833:
It is working with bsnes v070. Something caused a regression over the years it seems like.

image

orbea:
> Something caused a regression over the years it seems like.

It was when the accurate PPU was added, I spent some time bisecting this a while ago.

Max833:
I see, interesting.

> This is a fix for a reset test in Peter Lemon's CPU test ROM as defined in Github issue: https://github.com/ares-emulator/ares/issues/427 It addresses both pixel renderer and scanline renderer.

So this means this test is working on real hardware, but we don't know for sure how it should be implemented on emulators so it's true to the EXACT behaviour of a real SNES?

The first thing that comes to my mind is this game that was hanging if you reset the system on Mesen-S (and that shouldn't be the case). Is this still working with bsnes-jg after the fix?

https://github.com/SourMesen/Mesen-S/issues/58

orbea:
> So this means this test is working on real hardware, but we don't know for sure how it should be implemented on emulators so it's true to the EXACT behaviour of a real SNES?

I think there is some nuance to this.

* We know the test works on real hardware
* We know the current bsnes behavior here is not accurate
* We know that the fix makes bsnes' behavior more accurate than it was before
* We don't know that if the new behavior will be 100% accurate or that it could be improved even more

So fixing this is a step in the right direction at the very least.

>  Is this still working with bsnes-jg after the fix?

Yes, it was working on my system as of a few days ago and there hasn't been any changes since.

Max833:
That's a good start at least. Which games require a reset as well? What's about SPC7110 games, do they work?

Maybe @PeterLemon can add some more detail to his test ROM as well.

orbea:
> Which games require a reset as well? 

 I am not sure there are any commercial content that depends on this working correctly.

> What's about SPC7110 games, do they work?

`Tengai Makyō Zero`was fixed for bsnes-jg a while ago, but I don't think it was ever fixed for this repo. Regardless that is an unrelated issue. I also tested `Momotaro Dentetsu Happy` and `Super Power League 4` now, they seem to work as expected.

carmiker:
I personally tested CPUMSC on both an original SNES and Super Nt, and this fix behaves exactly the same as on both consoles.

Ultimately, what matters is that the state of the registers when the game code begins running is what it should be. Things do not map as neatly from functions in an emulator to hardware as precisely as some may think. 

We took extreme caution when doing `bsnes-jg` to ensure there were no regressions, and in fact we fixed all known regressions we inherited with the codebase. We also removed all speed hacks aside from the optional coprocessor HLE. The last thing you could call a regression is support for competition carts, which was removed after higan v106, and I plan to reimplement that at some point in the future.

Anyhow, @LukeUsher, I endorse merging this pull request.

Issue #605, opened 2022-06-02
[N64] Donkey Kong 64 no longer boots


kev4cards:
As first noted by rasky, all three regions of Donkey Kong 64 no longer boot as of commit 57b4d68.

invertego:
Here is the root cause analysis I did last year.

> I'm looking at the regressions in DK64 and Banjo-Tooie caused by PI DMA being slowed down by a factor of 4x.
And in the case of DK64, it looks like a pretty straightforward race between a PI DMA of 0x100000 bytes and this loop running on the RSP:
```
1e4  lui     a1,$0020
1e8  bltz    a1,$258
1ec  mfc0    v1,SP_SEMAPHORE
1f0  bne     v1,0,$1e8
1f4  subi    a1,a1,$1
```
 
> If the bltz is taken (after 0x200000 loop iterations) before the PI DMA is done, there's no recovering from it.
During this time the CPU is spinning on the PI DMA, after which it releases the semaphore.
So basically... either this loop is consuming at least 6 RSP cycles per iteration, or PI DMA is a lot faster than it is currently implemented in ares.

```
X * 3 * 0x200000 > Y * 2 * 0x100000
X = RSP cycles per loop iteration
Y = CPU cycles per byte during PI DMA
```

Thanks to @rasky we know the timing on real hardware: X = 7 and Y = ~18. Again, thanks to rasky, ares emulates PI DMA timing quite well. However, it does not accurately track RSP cycles when pipeline stalls must be taken into account (in this case due to the mfc0 and bne instructions).

LukeUsher:
Some additional info, provided by Rasky on the ares discord server

> there are two things that need to be implemented in the RSP interpreter/recompiler for the DK64 boot to succeed:
> * taken branches must have 1 cycle penalty. This should be ~easy to implement on both the interpreter and the recompiler
> * mfc0 must have 3 cycle latency with stall. The code in the 6105 bootcode (used by DK64) does a mfc0 immediately followed  by a use of the register, so this must incur in 2 cycle stall. 
> 
> implementing both will make the dk64 bootcode slow down enough to lose the race with PI DMA, which is what happens on real hardware 
> 
> notice that this might have side effects on the RSP emulator, as effectively we are implementing two slow down behaviors  and our RSP is already slower than the real hardware because we are ignoring the pairing optimization between VU and SU; so I don't know if just doing the above would then have other side effects where the RSP is too slow for some games

PR #931, opened 2023-01-05
n64: Changed PI DMA timings to match MAME implementation

### LINKED ISSUE #633: [N64] Banjo-Tooie soft-locked at title screen on July 7th nightly
From today's nightly build cb4a79e , the game now starts with the camera at the origin-point instead of doing the title screen sequence.
This also happens in Dillonb's emulator between updates.

I got the nightly build to test DK64 due to changes days after V128 was released, but its on a black screen so it doesn't even boot now which is also what happened between updates on Dillonb's emulator as well.

mystise:
Fixes #633 and #605

Changes PI timing to account for the extra byte and utilizes the MAME per-byte cycle timings.

Additional commentary at: https://github.com/ares-emulator/ares/issues/633#issuecomment-1371897412

### DIFF
--- a/ares/n64/pi/io.cpp
+++ b/ares/n64/pi/io.cpp
@@ -94,11 +94,16 @@ auto PI::ioWrite(u32 address, u32 data_) -> void {
     io.pbusAddress = n32(data) & ~1;
   }
 
+  // DMA Timing notes:
+  // io.readLength and io.writeLength are 1 byte less than the actual value of bytes read or written
+  // According to MAME's implementation (linked below), each byte transfered during a DMA takes ~5.08 CPU cycles
+  // https://github.com/mamedev/mame/blob/ab6237da824a95898209ecbb568095981c4dbe69/src/mame/nintendo/n64_m.cpp#L1609
+  // Since ares cycles are 2x that (see CPU::synchronize() in cpu.cpp), multiply by 10 (~10.16)
   if(address == 2) {
     //PI_READ_LENGTH
     io.readLength = n24(data);
     io.dmaBusy = 1;
-    queue.insert(Queue::PI_DMA_Read, io.readLength * 36);
+    queue.insert(Queue::PI_DMA_Read, (io.readLength + 1) * 10);
     dmaRead();
   }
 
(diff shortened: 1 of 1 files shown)


rasky:
The PI timings are pretty well understood nowadays. These are some traces:
https://n64brew.dev/wiki/Peripheral_Interface#PI_Interface_Process

The actual timing are configured by the PI DOM1/DOM2 configuration registers (https://n64brew.dev/wiki/Peripheral_Interface#Domains). DOM1 covers most of the address space, so let's focus on that.

All commercial games configure the various fields in DOM1 registers as follows (notice all registers store "minus 1" values):
```
LAT = 0x40 = 65 RCP cycles
PWM = 0x12 = 19 RCP cycles
PGS = 0x07 = 1<<8 = 256 16-bit words pages size = 512 bytes page size
RLS = 0x03 = 4 RCP cycles
```

On top of this, there are two fixed waits that can't be configured.

```
ALEL = 7 RCP cycles
ALEH = 7 RCP cycles
```

The formula is: for each page, `ALEH + ALEL + LAT + (PWM + RLS) * N` where `N` is the number of 16-bit words transferred in that page. Pages are always aligned to their size, so the actual number of bytes per page depends on the actual address (a transfer might start mid-first-page and finish mid-last-page, plus covers a certain number of full pages).

For instance, let's assume a 512-byte transfer, fully covering one page. The timing is `7 + 7 + 65 + (19+4)*256` which is 5967 RCP cycles. In "Ares cycles", this would be times 3, so 17901 Ares cycles.

When I studied this, I plotted transfers of a few different sizes and did a linear fit. The linear factor that approximated better the numbers was 36 Ares cycles per byte. For instance, in the above example, it is 36 * 512 = 18432 Ares cycles, which is pretty close (3% error).

The number in this PR (10 Ares cycles per byte) are pretty much off. They would make the above example transfer of 512 bytes take 5120 Ares cycles, which is a 72% error.

To further confirm this, you can use this ROM:
https://github.com/rasky/n64-systembench
[n64-systembench.zip](https://github.com/ares-emulator/ares/files/10352374/n64-systembench.zip)

which is a very simple benchmark. Ares doesn't fare too well in general, but PI DMA numbers specifically are quite good:

Schermata 2023-01-05 alle 14 08 18

So unless if there is further evidence to be brought, I don't think this PR is correct, as it would make the PI timings much worse.

LukeUsher:
The fact that this PR fixed DK64 and Banjo-Tooie is mostly because it changes timing so much that a race condition in the game code no longer occurs.

The reason these games works on real hardware is related to pipeline stalls, and not so much PI timing: in order to properly support these games with more accurate PI timing, the cpu and rsp pipelines need to be better understood, and ares needs to be better at emulating pipeline stalls.

mystise:
Understood, good to know that this only fixes those games by coincidence, and nice to have more details on the actual timings.

PR #944, opened 2023-01-25
n64: initial estimation of PIF timings

rasky:
Numbers are based on this test ROM:
https://github.com/rasky/n64-systembench

There are still many other cases to handle but this should provide the initial framework for further experimentation with PIF timings.


rasky:
Keeping in draft waiting for some testing. I'm sure there will be fallouts, though the numbers seem correct.

rasky:
Testing OK, this is actually an improvement. Ready for review

PR #945, opened 2023-01-27
Radio the scale settings and allow 1x

FitzRoyX:
1x scaling is nice for recording videos. You can achieve the smallest filesize and it's easier to use the recorder app side by side with the emulator.

This PR allows 1x scaling and makes the scale settings explicit and radio. This also allows you to see the current scale setting whereas you couldn't before. "Auto centering" options were removed in favor of allowing 5x to simply exist without doing anything fancy like removing it based on the monitor.

The words "multiplier" and "size" were also used interchangeably with "scale" in the codebase. This PR also changes everything to "scale" as I believe it's the most common term for video and less generic.

### DIFF
--- a/ares/ng/lspc/lspc.cpp
+++ b/ares/ng/lspc/lspc.cpp
@@ -13,10 +13,10 @@ auto LSPC::load(Node::Object parent) -> void {
 
   screen = node->append<Node::Video::Screen>("Screen", 320, 256);
   screen->colors(1 << 17, {&LSPC::color, this});
-  screen->setSize(320, 256);
-  screen->setScale(2.0, 2.0);
+  screen->setSize(304, 240);
+  screen->setScale(1.0, 1.0);
   screen->setAspect(1.0, 1.0);
-  screen->setViewport(8, 8, 320 - 16, 256 - 16);
+  screen->setViewport(8, 8, 304, 240);
 
   vram.allocate(68_KiB >> 1);
   pram.allocate(16_KiB >> 1);
--- a/ares/ng/lspc/lspc.cpp
+++ b/ares/ng/lspc/lspc.cpp
@@ -13,10 +13,15 @@ auto LSPC::load(Node::Object parent) -> void {
 
   screen = node->append<Node::Video::Screen>("Screen", 320, 256);
   screen->colors(1 << 17, {&LSPC::color, this});
-  screen->setSize(304, 240);
+  screen->setSize(320, 256);
   screen->setScale(1.0, 1.0);
   screen->setAspect(1.0, 1.0);
-  screen->setViewport(8, 8, 304, 240);
+
+  overscan = screen->append<Node::Setting::Boolean>("Overscan", true, [&](auto value) {
+    if(value == 0) screen->setSize(304, 224);
+    if(value == 1) screen->setSize(320, 256);
+  });
+  overscan->setDynamic(true);
 
   vram.allocate(68_KiB >> 1);
   pram.allocate(16_KiB >> 1);
@@ -47,6 +52,7 @@ auto LSPC::unload() -> void {
   debugger.unload(node);
   vram.reset();
   pram.reset();
+  overscan.reset();
   screen->quit();
   node->remove(screen);
   screen.reset();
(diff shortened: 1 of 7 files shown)


ares/ng/lspc/lspc.cpp
```diff
@@ -13,10 +13,10 @@ auto LSPC::load(Node::Object parent) -> void {
 
   screen = node->append<Node::Video::Screen>("Screen", 320, 256);
   screen->colors(1 << 17, {&LSPC::color, this});
-  screen->setSize(320, 256);
-  screen->setScale(2.0, 2.0);
+  screen->setSize(304, 240);
```
LukeUsher: Note that for neo geo, some games do intend to use the entire 320x240 framebuffer, while others leave the outer pixels blank

We should use 320x240 here, and as the viewport, rather than clamping to 304.

desktop-ui/presentation/presentation.cpp
```diff
@@ -245,34 +252,23 @@ auto Presentation::resizeWindow() -> void {
...
-
-  if(settings.video.autoCentering) {
-    setGeometry(Alignment::Center, {viewportWidth, viewportHeight + statusHeight});
-  } else {
-    setSize({viewportWidth, viewportHeight + statusHeight});
-  }
-
-  setMinimumSize({256, 240 + statusHeight});
+  setSize({viewportWidth, viewportHeight + statusHeight});
+  setMinimumSize({16, 16 + statusHeight});
```
LukeUsher: 16x16 is a very small minimum window size, are you sure about that value?

LukeUsher (changes requested):

desktop-ui/presentation/presentation.cpp
FitzRoyX: Yeah I didn't know what to put here, maybe something like 160x144 is better since it's the lowest possible system resolution?

ares/ng/lspc/lspc.cpp
FitzRoyX: I'm pretty confused about this still. These are the only two combinations I could find that did not force the image into a non-integer scale

  screen = node->append<Node::Video::Screen>("Screen", 320, 256);
  screen->colors(1 << 17, {&LSPC::color, this});
  screen->setSize(320, 256);
  screen->setScale(1.0, 1.0);
  screen->setAspect(1.0, 1.0);
  screen->setViewport(0, 0, 320, 256);
mslug 2023-01-27 16-46-53

  screen = node->append<Node::Video::Screen>("Screen", 320, 256);
  screen->colors(1 << 17, {&LSPC::color, this});
  screen->setSize(304, 240);
  screen->setScale(1.0, 1.0);
  screen->setAspect(1.0, 1.0);
  screen->setViewport(8, 8, 304, 240);
mslug 2023-01-27 16-47-47

This is the 1.05ish scaling that the current code does that I was trying to avoid
gg1

FitzRoyX: Note that (like the Famicom) many games draw in the overscan area just fine and some even put things there that would otherwise look cut off. This is Blue's Journey in 320 vs 304:
bjourney 2023-01-27 17-54-17
bjourney 2023-01-27 17-55-08

desktop-ui/presentation/presentation.cpp
FitzRoyX: The only reason I can think of that this code exists is to prevent the user from resizing the menu away and the not figuring out that they can resize out of it? Because I think Windows itself safeguards the frame from being resized into oblivion. And I'd think any competent OS would.

(fwiw I don't personally like the resize ability, as it results in accidental resizes when trying to drag the title bar and you have to rescale twice to normalize the frame.) But most emulators allow it so I'm apparently in the minority here.

ares/ng/lspc/lspc.cpp
LukeUsher: In this case I believe we should crop the outer pixels based on the overscan setting, so setSize and setViewport would change based on the overscan flag.

LukeUsher (approved):

PR #963, opened 2023-02-06
n64: new accurate PIF/CIC HLE implementation

rasky:
Thanks to an ongoing team effort to fully reverse and document the PIF firmware in all its behaviors (https://github.com/GenericHeroGuy/pif-sm5-rom), a lot of new knowledge has surfaced around the exact behavior of PIF. The linked repo contains both a commented disassembly and a commented C decompilation. Many details are still being uncovered and analyzed, that will be necessary for a full PIF LLE implementation, but meanwhile there is enough material to make a new HLE implementation that is vastly more accurate than what Ares currently does.

In this PR, I rewrote the PIF implementation to strictly follow the behavior of the original firmware in many regards. Moreover, I also added a CIC HLE implementation, so that it is possible to accurately reflect the serial communication between PIF and CIC (which is now bit-accurate to the hardware).

What is emulated now:

 * PIF boot sequence is now very accurate. This is quite hard to test as much of the boot happens during IPL1/IPL2 so it is hardwired in the ROM, but the implementation does follow the original firmware in basically everything. 
 * PIF region lock is now emulated but disabled by default with a new flag in accuracy.hpp. It is possible to actually enable it so that an emulator compiled as NTSC (default) would start failing loading PAL games because the PIF region lock would trigger, leaving the user with a black screen like on real hardware (plus a message in console). By default, Ares still emulates a NTSC console in "universal" mode, that is letting PAL games boot.
 * PIF "IPL2 checksum" (that verifies the contents of IPL3) is now accurately emulated, causing a freeze on the CPU if it fails. The correct checksums expected for most known IPL3s are now hardwired in `CIC::power()`. This can be verified by altering a byte in the IPL3 of a game (in range 0x40 - 0x1000 in the ROM): in addition to the freeze, Ares now emits an error in console saying that the IPL2 checksum has failed. This can also be useful to development of new homebrew IPL3s.
 * PIF boot termination (or absence of) made by the game code is now accurately emulated. By making a ROM that fails to terminate (by writing the magic value to PIF-RAM), Ares now correctly freezes the game after a few seconds. This can be easily tested by patching libdragon.
 * PIF interface with SI is now 100% accurate, specifically in its interaction between reading and writing joyous command frames. This also fixes the FIXME in Ares where bit 0 of the PIF command byte weren't previously turned off as as a workaround. Now Ares correctly turns off bit 0 after SI DMA Write64B, and correctly emulates joybus transfers (poll controllers) on SI DMA Read64B even if the bit is not set.
 * Joybus command parsing is now 100% accurate to the firmware, including its split between parsing (at read time) and execution (at write time). This can have visible effects if a game changed PIF-RAM contents without setting bit 0 in the command byte.
 * CIC 6105 challenge protocol is now emulated by CIC, and the PIF transfers the challenge to CIC via the serial line and reads back the result. Also, the other CICs now accurately implement their "dumb" challenge protocol (which just complements the bytes of the challenge).

What is missing (left for followups):

 * 64DD is broken for now. I plan to fix it soon (keeping this in draft for now).
 * Reset emulation. Currently Ares does not support emulation of the Reset physical switch on N64. This is handled by PIF and CIC as well, including generating the pre-NMI interrupt on CPU. It is used by some games (eg. SM64) to provide a "fade out" effect when the reset button is pressed. 
 * CIC compare emulation. This is the protocol that is continuously run between CIC and PIF after boot. It is meant to detect tampering and also freeze the console if the communication is interrupted (eg: the cartridge is removed from the slot). I plan to add this to fully document the hardware behavior, but gate it behind an accuracy flag disabled by default as it is a gratuitous waste of CPU time in the case of Ares (unless we want to actually implement runtime cartridge removal in the future).
 * Timings. In the reverse engineering project, we have collected some evidence about timing of the PIF SM5K core, but this PR is not attempting to reproduce it in any way. The PIF thread is stepped by a "random" number.

Many thanks to @GenericHeroGuy, @invertego, and many other people that contributed to this effort.


rasky:
64DD support has been added to PIF/CIC HLE, so now the PR is ready for review and shouldn't cause any regression.

rasky:
I've also improved emulation of a few peripherals (Rumble Pak, Controller Pak) fixing a few games that had problems with them, and added emulation of RTC for Animal Forest and homebrew using it.

remutro:
Save states don't work with this branch in its current form. You need to add:   

s(cic);

To the serialize method in: ares/n64/system/serialization.cpp

PR #965, opened 2023-02-08
ws: rewrite RTC emulation

asiekierka:
* Support all S-3511 interrupt modes.
* Properly emulate S-3511 commands and the status register.
* Properly emulate the AM/PM hour mode. (I think)
* Restore the ability to skip RTC time during the emulator having been turned off.

Notes:

* The delays of communication with a real Seiko S-3511 are not currently emulated.
* The format of `.rtc` save files for WonderSwan emulation will be broken in the next release, at least as things currently stand.

Test ROM available [here](https://forums.nesdev.org/viewtopic.php?t=21513).

### DIFF
--- a/ares/ws/cartridge/cartridge.cpp
+++ b/ares/ws/cartridge/cartridge.cpp
@@ -48,8 +48,9 @@ auto Cartridge::connect() -> void {
   }
 
   if(auto fp = pak->read("time.rtc")) {
-    rtc.ram.allocate(fp->size());
+    rtc.ram.allocate(18);
     rtc.ram.load(fp);
+    rtc.load();
     has.rtc = true;
   }
 
@@ -82,6 +83,7 @@ auto Cartridge::save() -> void {
   }
 
   if(auto fp = pak->write("time.rtc")) {
+    rtc.save();
     rtc.ram.save(fp);
   }
 }
--- a/ares/ws/cartridge/rtc.cpp
+++ b/ares/ws/cartridge/rtc.cpp
@@ -3,10 +3,19 @@
 auto Cartridge::RTC::load() -> void {
   n64 timestamp = 0;
   for(auto n : range(8)) timestamp.byte(n) = ram.read(8 + n);
-  if(!timestamp) return;  //new save file
+  if(!timestamp || !(timestamp + 1)) return;  //new save file
+
+  if(status() & 0x15) {
+    // These status bits are always 0; reset state on invalid status.
+    initRegs(false);
+    return;
+  }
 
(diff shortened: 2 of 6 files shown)


ares/ws/cartridge/cartridge.cpp
```diff
@@ -48,8 +48,9 @@ auto Cartridge::connect() -> void {
   }
 
   if(auto fp = pak->read("time.rtc")) {
-    rtc.ram.allocate(fp->size());
+    rtc.ram.allocate(18);
```
LukeUsher, line 51: consider leaving this as fp->size, and having mia detect older .rtc files based on size, and re-initializing them to default settings.

ares/ws/cartridge/rtc.cpp
```diff
@@ -16,146 +25,245 @@ auto Cartridge::RTC::save() -> void {
 }
 
 auto Cartridge::RTC::tickSecond() -> void {
-  if(++second() < 60) return;
+  static auto bcdIncrement = [](n8& data) -> n8 {
```
LukeUsher, line 28: Do you think perhaps BCD handling could be added to nall?

I ask because as well as this bcdIncrement function, GB TAMA emulation has it's own toBCD/fromBCD functions, 
Nall::CD also contains BCD encode/decode functions used in all CD-ROM based systems.

Perhaps this is a good opportunity to standardise things with a BCD implementation in nall, and updating other places to use it rather than their own.

ares/ws/cartridge/cartridge.cpp
asiekierka: I honestly have no clue how to do that with Mia's system - there's no documentation that I'm aware of, and VSCodium/clangd not working too well doesn't help browse the codebase...

ares/ws/cartridge/rtc.cpp
asiekierka: I don't see why not.

asiekierka: I'm not sure whether this should be pursued as utility math functions or as a separate type, though; and if it's a type, I'm not sure if I can get the casting to work right (notice how the `second()`, `minute()`, ... functions operate) confidently enough, with my current C++ skillset.

rasky: We have a `BCD::decode` and `BCD::encode` in n64 as well. I think utility functions are fine, overtyping seems a recipe for more complex code with very little benefit.

LukeUsher: I'll pick this up after merging, dw about it for now

PR #971, opened 2023-02-13
ares: support disconnecting controller from emulated device

### LINKED ISSUE #969: [SNES] Can't disable Controller Port 2
**Describe the bug**

There are some games where you have to disable controller port 2 or you can't play further. A game like this is "Kessen! Dokapon Oukoku IV - Densetsu no Yuusha-tachi". It looks like it hangs after the first character selection screen. Right here:

image

But it just waits for Controller 2. The issue is that you can't disable P2 permanently, it gets re-activated after a restart. You can change the controller 2 to "nothing" after you boot a game, but this doesn't do anything.

**To Reproduce**
1. Start the game and and press "A" until you get to the position on the image.

**Additional context**
It does work with bsnes.

invertego:
Selecting "nothing" for a controller port from the UI only consistently blocked input from the host, and it did not reliably do the extra work required to actually disconnect the controller from the perspective of the emulated device.

This was handled only by the MD and N64 cores, but now all cores are standardized on the disconnection method originally implemented for MD.

Fixes #969. See #514 and #722 for the prior MD and N64 implementations.

### DIFF
--- a/ares/ng/card/slot.cpp
+++ b/ares/ng/card/slot.cpp
@@ -9,6 +9,7 @@ auto CardSlot::load(Node::Object parent) -> void {
   port->setType("Memory Card");
   port->setHotSwappable(true);
   port->setAllocate([&](auto name) { return allocate(name); });
+  port->setDisconnect([&] { device.reset(); });
   port->setSupported({"Memory Card"});
 }
 
--- a/ares/ps1/peripheral/port.cpp
+++ b/ares/ps1/peripheral/port.cpp
@@ -12,6 +12,7 @@ auto PeripheralPort::load(Node::Object parent) -> void {
   port->setType(type);
   port->setHotSwappable(true);
   port->setAllocate([&](auto name) { return allocate(name); });
+  port->setDisconnect([&] { device.reset(); });
   if(type == "Controller") {
     port->setSupported({"Digital Gamepad"});
   }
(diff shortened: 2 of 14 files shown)


ares/ng/card/slot.cpp
```diff
@@ -9,6 +9,7 @@ auto CardSlot::load(Node::Object parent) -> void {
   port->setType("Memory Card");
   port->setHotSwappable(true);
   port->setAllocate([&](auto name) { return allocate(name); });
+  port->setDisconnect([&] { device.reset(); });
```
invertego, line 12: This is not a controller, though it is a designated "HotSwappable" device, albeit one without any associated UI. I included this for completeness, but I could just as easily leave it out, because it currently has no effect.

ares/ps1/peripheral/port.cpp
```diff
@@ -12,6 +12,7 @@ auto PeripheralPort::load(Node::Object parent) -> void {
   port->setType(type);
   port->setHotSwappable(true);
   port->setAllocate([&](auto name) { return allocate(name); });
+  port->setDisconnect([&] { device.reset(); });
```
invertego, line 15: I did not draw a distinction here for controllers vs. memory cards, though I could have. As with the NG cards, there is currently no associated UI.

PR #994, opened 2023-02-25
MCD fixes to improve compatibility

TascoDLX:
1. Fixes write access to 8-bit registers mapped into the mcd gate array, including comm flags and wordram ctrl (on subcpu side). Also, fixes initialization of h-int vector & allows cdc status reads from maincpu.
2. Ignore maincpu access to wordram & prg-ram when it's disallowed, which avoids memory corruption in games with poorly coded sync. Similarly, the write protection feature was implemented on prg-ram access by the subcpu. All of this should allow MCD Dungeon Explorer to work correctly.
3. Implements io serial registers but not the transfer logic. Currently, data can be sent out but the tx buffer remains empty, allowing MCD The Animals! to boot. Also, io data registers were revised to handle signal latching correctly.

### DIFF
--- a/ares/md/controller/port.cpp
+++ b/ares/md/controller/port.cpp
@@ -31,10 +31,18 @@ auto ControllerPort::allocate(string name) -> Node::Peripheral {
 
 auto ControllerPort::power(bool reset) -> void {
   if(!reset) {
-    control = 0x00;
+    control        = 0x00;
+    dataLatch      = 0x7f;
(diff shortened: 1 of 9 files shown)


invertego:
cf1e8a7183e52e0d01ab8be6e9357b2c04e08544 should also fix Lunar - Eternal Blue (Japan). It clobbers an interrupt handler unless prg-ram protection is implemented.

Issue #982, opened 2023-02-18
[General] Compilation issue on Linux Mint 21 on latest.


HeroponRikiBestest:
**Describe the bug**
On Linux Mint 21, the emulator fails to compile, with `../nall/platform.hpp:22:10: fatal error: 'atomic' file not found` appearing on several files, and `/usr/include/glib-2.0/glib/glib-typeof.h:39:10: fatal error: 'type_traits' file not found` appearing on ares/ares/memory/fixed-allocator.cpp.

**To Reproduce**
Try and compile the program on Linux Mint 21.

**Expected behavior**
I expected the program to compile successfully.

**Screenshots**

**Additional context**
```
bestest@bestest-ThinkPad-T14-Gen-1:~$ gcc --version
gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
Compiling with clang also throws the same errors.
```
bestest@bestest-ThinkPad-T14-Gen-1:~/ares$ clang --version
Ubuntu clang version 14.0.0-1ubuntu1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```
I have not had compilation issues on any other program. Just recently, I compiled ScummVM without issue, but that goes for every other program I've tried. If there is any other program someone would want me to try compiling, please let me know.

remutro:
This can be resolved by installing the 'libstdc++-12-dev' package. This seems like an issue with the packaging/installation on this specific distro, so not sure if this is something ares should try to workaround. There is more information in the following:

[https://stackoverflow.com/questions/74543715/usr-bin-ld-cannot-find-lstdc-no-such-file-or-directory-on-running-flutte/](url)

invertego:
Our makefile now defaults to clang if you have it installed, so your clang install is borked, probably thanks to this:

https://bugs.launchpad.net/ubuntu/+source/clang/+bug/1997600

The output of `clang -v` would help diagnose, but it's likely clang is selecting something other than gcc-11 and you're missing the requisite libstdc++-N-dev package for that version, probably libstdc++-12-dev in this case.

HeroponRikiBestest:
Output of clang -v
```
[...]

Issue #932, opened 2023-01-05
Not compiling properly under MinGW64


WinFutureX:
Running any make command (MinGW64 native, commit 9eb14aea6fa67045204335d13af4a0a5d86570a0):
```
$ make clean
../nall/GNUmakefile:123: *** unknown arch, please specify manually..  Stop.
```

Running any make command with `arch=amd64` specified:
```
$ make clean arch=amd64 && make -j8 arch=amd64
Deleting obj ...
Deleting out ...
Compiling libco/libco.c ...
Compiling thirdparty/sljit/sljit_src/sljitLir.c ...
Compiling thirdparty/sljitAllocator.cpp ...
Compiling thirdparty/libchdr/src/libchdr_bitstream.c ...
make: *** [../nall/GNUmakefile:265: obj/libco.o] Error 127
make: *** Waiting for unfinished jobs....
make: *** [../nall/GNUmakefile:265: obj/sljitLir.o] Error 127
make: *** [../nall/GNUmakefile:265: obj/sljitAllocator.o] Error 127
make: *** [../nall/GNUmakefile:265: obj/libchdr_bitstream.o] Error 127
```

Previously, make used to run just fine.

remutro:
What [Windows] OS? Is MSYS2 up-to-date? I can't replicate this using the same commands specified against the latest master branch:

```
$ mingw32-make clean arch=amd64 && mingw32-make -j8 arch=amd64
mingw32-make[1]: Entering directory 'C:/Users/xxx/source/ares/desktop-ui'
Deleting obj ...
Deleting out ...
mingw32-make[1]: Leaving directory 'C:/Users/xxx/source/ares/desktop-ui'
mingw32-make[1]: Entering directory 'C:/Users/xxx/source/ares/desktop-ui'
Compiling libco/libco.c ...
Compiling thirdparty/sljit/sljit_src/sljitLir.c ...
Compiling thirdparty/sljitAllocator.cpp ...
Compiling thirdparty/libchdr/src/libchdr_bitstream.c ...
Compiling thirdparty/libchdr/src/libchdr_cdrom.c ...
Compiling thirdparty/libchdr/src/libchdr_chd.c ...
Compiling thirdparty/libchdr/src/libchdr_flac.c ...
Compiling thirdparty/libchdr/src/libchdr_huffman.c ...
```

And it continues just fine....  It isn't a 32-bit Windows installation is it (just a random thought)?

[...]

PR #1005, opened 2023-03-04
Updated build instructions

remutro:
Updated build instructions. Linux builds include information about additional audio & input drivers. Info about clang was added. Some other miscellaneous changes/improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,6 @@ Once complete, open a CLANG64 terminal window and proceed with building ares.
 When building with clang, by default symbols will be generated using an MSVC compatible format (CodeView) for use with Windows debugging tools. In order to generate GDB compatible symbols (Dwarf), specify the following option:  
 `symformat=dwarf`  
 
-###### Console Output  
-By default, the console is disabled on Windows builds. To enable it, specify the following option:  
-`console=true`
-
 Compilation
 -----------
 
--- a/README.md
+++ b/README.md
@@ -20,20 +20,47 @@ Only the latest nightly build is kept.
 Prerequisites
 -------------
 
-#### *nix building
+### *nix building
 
+###### Minimum required packages:
 ```
-g++ make pkg-config libgtk2.0-dev libcanberra-gtk-module libgl-dev libasound2-dev
+g++ make pkg-config libgtk2.0-dev libcanberra-gtk-module libgl-dev libasound2-dev libao-dev libopenal-dev
 ```
 
-By default, GTK2 is used, but support for GTK3 is available. You will need to install the additional package `libgtk-3-dev` as well
-as specifying the command line option `hiro=gtk3` at compile time.
+###### GTK2 & GTK3
+By default, GTK2 is used, but support for GTK3 is available. You will need to install the additional package `libgtk-3-dev` as well as specifying the command line option `hiro=gtk3` at compile time.
 
-#### Windows building
+###### SDL2 for input
+If you would like to use SDL for input, you will need to install the following the `libsdl2-dev` and `libsdl2-0.0` packages and perform a clean build of ares. You should then be able to select SDL for input in the Settings > Drivers menu.   
+
+##### Building with clang
+
+clang++ is now the preferred compiler for ares as it is known to produce a higher performing executable. If it is detected, the build will default to building with clang. It is recommended to install the `clang` package. If you would like to manually specify a compiler, you can use the following option: `compiler=[g++|clang++]`
+
+### Windows building
 
 To build on Windows, using MSYS2 is recommended which can be download [here](https://www.msys2.org/). Follow the instructions
-on this page to install and setup an appropriate MINGW64 environment. Running the command  `pacman -S --needed base-devel mingw-w64-x86_64-toolchain` from the MSYS2 MSYS terminal should setup everything you need to compile Ares. Note that in order 
-to compile, you will want to be in a MINGW64 terminal window after install and setup is complete. 
+on this page to install and setup an appropriate MINGW64 environment. Running the command:  
+```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -20,20 +20,47 @@ Only the latest nightly build is kept.
 Prerequisites
 -------------
 
-#### *nix building
+### *nix building
 
+###### Minimum required packages:
```
invertego, line 25: It looks like this includes recommended packages in addition to "minimal required".

README.md
```diff
@@ -20,20 +20,47 @@ Only the latest nightly build is kept.
...
+###### GTK2 & GTK3
+By default, GTK2 is used, but support for GTK3 is available. You will need to install the additional package `libgtk-3-dev` as well as specifying the command line option `hiro=gtk3` at compile time.
 
-#### Windows building
+###### SDL2 for input
+If you would like to use SDL for input, you will need to install the following the `libsdl2-dev` and `libsdl2-0.0` packages and perform a clean build of ares. You should then be able to select SDL for input in the Settings > Drivers menu.   
+
+##### Building with clang
+
+clang++ is now the preferred compiler for ares as it is known to produce a higher performing executable. If it is detected, the build will default to building with clang. It is recommended to install the `clang` package. If you would like to manually specify a compiler, you can use the following option: `compiler=[g++|clang++]`
```
invertego: We've learned (#982) that for certain Linux Mint and Ubuntu installs, just installing the `clang` package leads to an incomplete build environment. The remediation for that case was highly specific, so I'm not suggesting we repeat it here. I'm really not sure what would be better general advice.

invertego: The situation with msys2 is even worse; users absolutely should not install the package simply named `clang` (#932). Since this is in the *nix section and you call out specific advice for msys2 below, I think that's fine.

README.md
```diff
@@ -20,20 +20,47 @@ Only the latest nightly build is kept.
...
+
+### Windows building
 
 To build on Windows, using MSYS2 is recommended which can be download [here](https://www.msys2.org/). Follow the instructions
```
invertego, line 47: This is probably still the minimal footprint toolchain for building ares, but it would be nice if we had instructions on how to replicate the official build configuration. I can help with this.

README.md
```diff
@@ -20,20 +20,47 @@ Only the latest nightly build is kept.
...
-on this page to install and setup an appropriate MINGW64 environment. Running the command  `pacman -S --needed base-devel mingw-w64-x86_64-toolchain` from the MSYS2 MSYS terminal should setup everything you need to compile Ares. Note that in order 
-to compile, you will want to be in a MINGW64 terminal window after install and setup is complete. 
+on this page to install and setup an appropriate MINGW64 environment. Running the command:  
+```
+pacman -S --needed base-devel mingw-w64-x86_64-toolchain
+```  
+from the MSYS2 MSYS terminal should setup everything you need to compile ares. Note that in order to compile, you will want to be in a MINGW64 terminal window after install and setup is complete.
```
invertego, line 52: The msys2 installer now defaults to opening a UCRT64 terminal after install rather than MINGW64. Since they are pushing users in that direction we may want to follow.

README.md
```diff
@@ -47,11 +74,11 @@ git clone https://github.com/ares-emulator/ares.git
 From the root of the project directory run:
 
 ```
-make -j4 build=release
+make -j$((`nproc`-1)) build=release
```
invertego: I'm not so sure about adding an implicit dependency on an sh-like shell and Unix utilities like nproc for this example. Neither is required to build ares.

invertego: On an unrelated note, this example already isn't matching the build configuration used for official releases (`build=optimized`). I think it would be fine to just let the user pick up the default (currently `optimized`).

nall/GNUmakefile
```diff
@@ -149,7 +149,9 @@ ifeq ($(symbols),true)
   flags += -g
   ifeq ($(platform),windows)
     ifeq ($(findstring clang++,$(compiler)),clang++)
-      flags   += -gcodeview
+      ifneq ($(symformat),gdb)
```
invertego: You might want to call this `dwarf` since that's the actual name of the format and the explicit override `-gdwarf`. You might actually want to provide that override here; I think Clang will default to -gcodeview under certain conditions.

README.md
remutro: I lazily added the additional audio driver options. They are now added to a new section.

README.md
remutro: Discussed in discord, the additional package for Mint 21 is more of an exception then the rule. We'll leave it as is to avoid having to write detailed instructions for every distro. 

clang on Windows/MSYS is addressed in the Windows section.

README.md
remutro: Feel free to pass on instructions to me or submit an updated pull request if you have these available.

README.md
remutro: Leaving MINGW64 for now as UCRT64 is for Windows 10 and up and we have a good number of users on Windows 7 still.

README.md
remutro: Reverted to old verbiage for parallel processes. Removed specific build so it accepts default and added section to describe build options if anyone is interested.

nall/GNUmakefile
remutro: Updated to specifically use -gdwarf flag when 'symformat=dwarf` is specified, otherwise it defaults to CodeView format (current default).

README.md
invertego: clang builds also depend on UCRT. UCRT was backported all the way to Windows XP, available via a redist.

PR #1023, opened 2023-03-11
Fix save for Sonic 3 & Knuckles

remutro:
This should cover all 3 regions for the 3 No-Intro verified images.


LukeUsher:
Would it be better to check for a second header at the point s3 starts rather than hash the entire Rom? This would allow sram in romhacks to function where the hash doesn’t match an existing Rom.

remutro:
> Would it be better to check for a second header at the point s3 starts rather than hash the entire Rom? This would allow sram in romhacks to function where the hash doesn’t match an existing Rom.

This has been updated now so it isn't based on rom hash.

Issue #1026, opened 2023-03-12
[SFC] Tengai Makyou Zero (+ English Translation) - Crashes on v132


Zet-sensei:
**Describe the issue**
A user informed me earlier about a crash with the emulator when using the English translation of Tengai Makyou Zero.

After some tests, I can confirm about a regression bug on v132.
The original Japanese  versions & English translated version are affected.
The emulator crashes, but the same files work as intended on earlier versions of the emulator (v131 and prior).

**Rom Information**
Tengai Makyou Zero (Japan) - 8620203da71d32d017bb21f542864c1d90705b87eb67815d06b43f09120318a

Tengai Makyou Zero (Japan) (Shounen Jump no Shou) - 69d06a3f3a4f3ba769541fe94e92b42142e423e9f0924eab97865b2d826ec82d

Tengai Makyou Zero (English v7.0) - eb07eed483f8f79ab64309c50ad0ef19560946b8e28fbd695470552ba68e356e

remutro:
Only crashes when compiled with -O2 & -O3 flags (-Og & -O1 don't result in a crash). Likely an out of bounds error or some other bug leasing to a crash.

remutro:
Fixed with pull request: https://github.com/ares-emulator/ares/pull/1036

Issue can be closed.

PR #1028, opened 2023-03-13
Snes rtc crash fix

remutro:
Fixes issue: https://github.com/ares-emulator/ares/issues/1026

Credit goes to @invertego who isolated the issue first while investigating this issue. 

Note: This bug has existed in the code base since 2012 (11 years ago) and was only exposed when compiling with recent versions of clang with optimizations enabled (crash doesn't happen when building with gcc).


remutro:
Closed and resubmitted as it contained changes from other pull request.

PR #1040, opened 2023-03-18
build fixes for gcc (<= 11) and clang (linux)

invertego:
Note that `lld` is now a required package for building on Linux unless you are compiling with `g++`.


remutro:
We probably need to add to the Linux clang section of the README.md to include installing the ldd package now as well.

Issue #36, opened 2021-08-05
BIOS setup is tedious


FitzRoyX:
Problem:
-manually setting up bios filepaths for this many systems takes a long time
-has to be repeated if the cfg is wiped
-requires GUI code

Proposed behavior:
Include "Firmware" folder in release packages. On rom load, check if that system's cfg bios filepath is empty or invalid. If so, scan ares/firmware/ for that system's bios (by hash, not filename).

Users can drag bios files en masse into that folder.

LukeUsher:
I can see this being useful, but I wouldn't say it's a high priority because I don't think this will impact most users; the most likely use case is people setup each system as/when they first play a game for that system

I think the chances of people setting up every single bios up-front is one that will rarely occur in practice, except perhaps among some dedicated testers.

I do think we should look into this, though: simplifying configuration/setup is never a bad thing, as long as it doesn't impact functionality.

Shideravan:
I agreed about the hash verification, as it avoid that bad dumps or incompatible formats being used as the BIOS files (even the name is correct). It's surely something that can avoid emulation errors.

> Problem:
> -manually setting up bios filepaths for this many systems takes a long time
> -has to be repeated if the cfg is wiped
> -requires GUI code
> 
> Proposed behavior:
> Include "Firmware" folder in release packages. On rom load, check if that system's cfg bios filepath is empty or invalid. If so, scan ares/firmware/ for that system's bios (by hash, not filename).
> 
> Users can drag bios files en masse into that folder.

Shideravan:
If you're not liking it, why you can't remake it youself and we check if we like it?
When somebody try to improve something in the interface you just run to complain.

birdybro:
You'd be surprised how many users of multi-system emulators just get a pre-configured zip of all the bios files you need and dump it somewhere simple. It may not come up for single-system emulators, but multi-system emulators and frontends (MiSTer, retroarch, mednafen, etc...) all have plenty of these kinds of all in one bios zips out there. I think it's more popular nowadays to do the readymade zips of bios files than it is to individually go looking for specific ones, one at a time. 

Just my two cents coming from helping many people setup their MiSTer FPGA's, their Retroarch setups, and their Mednafen setups frequently... the vast majority of them prefer the script (in the case of the MiSTer there is an unofficial script) that grabs all of the BIOS files for them automatically, or downloading a zip with everything, instead of manually finding the right BIOS and placing it there.

dracc:
[...]

PR #1049, opened 2023-03-23
removed auto-generated qt.moc

jeffythedragonslayer:
this file says that it was created by the Qt Meta Object Compiler and auto-generated files should generally be .gitignore'd


LukeUsher:
Did you test building with “make hiro=qt4” or “hiro=qt5” to make sure this didn’t break the build when using qt as a driver? You may need to do this on a Linux machine though; im not sure we ever properly hooked up/tested Qt on windows.

jeffythedragonslayer:
Not yet, I'm not sure how that is different from the checks github runs.

Screwtapello:
For what it's worth, higan applied a very similar change very early on:

https://github.com/higan-emu/higan/commit/6673a1c0e9344dc22f87571519afd3fb6e7f27c3

It wasn't necessary to do anything else to keep Qt builds working; the only advantage of keeping the .moc committed is that you don't need the moc binary around to do a build. But if you have the Qt development libraries present to link against, it's extremely unlikely you wouldn't have the moc binary.

jeffythedragonslayer:
I actually can't build ares at all, even on master branch:

```lto1: fatal error: bytecode stream in file ‘obj/libco.o’ generated with LTO version 11.2 instead of the expected 11.3```

uname -a reports:

```Linux bsnes-ubuntu 5.19.0-35-generic #36~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Feb 17 15:17:25 UTC 2 x86_64 x86_64 x86_64 GNU/Linux```

Screwtapello:
It works for me on Debian Testing, once I installed `lld`. You might want to `git clean` to get rid of files generated by older compilers first.

PR #1051, opened 2023-03-24
sh2: mirror cache control register

### LINKED ISSUE #1037: [32X] Virtua Fighter Bugs
There's a number of glaring bugs.

1. The hit detection is busted. Especially evident with moves that pounce someone on the ground, they will simply not register at all.
2. The AI seems pretty broken. It will even sometimes choose to not move or attack at all.
3. One of Wolf Hawkfield's grabs has a completely broken animation that causes you to float in the air behind him (You should be held above him).

I'm using ares v132. I tested this against PicoDrive.

file: Virtua Fighter (Japan, USA) (En).32x
sha256: C595E136C91A012B1E7D591ACFB9D965926CAC51ACEA998B90969D74391BFA08

invertego:
Fixes #1037

### DIFF
--- a/ares/component/processor/sh2/sh7604/io.cpp
+++ b/ares/component/processor/sh2/sh7604/io.cpp
@@ -177,6 +177,7 @@ auto SH2::internalReadByte(u32 address, n8 data) -> n8 {
 
   //CCR: cache control register
   case 0xffff'fe92:
+  case 0xffff'fe93:
     data.bit(0)   = cache.enable;
     data.bit(1)   = cache.disableCode;
     data.bit(2)   = cache.disableData;
@@ -719,6 +720,7 @@ auto SH2::internalWriteByte(u32 address, n8 data) -> void {
 
   //CCR: cache control register
   case 0xffff'fe92:
+  case 0xffff'fe93:
     cache.enable      = data.bit(0);
     cache.disableCode = data.bit(1);
     cache.disableData = data.bit(2);


LukeUsher:
I feel like it's worth mentioning that for most SH2 internal registers, MAME seems to map them to both addr and addr + 1

They didn't implement +1 for cache control, but their SH2 core does not actually emulate cache; so it doesn't matter there.

But this does lead credit to the theory that a lot of these registers have copies.

eg:

```
	map(0xfffffe80, 0xfffffe81).rw(FUNC(sh2_device::wtcnt_r), FUNC(sh2_device::wtcnt_w));
	map(0xfffffe82, 0xfffffe83).rw(FUNC(sh2_device::rstcsr_r), FUNC(sh2_device::rstcsr_w));
```

invertego:
Those are not examples of MAME mirroring 8-bit registers. Those are examples of MAME implementing 16-bit registers.

Still, you are not wrong to suggest that there may be other mirrors.

LukeUsher:
> Those are not examples of MAME mirroring 8-bit registers. Those are examples of MAME implementing 16-bit registers.
> 
> Still, you are not wrong to suggest that there may be other mirrors.

My bad; pro-tip: don’t look at code when sleep deprived :’)

invertego:
Marking ready, though at some point I would still like to get some HW verification on this (I've written a test ROM for this express purpose) and if the results demonstrate the need for further changes I'll make a follow-up PR.

LukeUsher:
We can reliably say that this mirror exists at this point in time, we just can't explain why/what other mirrors may exist, hardware testing will help that

PR #1070, opened 2023-04-06
YM2612 accuracy fixes

### LINKED ISSUE #858: [MegaDrive] Sonic 3 & Knuckles: Oddly Pitched Audio
Tried playing sonic 3 on ares and noticed some odd pitches and audio in the sfx and music, most notably the roll sound (roll sound issue also applies to sonic 1 and 2)
(could potentially be due to model differences in what ares emulates vs what most emulators/official releases base their audio off of, I'm not aware of the differences between each model, and whether that could be the reason why the audio sounds off to me)

Using the v130.1 release of Ares (though i tested on latest commit and noticed the same issue) and the latest nightly build of blastem for the comparison, blastem chosen simply due to being on hand and sounding "right" compared to what im used to.

Roll Sound: Blastem (right) first, Ares (left) second.

https://user-images.githubusercontent.com/102590697/197631008-ec0f705b-1ecd-451e-b099-4ba542a08687.mp4

Ares sound much higher pitched and more like a whine.

Knuckles breaking out of the ground: Ares (left) first, Blastem (right) second.

https://user-images.githubusercontent.com/102590697/197631612-b01662cf-f895-4f86-9f6f-10f1aa7215ad.mp4

Ares sounds slightly muffled.

Beginning of Special Stage theme: Blastem first video, Ares second.

https://user-images.githubusercontent.com/102590697/197632094-fc2a4ce6-a2f9-44ff-8299-a75ee1998c2f.mp4

https://user-images.githubusercontent.com/102590697/197634120-239d0a62-08b0-4d6b-9bad-bbe835272682.mp4

Ares sounds slightly twangier? It's like a silent part of the background is breaking out? Hard to describe.

End of Act Jingle: Blastem first video, Ares second.

https://user-images.githubusercontent.com/102590697/197634848-77a0e5d7-d7ad-4b72-975b-ab5b691071b0.mp4

https://user-images.githubusercontent.com/102590697/197634861-e5ec64d1-0805-41be-8740-578f39ac081e.mp4

Ares starts slightly too high pitched

**Rom Information**
Sonic_Knuckles_wSonic3.bin
image
Rom sourced from the steam release of SEGA Mega Drive & Genesis Classics (which means Japanese Sonic 3 stitched onto International Sonic and Knuckles)

TascoDLX:
1. Operator modulation is affected by pipeline delays, which is known thanks to RE work by Sauraen et al. Fixes #858.
2. Adding crossover distortion for authenticity. This is often referred to as "ladder effect", which is a different thing but... these elements interact to produce a distinct quality of noise at very low output volumes (and even when output is suppressed!). Note, this distortion was "fixed" in the YM3438 (used in many MD model 2 revisions) which is basically the only functional change within the chip (external amp circuit was also changed).

### DIFF
--- a/ares/component/audio/ym2612/channel.cpp
+++ b/ares/component/audio/ym2612/channel.cpp
@@ -156,6 +156,7 @@ auto YM2612::Channel::power() -> void {
     op.outputLevel = 0x1fff;
     op.output = 0;
     op.prior = 0;
+    op.priorBuffer = 0;
 
     op.pitch.value = 0;
     op.pitch.reload = 0;
--- a/ares/component/audio/ym2612/serialization.cpp
+++ b/ares/component/audio/ym2612/serialization.cpp
@@ -50,6 +50,7 @@ auto YM2612::Channel::Operator::serialize(serializer& s) -> void {
   s(outputLevel);
   s(output);
   s(prior);
+  s(priorBuffer);
 
   s(pitch.value);
   s(pitch.reload);
--- a/ares/component/audio/ym2612/ym2612.cpp
+++ b/ares/component/audio/ym2612/ym2612.cpp
@@ -30,12 +30,16 @@ auto YM2612::clock() -> array<i16[2]> {
       return y < 0x1a00 ? pow2[y & 0x1ff] << 2 >> (y >> 9) : 0; // -78 dB floor
     };
 
-    s32 feedback = modMask & op[0].output + op[0].prior >> 9 - channel.feedback;
+    s32 feedback = modMask & op[0].prior + op[0].priorBuffer >> 9 - channel.feedback;
     s32 accumulator = 0;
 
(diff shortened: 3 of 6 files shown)


LukeUsher:
Although not strictly hardware correct, the ym2612 emulation core is currently used to emulate the FM portion of the Neo Geo's YM2610; so the neo geo serializer version would need to be updated also

TascoDLX:
Oh, I see. I would assume that the crossover distortion is not present in the ym2610, in which case maybe it should be implemented as an option instead?

LukeUsher:
> Oh, I see. I would assume that the crossover distortion is not present in the ym2610, in which case maybe it should be implemented as an option instead?

I wouldn't worry about anything other than just incrementing the neo geo serializer version for now; the ym2610 has less FM channels than the YM2612 anyway, so the hack we're doing isn't exactly accurate

PR #1085, opened 2023-04-19
n64: Adjust initial stick range and introduce response curve

kev4cards:
This PR addresses the undercutting range used for the control stick currently. Assuming the current parameters set for the N64 octagonal gate and the axial dead-zone, the initial preprocessed range of 85 needs to increase to 102 in order to fully cover the diagonal value of 69 prior to the octagon clamping code. Otherwise, the controller will only reach 60 then jump to 69 after octagon clamping.

Additionally, this PR replaces the normalized linear scaling curve with a configurable normalized sinusoidal response curve. A sinusoidal curve is necessary to keep both the diagonal and cardinal maximum covered. Otherwise, a piecewise power function will be needed. The default values will simplify the equation back to a normalized linear curve. The general non-normalized scaling function is as follows:

`lengthAbsoluteX = pow(((lengthAbsoluteX - innerDeadzone)/(preprocessedCardinalMax - innerDeadzone)) , (responsePower1 / responsePower2)) * preprocessedCardinalMax * pow(((1.0 - cos(((lengthAbsoluteX - innerDeadzone)/(preprocessedCardinalMax - innerDeadzone)) * Math::Pi)) / 2.0) , ((responsePower2 - responsePower1) / responsePower3))`

The three responsePower variables will have the largest impact on scaling. Setting responsePower1 = responsePower2 will cause linearity. When responsePower3 is less than responsePower2 - responsePower1, the initial rise past the deadzone will slow dramatically.

Furthermore, a proportional factor can be added to the above equation at the expense of changing the range.

Lastly, none of this is set in stone. Appropriate bounds need to be figured for the above variables that are not dependent upon another variable, for example.

### DIFF
--- a/ares/n64/controller/gamepad/gamepad.cpp
+++ b/ares/n64/controller/gamepad/gamepad.cpp
@@ -195,44 +195,57 @@ auto Gamepad::read() -> n32 {
   platform->input(z);
   platform->input(start);
 
-  //scale {-32768 ... +32767} to {-85 ... +85}
-  auto ax = x->value() * 85.0 / 32767.0;
-  auto ay = y->value() * 85.0 / 32767.0;
+  auto cardinalMaxN64          =  85.0;
+  auto diagonalMaxN64          =  69.0;
+  auto preprocessedCardinalMax = 102.0; //TODO: make this a function of diagonalMaxN64 and innerDeadzone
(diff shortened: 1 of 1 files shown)


kev4cards:
I am going to close this in favor of creating two pull requests. One to address the undercut range and another in the future for customization.

Issue #1086, opened 2023-04-20
[PC Engine CD] Akumajou Dracula X: Chi no Rondo - Blue characters and bugged cutscenes


ObotPrime:
I played through Chi no Rondo on Ares's PC Engine CD core as thoroughly as I could, testing every cutscene, level, and boss. The game can be finished to completion, but there are a few minor problems:
- Green flame in the top left corner of the menu screen (only appeared before starting the game, seemed to go away after clearing a few stages)
- Player characters will use their underwater palette (meaning they turn a solid shade of blue) in certain sections of stages:
	- Both water cave areas in Stage 2 (see attached image)
	- Inside the ship in Stage 5
	- Entering the third area of Stage 2' (before the stairs leading back onto the bridge)
	- Second area of Stage 4' (if the palette gets corrected after taking a hit, entering other secions will set it back to blue)
	- Second and fourth areas of Stage 5'
- The following cutscenes have the audio cutting out after a few seconds with the rest of the scene playing in silence:
	- Richter rescuing Maria
	- Both Tera rescue scenes
	- Both Iris rescue scenes
	- Both Anette rescue scenes

Pixel Accurate mode didn't seem to affect any of these.

Rom information:
http://redump.org/disc/29878/

```
File: Akumajou Dracula X - Chi no Rondo (Japan) (FABT, FACT).cue
SHA256: bac448a11c78a03408d72581661f288302f3ab20245eaca06baa0d6e45911a02
```
![BlueRichter](https://user-images.githubusercontent.com/49524053/233414371-2057df0d-df39-4e93-bde0-98abe7b9349d.png)

Issue #443, opened 2022-03-09
[PC ENGINE CD] Asuka 120% Maxima - Burning Fest. Maxima (Japan):  Various visual issues


HeroponRikiBestest:
Emulator : Ares v127 Built 2022-03-09
Options : Default
Game : Asuka 120% Maxima - Burning Fest. Maxima (Japan)
System : PC Engine CD
Bios: Version 3.00

![image](https://user-images.githubusercontent.com/50224630/157419710-654b26ac-d1b1-461e-964e-41d80f1c4f8b.png)
![image](https://user-images.githubusercontent.com/50224630/157419795-99484428-8a82-4a92-9561-0986f7513713.png)
![image](https://user-images.githubusercontent.com/50224630/157419906-c91d1fe3-543b-4f8e-808b-4ce72e368910.png)
![image](https://user-images.githubusercontent.com/50224630/157419964-1905fd0a-ca51-4681-bd9d-c2b47f6067d0.png)

I won't go too into describing all of the issues present here, there are a lot of them, they vary, and the issues change somewhat often with updates to ares; this is possibly the most sensitive PC Engine CD game I've seen. I recommend just comparing against this video https://youtu.be/LUO4FOZr4WA or any similar ones.

HeroponRikiBestest:
I forgot to test on pixel accuracy mode. Here are a few shots of what it looks like with that on.
![image](https://user-images.githubusercontent.com/50224630/159146571-b6fb25b6-12f3-4ce5-9455-e30b420d4932.png)
![image](https://user-images.githubusercontent.com/50224630/159146596-3b46cae9-b12d-473b-a0ba-1a1fa140995f.png)
![image](https://user-images.githubusercontent.com/50224630/159146607-4dcd9777-e09e-48d4-9dc6-5d4989a7e6ee.png)
![image](https://user-images.githubusercontent.com/50224630/159146614-53ad1ccd-0473-4dc4-bf14-0ab6584d96f0.png)

HeroponRikiBestest:
[...]

PR #1090, opened 2023-04-24
pce: fix off-by-one error for scanline interrupts

invertego:
This change also removes a timing hack from the pixel accurate renderer. I believe this hack may have been an attempt to work around the bug addressed by this PR, though I can't say for sure. The commit that introduced it (a57cc9f) only mentions Air Zonk, but that title currently seems fine with or without the hack. I tried some other titles and discovered that it actually causes garbage on the title screen of Victory Run, so I'm opting to remove it.

Fixes the "blue characters" part of #1086

### DIFF
--- a/ares/pce/vdp-performance/vdc.cpp
+++ b/ares/pce/vdp-performance/vdc.cpp
@@ -62,10 +62,12 @@ auto VDC::vclock() -> void {
       timing.voffset = 0;
     } break;
   case VDS:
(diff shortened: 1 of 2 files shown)


HeroponRikiBestest:
This pull both fixes and creates an issue with https://github.com/ares-emulator/ares/issues/443#issuecomment-1523178587 ; this may not be any issue with the pull, as the resulting new issue is far less bad than the old one, and could be due to other issues that simply weren't visible before now.

Issue #1111, opened 2023-05-13
Add support for other patch formats than .bps


amayra:
this feature that some emulators offer. It can automatically patch the game if the patch and the ROM have the same filename
![like this](https://fantasyanime.com/images/tutorials/patching_ex.gif)

LukeUsher:
This feature has existed for a very long time, but it only works with the .bps format, other formats are not implemented and are not likely to be any time soon.

amayra:
most old patch are in ips format 

and patch dev is long missing or even dead to care about update it 

every other emulator support this too

aquova:
I'd be interested in making a pull request that adds IPS support in a similar vein to how BPS is supported now, if there's interest in having it.

LukeUsher:
Should we close this issue or is there another widespread format we should support?

aquova:
Personally, I think this could be closed. From what I've seen, IPS + BPS covers the vast majority of existing patches, especially for the sizes of games that Ares supports. More can be looked at on a case-by-case basis, but it does open a can of worms.

LukeUsher:
agreed , ips and bps is plenty

PR #1116, opened 2023-05-21
Add support for IPS patching

aquova:
Adds support for patching ROMs on the fly with a corresponding IPS patch file. Functions in a similar manner to the pre-existing BPS support. If a *.ips file is found with the same name as the ROM, it attempts to apply the patch. If both a BPS and IPS file are found, the BPS patch takes priority. This patch implements the IPS format as described here: https://zerosoft.zophar.net/ips.php.


aquova:
Feature was submitted in response to issue #1111

PR #1117, opened 2023-05-21
nall: Linux does support file creation time

jeffythedragonslayer:
I saw Screwtape mention this as part of a good first issue somewhere, thanks for pointing that out about statx

### DIFF
--- a/nall/inode.hpp
+++ b/nall/inode.hpp
@@ -95,7 +95,7 @@ struct inode {
     //the best that can be done in this case is to return st_mtime if it's older
     case time::create: return min((u32)data.st_birthtime, (u32)data.st_mtime);
     #else
-    //Linux simply doesn't support file creation time at all
+    //Linux doesn't support file creation time through stat(), but does through statx()
     //this is also our fallback case for unsupported operating systems
     case time::create: return data.st_mtime;
     #endif


Screwtapello:
This PR has a change to the comment, but doesn't actually try to *call* `statx()`. Was that deliberate, or did you forget to push a commit?

jeffythedragonslayer:
Yes it was deliberate, I haven't delved deep into the codebase yet to actually fix the call.  Just trying to stop people who are browsing the codebase from getting wrong information into their heads for now.

LukeUsher:
Please consider adding an implementation that uses statx; it seems odd to merge a commit that makes reference to another possibility without providing it.

flagrama:
This also does depend on what you want the minimum required Linux kernel version for the emulator to be. `statx` was introduced in the 4.11 Linux kernel, which is now over 6 years old so is probably needs suiting for most people, but may effectively increase the minimum kernel version, and the implementation of this might need to keep that in mind.

In fact mentioning a minimum Linux kernel version/maximum Linux kernel age the emulator wants to support might be nice to have in the README for situations like this.

LukeUsher:
There has been no movement on this for some time; closing for now; if someone does indeed want to implement statx, please make another PR doing so.

PR #1131, opened 2023-06-21
fix FTFBS with GCC-13

a-detiste:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1037580


Screwtapello:
This should have been fixed last month by https://github.com/ares-emulator/ares/commit/6a7898396a14eef257e63cd002fb26ffbf6e2581, although this is a more minimal patch (it only affects `nall/arithmetic/natural.hpp`, rather than all of `nall/arithmetic`).

a-detiste:
Ha yes ! Didn't noticed

PR #1141, opened 2023-06-29
Update GTK info and app terminal options in README

remutro:
Updated to show GTK3 is the default now, add missing terminal options.


LukeUsher:
I knew I'd done this, turns out I missed the file from my commit, oops! Will merge yours anyway

Issue #1133, opened 2023-06-22
[SFC] Some 65x pages are called "banks"


jeffythedragonslayer:
I think that `echo.bank` should be renamed to `echo.page` - ditto for `brr.bank`

A page in 65x is 256 bytes, and a bank is 256 pages.  See fullsnes on DIR and ESA - these registers specificy addresses in 256-byte steps.  It's harder to figure what these registers are the way the terminology is being (mis)used.

PR #1142, opened 2023-06-30
Bank2page brr

jeffythedragonslayer:
can someone please test this?  I can't.

for #1133


Screwtapello:
Well, it builds!

jeffythedragonslayer:
It's probably correct then, I would've gotten a compiler error if not enough things were renamed.

PR #1144, opened 2023-07-08
Don't silently fail loading a rom if it is too small

remutro:
It's annoying when trying to load a rom and mia determines the game is under a minimum length but doesn't alert you as to why it refuses to load. This will print a debug message when these scenarios are encountered so the reason has visibility. 

I also added the serial number to the 32X manifest, which was previously missing (we show it for the mega-drive already).


Screwtapello:
Just a style nit: Near used nall's `print()` rather than the C standard `printf()` for debug logging, since it's type aware and doesn't need the format codes and data types to be kept in sync:

    print("[mia] Failed to load ", rom.size(), "-byte ROM, expected at least 32768 (0x8000) bytes");

remutro:
> Just a style nit: Near used nall's `print()` rather than the C standard `printf()` for debug logging, since it's type aware and doesn't need the format codes and data types to be kept in sync:
> 
> ```
> print("[mia] Failed to load ", rom.size(), "-byte ROM, expected at least 32768 (0x8000) bytes");
> ```

You are [of course] correct. I forgot this existed in nall. I've updated all references appropriately (and tweaked the messaging with a little more consistency).

LukeUsher:
This will do for now, but in time perhaps we can extend mia with a getErrorMessage type function, such that the GUI can handle it and display friendly messages.

PR #1145, opened 2023-07-14
Add "Weave" deinterlacing support to N64 emulator

SpookyIluha:
When an interlaced output is made by the game, an N64 emulator right now only allows "Bob" deinterlacing, this PR also adds a "Weave" deinterlace option:
- when a Weave Deinterlacing option is selected in the configuration panel, parallel-rdp and vulkan are configured to produce such output
- weave deinterlacing internally is only enabled and used when VI's interlacing is enabled and parallel-rdp's supersampling is disabled, otherwise "Bob" deinterlacing mode is used
- internally the output image is also modified with even/odd lines swapped to fix graphical bug
- weave deinterlacing mode is experimental since it hasn't been tested with all games/homebrew, but it shouldn't cause any issues, especially with it being disabled


LukeUsher:
Not sure why a merge commit ended up in here, any chance you can fix that?

SpookyIluha:
> Not sure why a merge commit ended up in here, any chance you can fix that?

fixed, now it should just be this one commit

PR #1150, opened 2023-07-20
New Core: Nichibutsu - My Vision

Kalixtan:
This adds a New core for the [Nichibutsu - My Vision](https://en.wikipedia.org/wiki/My_Vision), made this over a few days as a first project in Ares as its quite simple hardware.

Working:
- TMS9918 video.
- AY-3-8910 sound.
- All 14 button inputs.
- Supports all dumped games with no known bugs.
- Supports the raw MYV (MyVision) roms.

Todo/Verify:
- Verify all buttons are fully working (seems to be perfect but due to lack of test program and not all games use all buttons etc its hard to say definitively).
- Add Mame rom support, I've added a "MyVision.bml" that has all info for all the known dumps based on info from Mame, I could not work out a way to add this meanwhile supporting the raw rom type.

image
image
image
image
image

### DIFF
--- a/desktop-ui/game-browser/game-browser.cpp
+++ b/desktop-ui/game-browser/game-browser.cpp
@@ -23,6 +23,12 @@ auto GameBrowserWindow::show(shared_pointer<Emulator> emulator) -> void {
   games.reset();
 
   auto tmp = (shared_pointer<mia::Medium>)mia::Medium::create(emulator->medium);
+  if(!tmp) {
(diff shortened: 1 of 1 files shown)


LukeUsher:
This commit was rebased and merged as c94c7a9b85d3108605738abaf024b71fc624db1d; the original PR was damaged by permissions issues when attempting to collapse into a single commit.

Issue #1154, opened 2023-07-21
N64 Performance slowly degrades on Apple Silicon as CPU, memory use steadily rise


jcm93:
**Describe the bug**
Hi, this seems to be an odd issue. On my M1 Max machine, loading ares v133 has great performance at the outset. After several minutes of gameplay, however, the framerate in the lower right begins to slow down and sound crackle starts to happen.

This behavior seems to happen regardless of what is going on in-game; that is, performance degradation seems to be purely correlated with uptime and not necessarily related to any particular ingame activity.

Memory use seems to slowly climb as well, from e.g. 610 to 624MB, for Majora's Mask, or from 760 MB to 780 MB in Star Wars Episode I: Racer, over the course of 2-3 minutes, while paused, so it certainly seems possible that there is a memory leak underlying this issue.

**To Reproduce**
Load into any game. Games tested include Majora's Mask and Star Wars Episode I: Racer. Stay ingame for several minutes and observe slowdowns.

Please let me know if there is any other information or debugging I can perform to help diagnose this issue! Ares looks super promising and I look forward to continuing to use it.

Edit:

<img width="695" alt="Screenshot 2023-07-21 at 2 42 59 PM" src="https://github.com/ares-emulator/ares/assets/6864788/fc56970d-4e97-481c-9ffe-50cb0b605fff">

Additionally, the number of mach ports used by ares seems to rise at a very rapid clip. I didn't notice this at first, but this is after the ~10 minutes spent writing this issue report. When ares launches, only around 1000 ports are open.

Azonnali:
I also observe this problem on a M1 Mac Mini. Was on V132 and V133. 
See discord support thread for a video of this.

jcm93:
Looking at this a bit more, I compiled myself with homebrew moltenVK and the leak/slowdowns appear to be gone entirely. Could be an issue with the version shipped by ares, will dig a bit more and report back.

LukeUsher:
should be fixed in 3de3f99

PR #1155, opened 2023-07-22
Update MoltenVK dependency to version 1.2.4

jcm93:
This PR updates the MoltenVK dependency to version 1.2.4 to resolve a memory leak and performance issue with the N64 core that appears when using MoltenVK 1.2.0 through 1.2.3.

Further context: https://github.com/ares-emulator/ares/issues/1154

I have tested MoltenVK 1.2.4 with Star Wars Episode I: Racer, and Majora's Mask, verifying that the leak issue is gone. While it is possible some compatibility issues may arise in the future with this version of MVK, resolving the leak seems to be fairly high priority.


hackgrid:
I guess MoltenVK could always be updated to the latest version for any new release @LukeUsher ?
Is there any reason not to do it, besides having to test if something broke? :D

Or maybe for all thirdparty stuff?

LukeUsher:
> I guess MoltenVK could always be updated to the latest version for any new release @LukeUsher ? Is there any reason not to do it, besides having to test if something broke? :D
> 
> Or maybe for all thirdparty stuff?

We usually do, but for this release it happens a couple of months back due to the long development time

PR #1156, opened 2023-07-23
mia: fix a few n64 serials

invertego:
- Beetle Adventure Racing (J): Replace "NBB" with "NB8" (fix substitution).
- J.League Live 64: Replace "NLJ" with "NJL" (fix transposition).
- Quake 64: Replace "NKQ" with "NQK" (fix transposition).
- Star Fox 64: Remove "NFP" variant. This was printed on the box/manual in some regions but only "NFX" appears in the ROM.


invertego:
FYI @remutro

remutro:
LGTM

PR #1157, opened 2023-07-23
n64: implement rsp vu pipeline timings

invertego:
Implement vector unit (VU) timing including dual issue with the SU. Tested against real hardware with a random instruction stream generator.

Known limitations:
- Invalid/missing instruction timings not implemented.
- CFC2/CTC2 dual issue restrictions not implemented.
- Some corner-case dual issue restrictions (e.g. LTV + VNOP) not implemented. On real hardware, VU instructions with a DE operand are sometimes prevented from dual issuing with vector loads under seemingly erroneous conditions.


LukeUsher:
What do you think of opening an issue to track the remaining limitations? Even if the issue just states that more research is required ?

invertego:
That's fine. I don't expect any future changes to substantially impact timing as a whole (barring undiscovered bugs).

PR #1160, opened 2023-07-23
Revert "ws: fix off-by-one in line IRQ handling"

asiekierka:
Reverts ares-emulator/ares#1158

The fix provides the correct visual results, but for the wrong reasons. After FluBBa's review, it seems a much deeper analysis of the PPU behaviour is required.


LukeUsher:
Looks like I was a little too quick to merge this time :P

PR #1169, opened 2023-07-30
n64: cpu timing improvements

invertego:
This PR was motivated by the observation that ares was running the main CPU much, much slower than real hardware. Unfortunately, this was covering for timing inaccuracies in other places. With this corrected, many hardware inaccurate slowdowns have disappeared, but in some cases ares is now visibly faster than real hardware. These will be addressed, in time, with further improvements to ares' timing accuracy.

I've been building a framework for collecting instruction traces and playing them back on real hardware for timing measurement. Using a trace from SM64 with all loads/stores removed and with all instructions fully loaded into the icache, ares went from 58% of real hardware speed to 97% after these changes. Note that these figures are not indicative the overall speed of the emulated system, but they highlight the severity of the CPU slowdown ares was experiencing in the most extreme cases.

Given that the emulated CPU is now running faster than before, there may be increased demand on the host CPU. However, there should be no impact for games that A) are not fully utilizing the CPU and B) have a trivial idle loop that can be detected by ares. To further minimize the impact, I have expanded the detection of idle loops to include jumps in addition to branch instructions.

### DIFF
--- a/ares/n64/cpu/cpu.hpp
+++ b/ares/n64/cpu/cpu.hpp
@@ -115,7 +115,7 @@ struct CPU : Thread {
         line.valid = 1;
         line.tag   = address & ~0x0000'0fff;
       } else {
-        self.step(2 * 2);
+        self.step(1 * 2);
       }
     }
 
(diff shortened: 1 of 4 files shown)


LukeUsher:
Let's get this merged; there should be enough time between this and the next release to address any fallout caused by exposing other timing isuses elsewhere

PR #1171, opened 2023-08-01
n64: clamp to octagon edge instead of scale

kev4cards:
This PR addresses an issue where the input was always being scaled by the edge of the octagon even when within the bounds of the octagon. This caused input to not be scaled as linearly as it should have been. Therefore, an if statement has been added to check if the input's current length is beyond the edge of the octagon. If so, a clamp is now applied with the values of edgex and edgey. The scale never considered the magnitude of the input, only direction. Thus, a snap back effect could occur with that method.

### DIFF
--- a/ares/n64/controller/gamepad/gamepad.cpp
+++ b/ares/n64/controller/gamepad/gamepad.cpp
@@ -234,9 +234,12 @@ auto Gamepad::read() -> n32 {
     auto edgey = copysign(min(abs(edgex * slope), cardinalMax / (1.0 / abs(slope) + (cardinalMax - diagonalMax) / diagonalMax)), ay);
     edgex = edgey / slope;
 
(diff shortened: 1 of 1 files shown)


invertego:
The snapping issue was not present in the last version of this code by Near (with the octagonal logic contributed by MerryMage):
https://github.com/ares-emulator/ares/blob/c5583ade47ec899518ec117b084d8b2c0127624b/ares/n64/controller/gamepad/gamepad.cpp

I believe the original code was performing as designed. It made it so the physical limit of a circular gate would coincide with the limit of the virtual octagonal gate. By clamping, this PR removes any tactile feedback or awareness of when you have approached (or exceeded) the limit of the input range. It's now as if there is an invisible octagonal gate that you just can't feel.

That said, there's at least one significant point in favor of clamping: it should work better than scaling with controllers that already have an octagonal gate (e.g. original N64 controllers or replicas).

PR #1178, opened 2023-08-05
Add support for system-wide installation of Ares on Linux

### LINKED ISSUE #1173: Upstream Aur/Flatpak patches
**Describe the bug**

Both the aur package and the Flatpak have issues with not picking up the shader and database folder by default.

The aur package circumvents this issue with a patch: https://aur.archlinux.org/cgit/aur.git/tree/ares-paths.patch?h=ares-emu

I've ported over that patch to the Flatpak here: https://github.com/flathub/dev.ares.ares/pull/29 and it also seems to fix the Flatpak. 

The maintainer of the Flatpak commented on the patch and suggested a way ares could incorporate the patch: https://github.com/flathub/dev.ares.ares/pull/29#issuecomment-1662363481

I am not the most experienced with coding so I am a little out of my league here. Could the patch somehow be upstreamed and incorporated into ares?

Mastergatto:
This PR, as the title says, adds support for system-wide installation of Ares on Linux. Basically it will attempt to locate the shaders and the databases also in the system folders (e.g. in /usr/share/ares), not just in the home and, if portable, in its application directory. 

This allows Ares, as-it-is, to load properly shaders and databases, thus avoiding additional maintenance burdens for packagers (Flatpak, AUR, and also distro packages).

I haven't yet seen issues with these changes, but I have received positive feedback so that's good enough for me.

The only caveat is that this PR is still untested on BSD, although I don't think it will cause issues on that platform too. But if someone is kind enough to test this on BSD for us...;)

It fixes #1173 .

### DIFF
--- a/desktop-ui/desktop-ui.cpp
+++ b/desktop-ui/desktop-ui.cpp
@@ -26,10 +26,15 @@ auto locate(const string& name) -> string {
   location = {Path::userData(), "ares/", name};
   if(inode::exists(location)) return location;
 
-  // On non-windows platforms, after exhausting other options,
-  // default to userData, on Windows, default to program dir
+  // On non-windows platforms, this time check the shared
+  // data directory, on Windows, default to program dir,
   // this ensures Portable mode is the default on Windows platforms.
   #if !defined(PLATFORM_WINDOWS)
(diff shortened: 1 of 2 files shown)


LukeUsher:
Looks great, thanks a lot!

PR #1186, opened 2023-08-09
recompiler: allocate code cache early in main

invertego:
On Windows and Linux, ares uses a buffer in the bss section for the recompiler code cache. On macOS this buffer must be dynamically allocated for security reasons. The distance between the code cache and the rest of the ares executable has a large influence on performance on all platforms because there are microarchitecural thresholds beyond which branches will never be correctly predicted by the CPU.

To make performance more consistent on macOS, we can make this dynamic allocation sooner, at the top of main(), before the address space has become too cluttered with allocations made by other code.

### DIFF
--- a/ares/ares/memory/fixed-allocator.cpp
+++ b/ares/ares/memory/fixed-allocator.cpp
@@ -2,12 +2,15 @@
 
 namespace ares::Memory {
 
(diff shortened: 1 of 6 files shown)


hackgrid:
Just wanted to document for the changelog of 134, that this change roughly doubled performance on M1 in some cases! :-) Thx!

PR #1187, opened 2023-08-10
desktop-ui: add fast load option to load menu

hackgrid:
I discovered you can drag & drop any game ROM to ares and it will load without choosing a emulator (mia can identity the correct emulator).
I like this and I want to use this via the menu too :)

If a file cannot be recognized or its 64DD stuff or something, there will be a error displayed, just like with the existing drag & drop functionality.

Looks like this:
grafik

### DIFF
--- a/desktop-ui/presentation/presentation.cpp
+++ b/desktop-ui/presentation/presentation.cpp
@@ -374,7 +374,24 @@ auto Presentation::loadEmulators() -> void {
     }
   }
   loadMenu.append(MenuSeparator());
+  
+  { MenuItem loadFastAction{&loadMenu};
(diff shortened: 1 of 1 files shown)


LukeUsher:
This is exactly why this feature isn’t exposed in the menu; when there are a lot of systems it doesn’t work for (ALL cd based systems, for example), it’s more unfriendly to the user to expose it; I’m not sure if everyone would agree or not but I’m reluctant to ship a menu item that doesn’t work for around 20% of our supported systems.

hackgrid:
I see! I guess you could add a core browser or something after the file does not get recognized, but that does not make sense to duplicate this already existing functionality...

PR #1210, opened 2023-08-29
Update README.md: Fixed typo

PeterSamuels:
Fixed typo (descendent → descendant)


hackgrid:
After checking varous dictionaries, "descendent" is correct too. :)

PeterSamuels:
> After checking varous dictionaries, "descendent" is correct too. :)

Only as an adjective ("is descendent from"). As a noun, it has to be an a.

PR #1223, opened 2023-09-15
GDB Server

HailToDodongo:
This PR adds a generic GDB server to ares with the ability to be used by all systems.
For now only the N64 systems was implemented here.

Most documentation on how the server/TCP socket works can be found in comments.
A detailed documentation for ares developers on how to add this to other systems is included in the `Readme.md` file.

The feature itself was also already tested by a few people on both linux and windows (with WSL) and seems to be stable now.


LukeUsher:
I don't have time to do a full review tonight; but a couple of quick notes I got from a glance over:

1. Consider squashing this PR into a single commit; there are a lot of changes and it's quite a noise merge
2. There are conflicts, please rebase on latest master
3. Since the GDB server is independent of ares itself, it might be better suited as part of nall rather than ares; consider moving ares/debug-server to nall/gdb/server

HailToDodongo:
Do you mean a manual squash? because usually i would just squash on merge in PRs.

Ok, i'm gonna get the latest master.

The issue i have with nall is, that right now i was careful to not make this a default include for everything.
Compilation times are already pretty bad in ares, and making changes in nall especially is taking a very long time to compile.
Not sure if that's something that would need fixing in general.

But i can move it there, as an optional include if that's ok.

LukeUsher:
> Do you mean a manual squash? because usually i would just squash on merge in PRs.
> 
> Ok, i'm gonna get the latest master.
> 
> The issue i have with nall is, that right now i was careful to not make this a default include for everything. Compilation times are already pretty bad in ares, and making changes in nall especially is taking a very long time to compile. Not sure if that's something that would need fixing in general.
> 
> But i can move it there, as an optional include if that's ok.

My thinking here is that nall is a generic stdlib of useful functionality that can be used by other applications; GDB server seems like a useful fit there; but yes it should be optional; not every product would need it so there's no need to make the headers global.

For merge strategy; we tend to use rebase merges here; in fact, all other merge types including squash are disabled
I'm open to trying squash merges, however.

HailToDodongo:
I've now moved the GDB code to nall, keeping it as an optional include.

For the conflicts+rebase issue, i think they are related.
I just pulled in the newest master into my branch, yet github still shows conflicts that don't exist.
If only rebases are allowed that might be the issue, since i just do a regular pull to get new changes from master into my branches.

So i would prefer trying a squash merge here, otherwise this would probably need something more involved.

On a side note: it would be a good idea to have a basic contribution guide in the readme, to mention things like this.
So other people don't run into this too.

HailToDodongo:
Ok just checked on my fork in github, rebase has conflicts, merge and squash don't have them

asiekierka:
I'd say a squash merge is fine given the context; preserving the full commit history in the pull request can be useful for reasoning about why/how certain code decisions were made.

nall/gdb/server.cpp
```diff
@@ -0,0 +1,535 @@
...
+
+          auto sepIdxMaybe = cmdName.find(",");
+          u32 sepIdx = sepIdxMaybe ? sepIdxMaybe.get() : 1;
+
+          u64 address = cmdName.slice(1, sepIdx-1).hex();
+          u64 unitSize = cmdName.slice(sepIdx+1, 1).hex();
+          u64 value = cmdParts.size() > 1 ? cmdParts[1].hex() : 0;
+
+          hooks.write(address, unitSize, value);
+          return "";
```
asiekierka: I think 'M' should return "OK" here, according to spec?

HailToDodongo: Ah that's right, interesting that this worked properly then.
Gonna test if fixing it breaks something else.

nall/gdb/server.cpp
```diff
@@ -0,0 +1,535 @@
...
+          u64 regValue = cmdName.slice(sepIdx+1).hex();
+
+          return hooks.regWrite(regIdx, regValue) ? "OK" : "E00";
+        }
+      break;
+
+      case 'q':
+        // This tells the client what we can and can't do
+        if(cmdName == "qSupported"){ return {
+          "PacketSize=", MAX_PACKET_SIZE,
```
asiekierka: This isn't strictly harmful, but at least LLDB parses this as a hexadecimal value, not a decimal value; might want to verify if that is the case on GDB too.

HailToDodongo: Looking through gdb's sources, it seem to be hex too.. gonna fix that then

nall/gdb/server.cpp
```diff
@@ -0,0 +1,535 @@
...
+    insideCommand = false;
+    cmdBuffer = "";
+    haltSignalSent = false;
+    forceHalt = false;
+    singleStepActive = false;
+
+    currentThreadC = -1;
+    hasActiveClient = false;
+    handshakeDone = false;
+    requestDisconnect = false;
```
asiekierka, line 533: Missing `nonStopMode = false;`. Even if it doesn't work yet, it's good habit to reset it on a new connection.

HailToDodongo:
The 3 issues are fixed now (and i fixed a missing clarification to the readme)

LukeUsher:
Let's go, we can refine as we go

PR #1229, opened 2023-09-17
fc: Fully utilize NES 2.0 extended PRG/CHR size variables.

asiekierka:
This PR may necessitate more third-party verification than usual. It also allows for detection of the correct MMC5 board based on NES 2.0 header alone.

### DIFF
--- a/mia/medium/famicom.cpp
+++ b/mia/medium/famicom.cpp
@@ -162,6 +162,19 @@ auto Famicom::analyzeINES(vector<u8>& data) -> string {
   if(iNes2) {
     mapper |= ((data[8] & 0xf) << 8);
     submapper = data[8] >> 4;
+
+    prgrom |= (data[9] & 0xf) * 0x400000;
+    chrrom |= (data[9] >> 4) * 0x200000;
+
+    u8 prgramShift =   (data[10] & 0xf);
+    u8 eepromShift =   (data[10] >> 4);
+    u8 chrramShift =   (data[11] & 0xf);
+    u8 chrnvramShift = (data[11] >> 4);
+    prgram =  prgramShift == 0 ? 0 : 64 << prgramShift;
+    eeprom =  eepromShift == 0 ? 0 : 64 << eepromShift;
+    chrram = (chrramShift == 0 ? 0 : 64 << chrramShift)
+           + (chrnvramShift == 0 ? 0 : 64 << chrnvramShift);
+
     u32 timing = data[12] & 3;
 
     // TODO: add DENDY (pirate famiclone) timing
@@ -190,7 +203,7 @@ auto Famicom::analyzeINES(vector<u8>& data) -> string {
   case   1:
     s += "  board:  HVC-SXROM\n";
     s += "    chip type=MMC1B2\n";
-    prgram = 8192;
+    if(!iNes2) prgram = 8192;
     break;
 
   case   2:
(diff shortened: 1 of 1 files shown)


invertego:
This will definitely break stuff if it goes in as-is. I tried to make a similar change earlier this year before I got burnt out trying to validate it. Happy to share my validation efforts, though I'll need a little time to dig through my mess.

asiekierka:
Does that imply there are NES 2.0-marked ROMs with invalid metadata?

invertego:
As you've probably noticed, the mapping between NES 2.0 fields and ares is not 1:1.

- NES 2.0 does not distinguish between prg-nvram and eeprom, but ares does.
- ares does not distinguish between prg-ram and prg-nvram, but NES 2.0 does.
- Generally speaking, this means prg-ram and prg-nvram should be added together (much like you've done with the chr-ram and chr-nvram fields) and assigned to `programRAM` unless it's an eeprom mapper (16, ~153~, 157, 159).
- For mapper 157, the eeprom field in NES 2.0 refers to the external eeprom, whereas in ares the internal eeprom size is expected (the external eeprom is not yet implemented).
- NES 2.0 supports encoding prg-rom and chr-rom fields in exponent-multiplier notation when necessary (e.g. Galaxian), but this PR does not implement it.

I think dealing with these things will be enough to correctly handle all licensed ROMs, at least. Of course, ares has a database for those, but with proper NES 2.0 header support we *ought* to get things right even without a db match.

LukeUsher (changes requested):
Quoting invertego: 

> As you've probably noticed, the mapping between NES 2.0 fields and ares is not 1:1.
> 
> * NES 2.0 does not distinguish between prg-nvram and eeprom, but ares does.
> * ares does not distinguish between prg-ram and prg-nvram, but NES 2.0 does.
> * Generally speaking, this means prg-ram and prg-nvram should be added together (much like you've done with the chr-ram and chr-nvram fields) and assigned to `programRAM` unless it's an eeprom mapper (16, 153, 157, 159).
> * For mapper 157, the eeprom field in NES 2.0 refers to the external eeprom, whereas in ares the internal eeprom size is expected (the external eeprom is not yet implemented).
> * NES 2.0 supports encoding prg-rom and chr-rom fields in exponent-multiplier notation when necessary (e.g. Galaxian), but this PR does not implement it.
> 
> I think dealing with these things will be enough to correctly handle all licensed ROMs, at least. Of course, ares has a database for those, but with proper NES 2.0 header support we _ought_ to get things right even without a db match.

I believe these should be addressed before merging,

asiekierka:
The listed issues should be fixed now.

mia/medium/famicom.cpp
```diff
@@ -668,10 +703,10 @@ auto Famicom::analyzeINES(vector<u8>& data) -> string {
     s += "      volatile\n";
   }
 
-  if(eeprom) {
+  if(prgnvram) {
     s += "    memory\n";
     s += "      type: EEPROM\n";
-    s +={"      size: 0x", hex(eeprom), "\n"};
+    s +={"      size: 0x", hex(prgnvram), "\n"};
     s += "      content: Save\n";
```
invertego: I think this is a mistake. prgnvram is reset to 0 above, so I would expect eeprom to be referenced here. Please do try a game from each of the eeprom mappers, if you can. If not, we can ask a tester to take a look.

asiekierka: It is a mistake, sorry. I typically rely on test ROMs, but the EEPROM mappers don't have much in this way; I'd appreciate a tester taking a more extensive look as described above.

invertego (approved):

PR #1248, opened 2023-09-29
Add "Games" app category to bundle .plist on macOS to facilitate Sonoma's Game Mode

jcm93:
macOS 14 (Sonoma)'s Game Mode offers some small benefits for ares, including lower latency for bluetooth peripherals and potential minor performance gains. The system has various ways of detecting that an application is a game and enabling game mode, but ares does not currently engage it. Adding the "Games" app category key value pair to the bundle .plist is one way of hinting to macOS that Game Mode should be enaged. In my testing, adding this key to ares's bundle .plist successfully activates Game Mode when ares is fullscreened.

image


hackgrid:
Looks good! :-)

PR #1270, opened 2023-10-25
change n64 1x res

FitzRoyX:
Seems to work. Please test


LukeUsher:
There has been significant discussion on this in discord but the TLDR Is that the N64 is designed to be displayed at 640x480 and scaling back to 320x240 is not *right*

I think we're looking in the wrong place for a solution here: the issue people are having is unexpected window size changes

I believe window size and screen scaling should not be so connected

Just let users set a window size and we scale to that according to preference (best fit, stretch, integer, etc) rather than the behaviour we currently have,

PR #1272, opened 2023-10-26
N64 Mia Update - Chameleon Twist 2

remutro:
Update Mia DB to include US/EU Ids for Chameleon Twist 2.  The Japanese Id starts with NV2 while the other regions have an Id that starts with N2V.


kev4cards:
Looks good to me. Resolves the warning screens about the controller pak not being attached.

PR #1274, opened 2023-10-28
fc: Initial EPSM implementation.

asiekierka:
More information: https://www.nesdev.org/wiki/Expansion_Port_Sound_Module

While this is a homebrew device, it has already shipped to consumers in this manner.

TODO (I'm not sure how to fix these issues myself):

- [x] Switching from EPSM back to another peripheral (None or Family Keyboard) unloads not just the additional EPSM-provided audio streams, but seems to break *all* audio from the NES core.
- [x] NES 2.0 cartridges can declare that they wish to have the EPSM peripheral installed by default (Extended Console Type 0x04), but I'm not sure how to hook this nicely from the pak data to the expansion port node.
- [ ] Missing serialization, but this affects the Family Keyboard to some limited extent already, so...

Not urgent at all. It has shipped to consumers, but the amount of software using it is still lacking.

### DIFF
--- a/ares/fc/cpu/memory.cpp
+++ b/ares/fc/cpu/memory.cpp
@@ -15,6 +15,11 @@ inline auto CPU::readBus(n16 address) -> n8 {
 }
 
 inline auto CPU::writeBus(n16 address, n8 data) -> void {
+  // The EPSM can be mapped via $4016 writes, but also using a "passthrough"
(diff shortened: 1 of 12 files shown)


LukeUsher:
Not strictly related to this but I'm curious as to why it was added as an extended console type rather than as an expansion device (byte 15) within the NES2.0 spec

Any insight on to that?

PR #1287, opened 2023-11-15
desktop-ui: disable color bleed for HD content

invertego:
This effect runs on the CPU, which is expensive for HD content and incredibly so for UHD content. It doesn't make much sense to apply to non-SD content, in any case.

Currently this only affects the N64 core in HD and UHD modes.


invertego:
In response to a user-reported performance regression in nightly compared to v133.

PR #1288, opened 2023-11-15
n64: put dd rtc into a struct + serialize rtc + update serializer version

LuigiBlood:
I seperated the RTC into its own struct as it can be easily seperated, and added serialization for it for save states.
Tested working.


remutro:
Since you have modified serialization/save states, I believe you should increase the serialization version in your pull request. It's the first line in ares/n64/system/serialization.cpp

LuigiBlood:
should I name it 134.1?

LuigiBlood:
(eventually just looked at previous commits and figured it out)

PR #1292, opened 2023-11-18
ruby: queue no more than "latency" samples in wasapi driver

### LINKED ISSUE #958: [General issue] Input delay when fast forwarding
### Describe the bug
The default audio driver WASAPI has issues when using fast forward. What I mean by that is that every time the fast forward key is released, there is an input delay for like a second where the game stops completely. This issue does not occur with XAudio 2.1.

### To Reproduce
1. Start a SFC ROM and fast forward a little bit.

invertego:
The WASAPI driver in ruby is a bit of an oddball. It has a fixed length circular queue for samples, and before this commit there was no logic to prevent queuing more than the requested "latency" number of samples or even to prevent the queue from overflowing entirely.

In practice this wasn't much of an issue when audio or video synchronization were enabled and therefore bounding the rate of audio generation. It did not, however, play nice with the fast-forward feature, and it caused a noticeable pause after returning to normal speed as the queue drained from whatever level of fullness it had wrapped around to (effectively random but averaging over half a second).

Now the length of the sample queue is (mostly) bounded to the "latency" parameter which is more consistent with the other drivers.

Fixes #958

### DIFF
--- a/ruby/audio/wasapi.cpp
+++ b/ruby/audio/wasapi.cpp
@@ -78,8 +78,13 @@ struct AudioWASAPI : AudioDriver {
     self.queue.count++;
 
     if(self.queue.count >= self.bufferSize) {
+      //this event is signaled at the device period which is no more than half of bufferSize
+      //(in shared mode) or equal to bufferSize (in double-buffered exclusive mode)
       if(WaitForSingleObject(self.eventHandle, self.blocking ? INFINITE : 0) == WAIT_OBJECT_0) {
         write();
+      } else {
+        self.queue.read++;
+        self.queue.count--;
       }
     }
   }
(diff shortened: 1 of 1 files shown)


ruby/audio/wasapi.cpp
```diff
@@ -188,7 +193,7 @@ struct AudioWASAPI : AudioDriver {
       waveFormat = *(WAVEFORMATEXTENSIBLE*)waveFormatEx;
       CoTaskMemFree(waveFormatEx);
       if(self.audioClient->GetDevicePeriod(&self.devicePeriod, nullptr)) return false;
-      auto latency = max(self.devicePeriod, (REFERENCE_TIME)self.latency * 10'000);  //1ms to 100ns units
+      auto latency = max(self.devicePeriod * 2, (REFERENCE_TIME)self.latency * 10'000);  //1ms to 100ns units
```
invertego, line 196: The change to this line is motivated by an abundance of caution rather than an observed need. On my machine, the default period is 10 ms and WASAPI won't let me set a buffer duration less than 22 ms. In principle, because shared streams only have a single buffer, it wouldn't make much sense to even allow the buffer duration to be less than double the period. I don't know for sure though if that's enforced by WASAPI, and thus the change here.

invertego:
I've tested this change in shared and exclusive modes and with synchronization enabled and disabled for both the audio and video streams. I'll note that fast-forwarding does sound different with this change because the queue is no longer just being allowed to fill up with ~1.5 seconds of audio until it wraps around. Perceptually speaking, the difference is something like this:

Before: audio plays at normal speed but periodically jumps ahead
After: audio plays at fast-forward speed with noticeable aliasing (because we're just dropping samples)

Subjectively, the "before" behavior may sound better to some. The "after" behavior matches all the other drivers, however.

PR #1296, opened 2023-11-19
mia: improve CIC detection algorithm

rasky:
Currently, we detect the CIC to emulate by calculating a CRC32 over IPL3 bootcode in ROM, and keeping a database of the known IPL3s. This works because right now there are only a handful of IPL3 variants (exactly one per CIC).

In preparation for libdragon releasing an open source IPL3 that could evolve and be forked an unbounded number of times, we must stop keeping a database of IPL3s.

This commit changes strategy: it simulates the IPL2 checksum hash over IPL3 (which is exactly what a real N64 does at boot) and verifies if the checksum matches that expected by any of the known CIC variants. This works because open source IPL3 variants must anyway have a checksum that collides with that of CICs (which must be obtained via a second preimage attack, bruteforced on GPUs), otherwise they would not work on real hardware nor on Ares itself (which correctly simulates the boot sequence and would refuse to boot an "unsigned" IPL3).

### DIFF
--- a/mia/medium/nintendo-64.cpp
+++ b/mia/medium/nintendo-64.cpp
@@ -4,7 +4,83 @@ struct Nintendo64 : Cartridge {
   auto load(string location) -> bool override;
   auto save(string location) -> bool override;
   auto analyze(vector<u8>& rom) -> string;
+  auto ipl2checksum(u32 seed, array_view<u8> rom) -> u64;
 };
+
+auto Nintendo64::ipl2checksum(u32 seed, array_view<u8> rom) -> u64 {
+  auto rotl = [](u32 value, u32 shift) -> u32 {
+    return (value << shift) | (value >> (-shift&31));
+  };
+  auto rotr = [](u32 value, u32 shift) -> u32 {
+    return (value >> shift) | (value << (-shift&31));
+  };
+
+  auto csum = [](u32 a0, u32 a1, u32 a2) -> u32 {
+    if (a1 == 0) a1 = a2;
+    u64 prod = (u64)a0 * (u64)a1;
+    u32 hi = (u32)(prod >> 32);
+    u32 lo = (u32)prod;
+    u32 diff = hi - lo;
+    return diff ? diff : a0;
+  };
+
+  // create the initialization data
+  u32 init = 0x6c078965 * (seed & 0xff) + 1;
+  u32 data = rom.readm(4);
+  init ^= data;
+
+  // copy to the state
+  u32 state[16];
+  for(auto &s : state) s = init;
+
+  u32 dataNext = data, dataLast;
+  u32 loop = 0;
+  while(1) {
+      loop++;
+      dataLast = data;
+      data = dataNext;
+
+      state[0] += csum(1007 - loop, data, loop);
+      state[1]  = csum(state[1], data, loop);
+      state[2] ^= data;
+      state[3] += csum(data + 5, 0x6c078965, loop);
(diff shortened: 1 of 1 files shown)


mia/medium/nintendo-64.cpp
```diff
@@ -4,8 +4,97 @@ struct Nintendo64 : Cartridge {
   auto load(string location) -> bool override;
   auto save(string location) -> bool override;
   auto analyze(vector<u8>& rom) -> string;
+  auto ipl2checksum(u32 seed, array_view<u8> rom) -> u64;
 };
 
+auto Nintendo64::ipl2checksum(u32 seed, array_view<u8> rom) -> u64
+{
+  auto rotl = [](u32 value, u32 shift) -> u32 {
+    return (value << shift) | (value >> (32 - shift));
```
invertego: Shifting a u32 by 32 bits is UB. Same for rotr.

Might be good to move the rotate helpers to nall/bit.hpp. We can eventually just switch to the C++ standard library implementation when we move to C++20.

https://en.cppreference.com/w/cpp/numeric/rotl

rasky: Done, but I don't want to move it to nall as I don't want to think now how to make that a template.

mia/medium/nintendo-64.cpp
```diff
@@ -4,8 +4,97 @@ struct Nintendo64 : Cartridge {
...
+  u32 state[16];
+  for (auto &s : state) s = init;
+
+  u32 dataNext, dataLast;
+  u32 dataIndex = 0;
+  u32 loop = 0;
+  while (1) {
+      loop++;
+      dataLast = data;
+      data = read(dataIndex);
```
invertego: FYI, array_view has a built-in helper for big endian reads.
```
      data = rom.readm(4);
      rom += 4;
```

mia/medium/nintendo-64.cpp
```diff
@@ -4,8 +4,97 @@ struct Nintendo64 : Cartridge {
   auto load(string location) -> bool override;
   auto save(string location) -> bool override;
   auto analyze(vector<u8>& rom) -> string;
+  auto ipl2checksum(u32 seed, array_view<u8> rom) -> u64;
 };
 
+auto Nintendo64::ipl2checksum(u32 seed, array_view<u8> rom) -> u64
+{
```
invertego: Nit: brace style

mia/medium/nintendo-64.cpp
```diff
@@ -4,8 +4,97 @@ struct Nintendo64 : Cartridge {
...
+      state[10]  = csum(state[10] + data, dataNext, loop);
+      state[11]  = csum(state[11] ^ data, dataNext, loop);
+      state[12] += state[8] ^ data;
+  }
+
+  u32 buf[4];
+  for (auto &b : buf) b = state[0];
+
+  for (loop = 0; loop < 16; loop++)
+  {
```
invertego: Nit: brace style

invertego (commented):
Looks good, just some minor suggestions.

mia/medium/nintendo-64.cpp
rasky: Done, though `readm` already advances the pointer

mia/medium/nintendo-64.cpp
rasky: Done

mia/medium/nintendo-64.cpp
rasky: Done

PR #1297, opened 2023-11-20
build: fix level 2 msvc warnings

invertego:
Level 2 warnings are considered "significant" and turning them on revealed one issue that could, in the future, have affected MSVC-ABI builds using either CL or Clang.

MSVC's C++ name mangling distinguishes between classes and structs, and throughout the ares codebase there were friend declarations using "class" to refer to types that were defined with "struct".

I don't see any value in retaining this inconsistency, especially considering that ares otherwise avoids the "class" keyword (though I spotted a few uses that crept in from recent commits).

### DIFF
--- a/ares/ares/scheduler/scheduler.hpp
+++ b/ares/ares/scheduler/scheduler.hpp
@@ -41,7 +41,7 @@ private:
   vector<Thread*> _threads;
   bool _synchronize = false;
 
-  friend class Thread;
+  friend struct Thread;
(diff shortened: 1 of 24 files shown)


invertego:
I looked at level 3 and the additional warnings were all ones we'd just disable. With that in mind, there may not be much value in raising the level above 2. My main concern is to catch stuff that is moderately likely to generate bad code.

Issue #1305, opened 2023-11-29
[N64] [macOS] [arm64] Crash on CPU allocator flush when using recompiler


jcm93:
**Describe the bug**
Ares crashes when flushing the N64 CPU code cache on Apple Silicon. Not sure if unique to macOS or arm64, this is just my test environment. At least one report that it is not reproducible on x64. I narrowed down the offending commit by performing a bisect and found that the behavior is present since 8ed4af8b57bb5a67a9976e9dbd58db1b75152a2f.

**To Reproduce**
Any N64 game seems to trigger the crash. Tested with Snowboard Kids, Majora's Mask, Star Wars Episode I: Racer. It can be reproduced quickly by reducing the size of the cache in [`ares/n64/cpu/recompiler.cpp:27`](https://github.com/ares-emulator/ares/blob/2a819403db36a585fa5b1a268196f46e47d2be1f/ares/n64/cpu/recompiler.cpp#L27). I replaced `1_MiB` with `63_MiB` for testing. With that small a cache, most games will crash on startup or soon afterward.

**Expected behavior**
Ares should not crash when flushing the cpu code cache :(

Edit: I should probably attach a crash trace. [ares-2023-11-29-135053.ips.zip](https://github.com/ares-emulator/ares/files/13505787/ares-2023-11-29-135053.ips.zip)

LukeUsher:
Fixed as of #1306

PR #1306, opened 2023-11-30
recompiler: allow jitprotect() calls to be nested

invertego:
The following applies only to macOS when JIT page protection is enforced (namely on Apple silicon).

A sequence of jitprotect(false) calls must now be followed by an equal number of jitprotect(true) calls to restore executable protection. A further assumption is that nesting always starts with (false), which is only logical because you have to write something before you have anything to execute. This is explicitly checked in debug builds.

The N64 and PS1 recompilers have needed this since commit 8ed4af8 in order to handle the path taken on recompiler cache flushes. It's subtle, but emit() will evict all pools and so the second call to pool() from block() will reallocate the pool and unexpectedly toggle the protection out from under it.

Technically the SH2 recompiler has needed this since commit 241f0e9 (from 8 months ago!) but that same commit made cache flushes practically nonexistent, so it's likely not impacting the retail title library.

### DIFF
--- a/nall/memory.hpp
+++ b/nall/memory.hpp
@@ -198,7 +198,15 @@ template<u32 size, typename T> auto writem(void* target, T data) -> void {
 inline auto jitprotect(bool executable) -> void {
   #if defined(PLATFORM_MACOS)
   if(__builtin_available(macOS 11.0, *)) {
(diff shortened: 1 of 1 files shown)


invertego:
Fixes #1305

PR #1313, opened 2023-12-01
ruby: add SDL to macOS

jcm93:
### Description
Up to this point, SDL as an input library has not been present in macOS builds. The plumbing was already in place for macOS to make use of SDL, but SDL was not actually linked. This PR rectifies that. Seemingly unnecessary references to the IOKit joypad are also removed from `sdl.cpp`.

To facilitate adding SDL to ruby for macOS, a stubbed out macOS mouse driver, `gcmouse.cpp`, has been added, to avoid needing to make numerous modifications to `sdl.cpp` to work around mouse-related code.
### Motivation
The legacy macOS input uses raw values from IOKit, which are not available for all supported controllers, and where present, are uncalibrated and generally not ideal. SDL's input support is much more advanced comparatively, properly handling and interpolating input from a wide variety of controllers where raw IOKit input values aren't sufficient. This should resolve a number of situations where ares won't recognize input from controllers connected to macOS machines, among other macOS input-related issues.

### Discussion
I am not super well-acquainted with ares's code and build system, and it's entirely probable that I missed something here or went about this not-quite-correctly; I did so as straightforwardly as possible and may have missed something, for example fallbacks if SDL is not present at build-time; I am not sure if this is currently handled. I encourage any and all edits from maintainers if I broke something or went about something incorrectly.

### Testing?
Tested locally on my Apple Silicon machine on macOS Sonoma. SDL can now be selected and used successfully as an input driver for all supported connected controllers.

### DIFF
--- a/ruby/GNUmakefile
+++ b/ruby/GNUmakefile
@@ -19,6 +19,7 @@ ifeq ($(ruby),)
     ruby += video.cgl
     ruby += audio.openal
     ruby += input.quartz #input.carbon
+    ruby += input.sdl
   else ifeq ($(platform),linux)
     pkg_check = $(if $(shell pkg-config $1 && echo 1),$2)
     ruby += video.glx video.glx2 video.xshm
--- a/ruby/input/mouse/gcmouse.cpp
+++ b/ruby/input/mouse/gcmouse.cpp
@@ -1,0 +1,32 @@
+#pragma once
+
+struct InputMouseGC {
+  Input& input;
+  InputMouseGC(Input& input) : input(input) {}
+
+  uintptr handle = 0;
+  bool mouseAcquired = false;
+
+  struct Mouse {
+  } ms;
+
+  auto acquired() -> bool {
+    return false;
+  }
+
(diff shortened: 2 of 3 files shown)


LukeUsher:
Do you think it would be possible to build an SDL Mouse driver that could be used on all platforms? I suspect it might not be possible since the window is NOT an SDL  window but it's worth investigating.

Alternatively we do need a native, working, fallback because without the mouse driver, it is impossible to play software using Light Guns, or the Sega/N64/etc mouse.

jcm93:
I'm guessing that you might be able to use SDL as a mouse driver across Win/Linux/macOS, but it might take someone more familiar with ares' codebase than myself. I think you would need to create SDL windows from the various native window handles (gtk, qt, cocoa, win) and go from there, which seems like it might be a bit involved.

There currently is no macOS mouse driver in ares period, SDL or no, as I understand, so I think I was thinking of this PR as being a purely an improvement on existing functionality for controllers, and a mouse driver could be PR'd separately.

PR #1328, opened 2023-12-06
hiro: add cursor auto-hide

jcm93:
### Description
Adds a cursor auto-hide option for when the cursor has been idle for at least two seconds and is over the game window, and the application is focused. Currently only implemented via the Cocoa UI on macOS in hiro.

To `hiro/core/window`, the `hidesCursor` stored property is added, along with a setter, so a window knows whether it should use the auto-hide behavior. To `desktop-ui/settings/settings`, the `autoHideCursor` property is added to `general`, and a corresponding option has been added under a new Miscellaneous header under "Options" in the Configuration window. An option is also added to the main menu bar via `desktop-ui/presentation/presentation.cpp` and state there is kept in sync with the Configuration window option.

Screenshot 2023-12-05 at 6 03 44 PM
Screenshot 2023-12-05 at 12 57 11 PM

Toggling the option calls out to specifically the presentation window, calling `setHidesCursor()`. Correspondly, during initialization, the option only is referenced when starting up the presentation window (`desktop-ui/presentation/presentation.cpp:230`), as other windows should never auto-hide the cursor. 

On the concrete implementation side, in Cocoa, `hidesCursor` defaults to `false` (`hiro/cocoa/window.cpp:9`).
> [!IMPORTANT]
> Future implementations for other UIs should initialize `hidesCursor` in the concrete `window.cpp` implementation as `false`, so the cursor is never hidden over non-game windows.

The Cocoa implementation is fairly straightforward; on window initialization, an `NSTimer` is added to the run loop that ticks every second and calls `hideCursor()`, and an `NSTrackingArea` is created for the bounds of the window on every size event. `hideCursor()` hides the cursor if `hidesCursor` is true, at least 2 seconds idle have elapsed, and the cursor is within the tracking area for the window. Cursor presence in the window is tracked with `-(void)mouseEntered:(NSEvent *)` and `-(void)mouseExited:(NSEvent *)` and the `cursorIsInWindow` property. Cursor is hidden simply via `[NSCursor setHiddenUntilMouseMoves:YES];`

### Motivation
Hiding the mouse cursor when it falls idle over the game window is typical for most games and emulators, and is probably behavior that users will generally want.

### Discussion / Possible Issues
Currently, the option is only implemented on macOS, and so the configuration window and menu bar options are currently hidden behind a simple `#ifdef PLATFORM_MACOS`. I am not sure if I will be able to effectively test this for other platforms and UIs since my only machine is Apple Silicon (though I will try), so I wanted to go ahead and get this in flight for review as is just in case.

In terms of UI/UX, I am not sure if the current location is the best place for the option to live in the Configuration window. Currently, the option lives both there as well as the menu bar. I considered only having the option in the menu bar as well, which might fit better. Currently there are no other options in ares that are settable both via the menu bar as well as the Configuration window. I will of course defer to maintainers' thoughts on this.

### DIFF
--- a/desktop-ui/presentation/presentation.cpp
+++ b/desktop-ui/presentation/presentation.cpp
@@ -15,6 +15,11 @@ Presentation::Presentation() {
 
   settingsMenu.setText("Settings");
   videoSizeMenu.setText("Size").setIcon(Icon::Emblem::Image);
+  
+  bool hideCursorImplemented = false;
+#if defined(PLATFORM_MACOS)
+  hideCursorImplemented = true;
+#endif
 
   //generate size menu
   u32 multipliers = 5;
@@ -95,6 +100,14 @@ Presentation::Presentation() {
   muteAudioSetting.setText("Mute Audio").setChecked(settings.audio.mute).onToggle([&] {
     settings.audio.mute = muteAudioSetting.checked();
   });
+  autoHideCursorSetting.setText("Auto-Hide Cursor").setChecked(settings.general.autoHideCursor).onToggle([&] {
+    settings.general.autoHideCursor = autoHideCursorSetting.checked();
+    settingsWindow.optionSettings.autoHideCursorOption.setChecked(settings.general.autoHideCursor);
+    presentation.setHidesCursor(settings.general.autoHideCursor);
+  });
+  if(!hideCursorImplemented) {
+    autoHideCursorSetting.setVisible(false);
+  }
   showStatusBarSetting.setText("Show Status Bar").setChecked(settings.general.showStatusBar).onToggle([&] {
     settings.general.showStatusBar = showStatusBarSetting.checked();
     if(!showStatusBarSetting.checked()) {
@@ -226,6 +239,8 @@ Presentation::Presentation() {
       program.load(emulator, filename);
     }
   });
+
+  setHidesCursor(settings.general.autoHideCursor);
 
   iconLayout.setCollapsible();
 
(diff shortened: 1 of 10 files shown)


FitzRoyX:
> Hiding the mouse cursor when it falls idle over the game window is typical for most games and emulators, and is probably behavior that users will generally want.

That doesn't sound accurate and it seems incredibly lazy to me. Like, you can use the mouse for hours doing all sorts of stuff, but the moment you load a rom file it becomes a burden to flick your cursor out of view? To the point of needing hundreds of lines of code and OS specific solutions?

Also, the cursor phasing out of existence basically means that when you're done playing, you have to do that flick anyway to know where to move the mouse. So you're never really getting rid of the act, you're just shifting it to another time.

jcm93:
> That doesn't sound accurate and it seems incredibly lazy to me. Like, you can use the mouse for hours doing all sorts of stuff, but the moment you load a rom file it becomes a burden to flick your cursor out of view? To the point of needing hundreds of lines of code and OS specific solutions?
> 
> Also, the cursor phasing out of existence basically means that when you're done playing, you have to do that flick anyway to know where to move the mouse. So you're never really getting rid of the act, you're just shifting it to another time.

Hmm. I think the main justification for hiding the cursor in programs like ares isn't exactly removing the burden of moving the mouse out of the way. Rather, it's the lack of reason for the cursor to be visible when it already *is* over the window. Assuming the person using ares is playing a game with either the keyboard or, more often, a controller, there isn't anything cursor-interactive over the window. With no normal reason for the cursor to be there, if it *is* there, it's convenient to hide it, or at least to have the option to do so.

I do not think it's a stretch to say that this is a standard form of behavior for programs with non-cursor-interactive content. Such as media player windows, and games where a controller is the input device.

The other main justification would be that, if ares is in full screen, which is a fairly normal way to play, you can't actually make the cursor not visible on some platforms. It's always slightly visible even at the edge, depending on the desktop and compositor and other settings. This is not unique to macOS. This is at least somewhat distracting for a decent amount of people, which I again don't consider a huge stretch to say.

As far as OS-specific solutions, there seem to be a fair amount of those in ares, where and when appropriate, to make sure the desired behavior is smooth and correct. I took some amount of care to make sure this feature behaved well, with parity to other games and applications. To my eye, this is the sort of thing that can and should involve platform-specific APIs since mouse cursors are treated differently on different platforms.

PR #1336, opened 2023-12-08
Mega Mouse support for 32X + CD

remutro:
Doom Resurrection patch version 3.0 (currently at version 3.1) added support for the Mega Mouse, as well as making use of Mega CD hardware if available. When I added mouse support for the Mega CD, the patch for Doom Resurrection was at version 2.2, and since there was no licensed 32X software that supported the Mega Mouse I declined to add support at the time. But leave it to homebrew, especially one as comprehensive as this patch to add support. 

This adds Mega Mouse support for both the 32X and 32X/CD combo, so it is now available for all Mega * consoles.


viciious:
Just my 2 cents: D32XR feature Mega Mouse support since v1.0 ;)

PR #1338, opened 2023-12-08
ColecoVision - Added "Super Action Controller" support 

Kalixtan:
Added support for the "Super Action Controller".

I an unsure how exactly the best setup for the controls when it comes to that binding thing. Unsure exacty how its ment to be layed out, but this works and works in the controller test roms.

This is kinda a hack job with copied and modified code (controllers are nearly identical electronically to the original sock version just with two extra keys added).

Also note the keys for this controller are typically listed from F1 to F4 renaming the L and R buttons... though that said cant find meeny first hand docs nameing those buttons on the oringal controller other than button 1 and button 2... so maybe they should be renamed in the emulator?.. idk, up to you lot.

Let me know of any problems.

### DIFF
--- a/ares/cv/controller/controller.cpp
+++ b/ares/cv/controller/controller.cpp
@@ -4,5 +4,6 @@ namespace ares::ColecoVision {
 
 #include "port.cpp"
 #include "gamepad/gamepad.cpp"
+#include "super_action/super_action.cpp"
 
 }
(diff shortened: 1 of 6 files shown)


Kalixtan:
Wait you cant press F3 and F4 at the same time and i completely forgot the roller existed on these controllers....

How can you unpull something?

PR #1340, opened 2023-12-09
Famicom/NES - Added SNES-2-NES Controller Adaptor support.

Kalixtan:
Dout anyone will use it as only a very few bits of homebrew use the extra inputs. But the NES-2-SNES cable is added, all keys work.
If anyone wants i can add other controllers other than the "GamePad" one useing this adapter,but yet to see any software that uses them.

The other way is possible but doubtful its worth emulating as it should just work identically to the SFC controller with buttons missing.

### DIFF
--- a/ares/fc/controller/sfcgamepad/sfcgamepad.cpp
+++ b/ares/fc/controller/sfcgamepad/sfcgamepad.cpp
@@ -1,0 +1,94 @@
+SFC_Gamepad::SFC_Gamepad(Node::Port parent) {
+  node = parent->append<Node::Peripheral>("Super Famicom Gamepad");
+
+  up     = node->append<Node::Input::Button>("Up");
+  down   = node->append<Node::Input::Button>("Down");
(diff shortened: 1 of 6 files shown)


ares/fc/controller/sfcgamepad/sfcgamepad.cpp
```diff
@@ -0,0 +1,94 @@
+SFC_Gamepad::SFC_Gamepad(Node::Port parent) {
```
LukeUsher, line 1: nit: We don't usually use underscores in class names; consider `SuperFamicomController`, as SFCController looks awkward.

Kalixtan: Make whatever changes you want, dont have time to deal with this right now, personal things are happening in my life and its likey 5 seconds for you anyways....

Same with the other pull request for those Coleco-Vision Mappers.

PR #1350, opened 2023-12-16
macOS: Add minimum macOS version and copyright to ares.plist

shinra-electric:
Added the minimum macOS version key, but kept it at 10.9 which is what macOS currently reports it as. 
Now that the key is here, it will be easier to update it in the future. 

Also added the copyright line attributing the ares team and contributors.


shinra-electric:
It's also possible to add a version number and build number to the plist, but it would be preferable if that was some global variable.

PR #1370, opened 2024-01-21
n64: reenable isviewer even outside of homebrew mode

rasky:
ISViewer is currently gated by the homebrew mode flag. It seems like this is not ideal for rom hackers because they would still need a way to log but homebrew mode is too verbose for them and hide the ISViewer messages in streams of actual problems in game code that they don't intend to fix. So restore ISViewer functionality without homebrew mode.


Dragorn421 (approved):
Perfect

Thanks a lot!

PR #1389, opened 2024-02-13
Continue if xcode-select fails building SDL

jcm93:
The `build-SDL.sh` script I added was not fit for local builds because it selects specific Xcode builds that are only normally present on CI machines. This PR changes the script to continue if those commands fail, allowing `./buildSDL.sh` to work without modifications for users building locally.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,10 +1,6 @@
 name: Build
 on:
-  push:
-    branches: [ master ]
-    tags: [ 'v*' ]
-  pull_request:
-    branches: [ master ]
+  push
 jobs:
   build:
     strategy:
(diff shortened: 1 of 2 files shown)


LukeUsher:
wouldn't a better solution be doing something like xcode-select --print-path to get the current path before changing it, then restoring it after the build?

jcm93:
It would, though the `Xcode_14.2.app` naming conventions are only used by GitHub Actions, so these are basically set up to fail on any local machines anyway.

Honestly it would be better to just not touch `xcode-select` in any way unless we know we are on CI. I might look into a better way to handle this in general.

jcm93:
Going to close this for now while I do a bunch of tests on this branch so I don't spam anyone with notifications. Will open a new one later.

PR #1391, opened 2024-02-13
n64: revert previous commit (add HLE RDRAM initialization)

rasky:
This is actually still required as we still skip RDRAM initialization anyway and proprietary IPL3 doesn't initialize the RAM size in this case.


loganmc10:
If you're interested in getting rid of this hack, I know that mupen64plus has done it, see here: https://github.com/mupen64plus/mupen64plus-core/pull/501

PR #1393, opened 2024-02-14
Make macOS SDL build portable, build script runnable locally

jcm93:
Resubmission of https://github.com/ares-emulator/ares/pull/1389 with improvements.

Now the SDL build script will not call `xcode-select`, which can have undesired effects on the user build environment, unless it is on CI (detected with the `$GITHUB_ACTIONS` env var).

It will also not call ``cmake --install .``, which would previously replace the user's SDL installation.

This second change required some minor changes to ruby's makefile. We now include the header directory from the cloned SDL source, and inform the linker correctly of the location of the SDL .dylib.

Also added the cloned SDL repository and .dylib to the .gitignore for easier development workflow.

To ensure compatibility with https://github.com/ares-emulator/ares/pull/1390, the `MACOS_COMPILED_SDL` flag is added so we can include the proper "SDL.h" header for compiled builds and `<SDL2/SDL.h>` header for system builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ obj-amd64/
 obj-arm64/
 out-amd64/
 out-arm64/
+thirdparty/SDL/SDL
+thirdparty/SDL/libSDL2-2.0.0.dylib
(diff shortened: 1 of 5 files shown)


jcm93:
Noticed an issue with this while fixing something else (was linking SDL even if `sdl2=false`). If there's other issues with this let me know!

PR #1425, opened 2024-03-24
ws: APU emulation improvements

asiekierka:
* Significantly improved Hyper Voice implementation accuracy (verified, at least for Sound DMA writes, with a test ROM)
* Implemented emulation of unused/undocumented debug ports (0x64-0x67, 0x69, 0x95-0x9B)
* Improved cycle accuracy of APU channel handling
* Added slightly faster "inaccurate" APU emulation mode (for most games, output should be essentially identical; debug ports will not match up, though)

### DIFF
--- a/ares/ws/apu/apu.cpp
+++ b/ares/ws/apu/apu.cpp
@@ -50,7 +50,7 @@ auto APU::main() -> void {
     case 122:
       // TODO: Sound DMA should be blocking the CPU instead.
       dma.run();
-      io.output = {};
+      apu.sequencerClear();
       break;
     case 123: if(channel5.io.enable)                      channel5.runOutput(); break; // TODO: Which cycle is this performed on?
     case 124: if(channel1.io.enable)                      channel1.output(); break;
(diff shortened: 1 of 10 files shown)


LukeUsher (commented):
Nice work

Regarding the two apu implementations, how big of a performance difference are we talking? Is it enough to make it worth maintaining two implementations?

asiekierka:
Last I checked, it was 2-3%, but the only difference between the two codepaths is a "manually unrolled" main loop, so I don't think it's a big burden to maintain. It is possible to get a bit more optimization out of it, though (notice how we call `tick()` on the channels 128 times? This could be done once with some clever rewriting).

The main benefit is that, combined with the "inaccurate" PPU which tries to group `step()` calls, the CPU core can run for much longer uninterrupted.

asiekierka:
(Also, for context, the tester ROM's source code is here: https://github.com/asiekierka/ws-test-suite/tree/main/src/tools/hyper_voice_tester )

PR #1426, opened 2024-03-26
Make Alt+Tab work correctly when in fullscreen

CasualPokePlayer:
Using WS_EX_TOPMOST on a fullscreen window prevents Alt+Tab from switching to any different window. This is a bit of an annoyance as it ruins the point of Alt+Tab, especially in the case of a non-exclusive mode being set. It however is very very bad in any case ares hangs, and thus prevents being able to switch to Task Manager to actually kill ares. This means the average user could potentially run into a situation requiring them sign out or shutdown their PC, and thus potentially lose data in other apps.

It also happens to be the case that if the user attempts to Alt+Tab while in exclusive fullscreen, ares will hang for whatever reason. This commit doesn't fix that hang, but it does at least allow the user to Alt+Tab out of the situation.

Note that the windows are being set with WS_DISABLED more to avoid what I believe is a Windows bug. Normally WS_EX_NOACTIVATE should prevent clicking on the window from actually activating the window. However, if you were to Alt+Tab to some other window, then have that other window being shunk so ares ends up shown, you can proceed to click on the ares window, and it will actually activate that window. This thus takes focus away from ares' presentation window with generally no way to give it focus (assuming inputs get blocked). WS_DISABLED prevents mouse input from working on the fullscreen window (which it normally doesn't do anything due to WS_EX_NOACTIVATE, only in the weird case I laid out does it do something)

Also this commit likely breaks macOS/Linux (haven't tested / can't test those), probably need to check for focus differently depending on platform defines?

### DIFF
--- a/desktop-ui/input/hotkeys.cpp
+++ b/desktop-ui/input/hotkeys.cpp
@@ -145,7 +145,7 @@ auto InputManager::pollHotkeys() -> void {
   if(Application::modal()) return;
 
   if(!driverSettings.inputDefocusAllow.checked()) {
-    if (!presentation.focused() && !ruby::video.fullScreen()) return;
+    if (!presentation.focused()) return;
   }
 
   for(auto& hotkey : hotkeys) {
--- a/desktop-ui/program/platform.cpp
+++ b/desktop-ui/program/platform.cpp
@@ -173,7 +173,7 @@ auto Program::audio(ares::Node::Audio::Stream node) -> void {
 
 auto Program::input(ares::Node::Input::Input node) -> void {
   if(!driverSettings.inputDefocusAllow.checked()) {
-    if(!ruby::video.fullScreen() && !presentation.focused()) {
+    if(!presentation.focused()) {
       //treat the input as not being active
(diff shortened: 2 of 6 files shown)


LukeUsher:
If anybody tests this on macOS or Linux (myself included, if I manage to find time today), please comment the result here so we can verify that it has been tested

jcm93:
This breaks input while fullscreen on macOS with OpenGL but not Metal. Will take a look and see if I can suggest a change after investigating.

edit: It only breaks input if using the ares fullscreen setter; platform-native fullscreen just using the window controls works fine.

I don't fully understand why the macOS OpenGL fullscreen setter is the way that it is. I set up Metal fullscreen differently for my own reasons, but I'm reluctant to change OpenGL to do the same thing if there are legit reasons to do it this way for older versions of macOS or something.

mayhair:
The GNOME desktop environment for Linux also implements Alt+Tab. A bit surprisingly, it encounters the same problem as you have described for Windows. It is still possible to change windows via the Super key, however.
(I'm running Void Linux x86_64 glibc, kernel version 6.6.32_1, GNOME version 45.5.)

PR #1430, opened 2024-03-29
ares/screen: Reduce excessive polling for new frames

jcm93:
ares currently polls every microsecond in a dedicated thread looking for new frames. This is excessively often, and uses a large amount of processing power.

There is probably a more elegant way to handle this polling, but changing the polling to happen every millisecond instead of every microsecond seems like a reasonable start.

In my testing, this change reduces ares' CPU use by 25-30%, without adversely impacting performance in any way. 

This change should be tested on other platforms to make sure there are no adverse effects that might not be accounted for.

### DIFF
--- a/ares/ares/node/video/screen.cpp
+++ b/ares/ares/node/video/screen.cpp
@@ -25,7 +25,7 @@ Screen::~Screen() {
 
 auto Screen::main(uintptr_t) -> void {
   while(!_kill) {
-    usleep(1);
+    usleep(1000);
     if(_frame) {
       refresh();
       _frame = false;


jcm93:
Per discussion offthread, this seems to cause flickering issues specifically with the PS1 core, and seemingly only the PS1 core so far in testing.

Preliminary thoughts are that this could be because of double buffering done by some PS1 titles, with this longer sleep interval somehow causing the backbuffer to be presented, but more investigation is still needed.

LukeUsher:
An alternate implementation has been introduced here https://github.com/ares-emulator/ares/commit/b145f0f000c4e0ddedf25a62cdcfe713deea1f6a

I used condition variables such that the thread wakes when notified of a new frame, with a 10ms timeout to prevent deadlocks when no frames are generated.

Thanks for the idea!

PR #1432, opened 2024-04-01
refactor(mos6502): refactor cpu mos6502

ZengGengSen:
1. add cpu mos6502 generate script
    2. refactor some instruction


asiekierka:
What is the benefit of this pull request? Does it fix any bugs or improve performance?

From a maintenance perspective, it appears to add more net code and require the programmer to learn a separate domain-specific language (the `.lst` format); it also might hurt code integration with IDEs, as a result.

ZengGengSen:
> What is the benefit of this pull request? Does it fix any bugs or improve performance?
> 
> From a maintenance perspective, it appears to add more net code and require the programmer to learn a separate domain-specific language (the `.lst` format); it also might hurt code integration with IDEs, as a result.

1. Implemented all mos6502 instructions, including Unofficial opcode and JAM (freeze CPU)
(It passed the 'instr_test_v5' and 'instr_timing' tests)
2. It takes the form of generating code to generate the code that the instruction runs, so it is faster than multiple function calls

For the lst format, I did not consider code integration with the 'IDE'. However, it will increase the cost of understanding and reading for developers, and its essence should be simple substitution (similar to macros).

If this causes maintenance problems, I will close the pull request and submit a new pull request that can be integrated with the ID

asiekierka:
> Implemented all mos6502 instructions, including Unofficial opcode and JAM (freeze CPU)

That is a good change; though it would require ensuring that all the devices using the mos6502 core in Ares have the same behaviour with regards to "illegal" opcodes.

> It takes the form of generating code to generate the code that the instruction runs, so it is faster than multiple function calls.

Did you check how big the improvement is in benchmarks? We're "paying" with development cost, so it's good to know how big the "return" is, to speak metaphorically.

First of all, Ares makes use of modern compiler optimizations extensively to allow making code more readable. Compiler-level inlining should already avoid the cost of many function calls; in addition, merely duplicating code into functions can be *slower* in some edge cases due to the behaviour of CPU caches.

Second, Ares is *not* designed as a performance-first emulator; the fact it relies on coroutine scheduling alone makes it uncompetitive in that regard - however, it makes the code much more readable, and it makes it cheaper/easier to make changes which improve compatibility. The policy for Ares and its preceding emulators was to be "fast enough for a modern desktop CPU", putting all the remaining effort into cleaner code or better accuracy instead.

LukeUsher:
I would accept the emulation changes to this PR, but can we please do so without the code generation/.lst method? It doesn't really fit the coding style/overall design of ares to do things this way, even if the 6502 is well suited to this approach.

PR #1434, opened 2024-04-02
v30mz: minor fixes

asiekierka:
Fixes some undefined behaviour handling and the BOUND opcode, in accordance with [WSCpuTest](https://github.com/FluBBaOfWard/WSCpuTest/).

### DIFF
--- a/ares/component/processor/v30mz/v30mz.hpp
+++ b/ares/component/processor/v30mz/v30mz.hpp
@@ -51,7 +51,7 @@ struct V30MZ {
   template<u32> auto setAccumulator(u32) -> void;
 
   //modrm.cpp
-  auto modRM() -> void;
+  auto modRM(bool forceAddress = false) -> void;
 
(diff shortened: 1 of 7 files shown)


ares/component/processor/v30mz/v30mz.hpp
```diff
@@ -251,6 +251,8 @@ struct V30MZ {
 
     u16 segment;
     u16 address;
+
+    bool useAddress;
```
remutro, line 255: Does this new field need to be added to v30mz/serialization.cpp with the rest of the members of this struct?

If yes, then that also means the serializer version for the WonderSwan should be changed in: 
ws/system/serialization.cpp

asiekierka: Ah, I forgot, sorry!

remutro: Looks good, thanks!

PR #1441, opened 2024-04-03
ws: implement sound test bits 1, 2, 3

asiekierka:
Up to date with the current research status quo.

### DIFF
--- a/ares/ws/apu/serialization.cpp
+++ b/ares/ws/apu/serialization.cpp
@@ -19,6 +19,8 @@ auto APU::serialize(serializer& s) -> void {
   s(io.seqDbgOutputForce55);
   s(io.seqDbgChForce4);
   s(io.seqDbgChForce2);
+  s(io.seqDbgNoise);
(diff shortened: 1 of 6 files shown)


ares/ws/apu/serialization.cpp
```diff
@@ -19,6 +19,8 @@ auto APU::serialize(serializer& s) -> void {
   s(io.seqDbgOutputForce55);
   s(io.seqDbgChForce4);
   s(io.seqDbgChForce2);
+  s(io.seqDbgNoise);
```
asiekierka, line 22: I assume there's no need to bump the serialization version if there hasn't been a release since the last bump?

LukeUsher: I think it's still beneficial; failure to bump will still break save states between nightly builds. sometimes in non-obvious ways.

asiekierka: Should be fixed now!

PR #1442, opened 2024-04-04
fix(apu): fix apu framecounter

ZengGengSen:
detail:
    1. Improve the emulational accuracy of apu framecounter
    2. pass the apu_test/4-jitter.nes

  todo:
    1. Does apu need to emulate blocking_write_4017?

  question:
    1. A white screen will appear when running blargg_apu_2005.07.30/04.clock_jitter.nes

### DIFF
--- a/ares/fc/apu/apu.cpp
+++ b/ares/fc/apu/apu.cpp
@@ -9,6 +9,7 @@ APU apu;
 #include "triangle.cpp"
 #include "noise.cpp"
 #include "dmc.cpp"
+#include "framecounter.cpp"
 #include "serialization.cpp"
 
 auto APU::load(Node::Object parent) -> void {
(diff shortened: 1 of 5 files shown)


LukeUsher:
Regarding the question: I can confirm the issue, until we get to the bottom of it, try this attached jitter test
[[blargg]apu_test^4-jitter.zip](https://github.com/ares-emulator/ares/files/14864970/blargg.apu_test.4-jitter.zip)

Update: oh, I see you already did, oops.

LukeUsher:
Can you clarify what you mean by blocking_write_4017?

ZengGengSen:
In the emulation of Mesen's APU FrameCounter, I saw `Do not allow writes to 4017 to clock the frame counter for the next cycle (i.e this odd cycle + the following even cycle)`

But I didn't see the relevant description in the NesDev wiki, and I'm not sure whether I should add it.

ZengGengSen:
> Regarding the question: I can confirm the issue, until we get to the bottom of it, try this attached jitter test [[blargg]apu_test^4-jitter.zip](https://github.com/ares-emulator/ares/files/14864970/blargg.apu_test.4-jitter.zip)
> 
> Update: oh, I see you already did, oops.

In https://www.nesdev.org/wiki/MMC5_audio 
MMC5 does not have an equivalent [frame sequencer](https://www.nesdev.org/wiki/APU_Frame_Counter) (APU $4017); envelope and length counter are fixed to a 240hz update rate.

So there's no need to fix it

PR #1447, opened 2024-04-09
screen/video: Add refresh rate hint mechanism

jcm93:
This constitutes one puzzle piece that will help improve ares's frame pacing, particularly for variable refresh rate displays.

While continued work on ares's audio drivers is likely to smooth the pace at which ares generates video frames, transient changes in application and system load inevitably means that frames will be delivered earlier or later than expected. Additionally, relying on video frames to be generated precisely when needed and presenting them immediately is fairly demanding. If we can pace frames some distance in advance, it can significantly lessen the burden on the CPU and GPU.

This PR is one tool to help the video backend deal with these issues. Cores can now inform the screen of their expected rate of frame generation at any given time, and the screen will inform the video backend. The backend can then employ various strategies to perform optimal frame pacing and drop frames when necessary.

The function is emphasized as a 'hint' to indicate that the rate is only that which is expected, rather than certain. Unpredictable changes in system conditions mean that we can never be entirely certain when a frame will be generated or presented.

> [!NOTE]
> No concrete `refreshRateHint` implementations are added in this PR. Given the variety of core refresh rates across ares, as well as the consideration that for some cores refresh rates can change during runtime, that work is left for separate PRs.

### DIFF
--- a/ares/ares/node/video/screen.cpp
+++ b/ares/ares/node/video/screen.cpp
@@ -72,6 +72,11 @@ auto Screen::setRefresh(function<void ()> refresh) -> void {
   _refresh = refresh;
 }
 
+auto Screen::refreshRateHint(double refreshRate) -> void {
+  lock_guard<recursive_mutex> lock(_mutex);
+  platform->refreshRateHint(refreshRate);
+}
+
 auto Screen::setViewport(u32 x, u32 y, u32 width, u32 height) -> void {
   lock_guard<recursive_mutex> lock(_mutex);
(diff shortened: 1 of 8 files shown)


LukeUsher:
A potentially useful side-effect of this is  being able to use it in the VPS counter to report a percentage of the speed we're running at, since we will know the current speed (VPS) and the intended (hint) where applicable.

LukeUsher:
Happy to merge the API; implementations can come later

PR #1449, opened 2024-04-10
fix(fc-apu): pass some apu test

ZengGengSen:
1. pass apu_reset/4015_cleared.nes
  2. pass apu_reset/4017_timing.nes
  3. pass apu_reset/4017_written.nes
  4. pass apu_reset/irq_flag_cleared.nes
  5. pass apu_reset/len_ctrs_enabled.nes
  6. pass apu_reset/works_immediately.nes

### DIFF
--- a/ares/fc/system/system.cpp
+++ b/ares/fc/system/system.cpp
@@ -109,6 +109,14 @@ auto System::power(bool reset) -> void {
   apu.power(reset);
   ppu.power(reset);
   scheduler.power(cpu);
+
+  // The CPU takes 8 cycles before it starts executing
+  // the ROM's code after a reset/power up
+  for (auto i : range(8)) {
+    apu.main();
+    cartridge.main();
+    (cpu.Thread::step)(cpu.rate());
+  }
 }
 
 }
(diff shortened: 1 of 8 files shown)


ares/fc/system/system.cpp
```diff
@@ -109,6 +109,14 @@ auto System::power(bool reset) -> void {
   apu.power(reset);
   ppu.power(reset);
   scheduler.power(cpu);
+
+  // The CPU takes 8 cycles before it starts executing
+  // the ROM's code after a reset/power up
+  for (auto i : range(8)) {
```
LukeUsher, line 115: This part looks rather suspect to me; since it is calling main functions of other portions out of band as well as accessing cpu.thread directly.


Consider updating CPU::power() in ares/fc/cpu/cpu.cpp to just call step(rate() *8 ); instead: I will work on implementing the proper 6502 reset process in a later commit, but this change should be enough to allow us to merge.

LukeUsher (changes requested):

ares/fc/system/system.cpp
ZengGengSen: The cpu will be in `scheduler.power(cpu);` reset clocks, in `cpu.power(reset);` call `step(rate() * 8);` It didn't work

PR #1457, opened 2024-04-20
fix windows tcp timeout

zZeck:
Gotta love questionable platform differences. On Windows the old code causes a disconnect after recv times out in 1 second, because errno will always be 0. [Windows does not use errno, or EAGAIN](https://learn.microsoft.com/en-us/windows/win32/winsock/error-codes-errno-h-errno-and-wsagetlasterror-2).
Now when the platform is windows [WSAGetLastError](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsagetlasterror) is checked for [WSAETIMEDOUT](https://learn.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2).

I could maybe do some more improvements in the file, (probably in another PR) but I don't know what the project is interested in.
I did not address other uses of errno in the file.
I also did not switch from strerror, which windows does not use, to [FormatMessage](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-formatmessage).

### DIFF
--- a/nall/tcptext/tcp-socket.cpp
+++ b/nall/tcptext/tcp-socket.cpp
@@ -244,7 +244,11 @@ NALL_HEADER_INLINE auto Socket::open(u32 port, bool useIPv4) -> bool {
       } else if(length == 0) {
         disconnectClient();
       } else {
+        #if defined(PLATFORM_WINDOWS)
+        if (WSAGetLastError() != WSAETIMEDOUT) {
+        #else
         if (errno != EAGAIN) {
+        #endif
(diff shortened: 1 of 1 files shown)


rasky:
I think we should work towards a small abstraction layer for these differences to include in a .h file. I did something similar here:

https://github.com/rasky/ootmm-multi-client/blob/934981522da3279becfdb39a1f6eb71a8941c21d/src/MultiClient/multi.h#L4-L55

With a few macros like here, it is possible to hide most differences.

PR #1458, opened 2024-04-21
ruby: Remove tight loops in SDL and OpenAL

jcm93:
When we are synchronizing to audio and waiting for buffers to play back before continuing in the emulation thread, the SDL and OpenAL audio drivers currently wait inefficiently in tight loops that consume many CPU cycles. This PR adds `usleep` calls to these loops.

In the case of the SDL driver, we sleep for ~~the time it takes one quarter of our prescribed buffer size to play~~ precisely the amount of time to play our queued audio that's in excess of our prescribed buffer size.

For the OpenAL driver, it was more difficult for me to determine how to see exactly how much time we should be waiting since the loop seems to work slightly differently. One millisecond seemed reasonable given that the minimum latency is 20 milliseconds, and was sufficient to introduce significant savings without impacting performance.

Exactly how much CPU use this saves depends on a couple factors; primarily, the host CPU and the guest core (more demanding cores will spend a smaller share of time waiting in the audio driver). Reduction seems independent of audio latency. Here are some sample figures for the M1 Max:

SNES core (Super Mario World): 95-105% CPU use before, 40-50% CPU use after with both OpenAL and SDL.
N64 core (Super Smash Bros): 100-110% CPU use before, 60-70% CPU use after with both OpenAL and SDL.

(macOS's activity monitor measures activity as 'percent of one core', hence the over-100% figures).

Less demanding cores and more powerful systems will benefit more from this change, generally.

I tested this pretty thoroughly on my system to make sure that it did not impact performance on even the lowest 10ms SDL latency on the most demanding N64 core. However, other people putting this through its paces, particularly on less powerful systems, would be a good idea, to make sure that platform differences in `usleep` behavior don't have unexpected results.

### DIFF
--- a/ruby/audio/sdl.cpp
+++ b/ruby/audio/sdl.cpp
@@ -21,6 +21,8 @@ struct AudioSDL : AudioDriver {
 
   auto hasBlocking() -> bool override { return true; }
   auto hasDynamic() -> bool override { return true; }
+  
+  double bitsPerSample = 0;
 
   auto hasFrequencies() -> vector<u32> override {
     return {44100, 48000, 96000};
@@ -43,8 +45,15 @@ struct AudioSDL : AudioDriver {
     if(!ready()) return;
 
     if(self.blocking) {
-      while(SDL_GetQueuedAudioSize(_device) > _bufferSize) {
+      auto bytesRemaining = SDL_GetQueuedAudioSize(_device);
+      while(bytesRemaining > _bufferSize) {
         //wait for audio to drain
+        auto bytesToWait = bytesRemaining - _bufferSize;
+        auto bytesPerSample = bitsPerSample / 8.0;
+        auto samplesRemaining = bytesToWait / bytesPerSample;
+        auto secondsRemaining = samplesRemaining / frequency;
+        usleep(secondsRemaining * 1000000);
+        bytesRemaining = SDL_GetQueuedAudioSize(_device);
       }
     }
 
@@ -60,6 +69,10 @@ struct AudioSDL : AudioDriver {
 private:
   auto initialize() -> bool {
     terminate();
+    
+#if defined(PLATFORM_WINDOWS)
+    timeBeginPeriod(1);
+#endif
 
     SDL_InitSubSystem(SDL_INIT_AUDIO);
 
@@ -74,6 +87,7 @@ private:
     _device = SDL_OpenAudioDevice(NULL,0,&want,&have,0);
     frequency = have.freq;
     channels = have.channels;
+    bitsPerSample = SDL_AUDIO_BITSIZE(have.format);
     _bufferSize = have.size;
     SDL_PauseAudioDevice(_device, 0);
 
@@ -84,6 +98,9 @@ private:
   }
 
   auto terminate() -> void {
+#if defined(PLATFORM_WINDOWS)
+    timeEndPeriod(1);
+#endif
     _ready = false;
     SDL_CloseAudioDevice(_device);
     SDL_QuitSubSystem(SDL_INIT_AUDIO);
--- a/ruby/audio/openal.cpp
+++ b/ruby/audio/openal.cpp
@@ -67,6 +67,7 @@ struct AudioOpenAL : AudioDriver {
         _queueLength--;
       }
       //wait for buffer playback to catch up to sample generation if not synchronizing
+      usleep(1000);
       if(!self.blocking || _queueLength < 3) break;
     }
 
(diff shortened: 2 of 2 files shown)


ruby/audio/sdl.cpp
```diff
@@ -45,6 +47,11 @@ struct AudioSDL : AudioDriver {
     if(self.blocking) {
       while(SDL_GetQueuedAudioSize(_device) > _bufferSize) {
         //wait for audio to drain
+        auto bytesToWait = (_bufferSize / 4.0);
+        auto bytesPerSample = bitsPerSample / 8.0;
+        auto samplesRemaining = bytesToWait / bytesPerSample;
+        auto secondsRemaining = samplesRemaining / frequency;
+        usleep(secondsRemaining * 100000);
```
asiekierka: Shouldn't this be `1000000`? To convert from seconds to microseconds.

jcm93: Yes, it should. How embarrassing! This probably means that I should only be sleeping for `1 / 40.0` of the buffer length instead of `1 / 4.0` to keep my testing results consistent. Thanks for the catch.

jcm93:
With the feedback from @asiekierka addressed, I saw no downsides performance-wise in sleeping for precisely the difference between the buffer size and our amount of queued audio. Hence, I've changed SDL to use that approach, since it's preferable to having magic numbers.

jcm93:
Rarely this will sleep for very very large intervals for some reason. Converting to draft until I hit it again and can debug.

jcm93:
Fixed the above issue; it was an underflow in the bytesToWait value due to not accounting for possible queue draining from another thread while within the loop. Should be good for review.

LukeUsher:
On Windows, the default timer resolution that drives thread switches, etc is 15.6ms; this means it is difficult to sleep for smaller periods unless we change this setting.

There is an API for this (timeBeginPeriod / timeEndPeriod). Before Windows 10 this changes the resolution globally. If we forget to call timeEndPeriod on exit (for example, if we crash) the timer period remains what we set; this may cause unexpected battery drain on mobile devices, etc due to waking more than usual.

Starting from Windows 10 build 2004 this is no longer a concern, as the operating system now tracks timeBeginPeriod per-application and uses the lowest resolution specified by the currently running application.

Just something to consider; and likely required for this PR (and ares in general, if other areas assume short sleeps)

jcm93:
I just pushed an update that adds ifdef'd calls to `TimeBeginPeriod` and `TimeEndPeriod` in the `initialize()` and `terminate()` functions respectively in OpenAL and SDL. I tested it on Windows (albeit in a VM) to verify that SDL audio still worked as expected at 10ms latency. WASAPI seems to result in much better video frame pacing than SDL, but strangely, adding these sleeps actually resulted in improved SDL frame pacing for whatever reason.

I looked at a couple other things like a condition wait, but I couldn't use that without totally changing the structure of the SDL driver to use callbacks. Given the talk of significantly reworking audio drivers, not to mention SDL3's release soon, it didn't seem worthy of changing SDL2 that significantly. I also looked at directly calling NtDelayExecution, but it seemed like increasing timer resolution and simply calling `Sleep` would basically be fine for the time precision that we require (effectively, on Windows, this PR seems to always result in a sleep somewhere between 0 and 1 millisecond depending on where we are in the time slice).

OpenAL was not available on Windows in my testing.

I figure that this looks OK now, particularly given that SDL audio is (I think) little-used on Windows, and Mac/Linux should have `nanosleep`-based usleep. If you think it requires more restructuring, please do let me know though.

CasualPokePlayer:
> There is an API for this (timeBeginPeriod / timeEndPeriod). Before Windows 10 this changes the resolution globally. If we forget to call timeEndPeriod on exit (for example, if we crash) the timer period remains what we set; this may cause unexpected battery drain on mobile devices, etc due to waking more than usual.
> 
> Starting from Windows 10 build 2004 this is no longer a concern, as the operating system now tracks timeBeginPeriod per-application and uses the lowest resolution specified by the currently running application.

FWIW, in practice, this is often always 1 ms regardless as typical programs such as Discord and Steam also just set this to 1 ms

Also, the `uses the lowest resolution specified by the currently running application` isn't quite right, it's rather the lowest resolution specified by *any* process. If process A sets 5 ms and process B sets 10 ms, both process A and B get 5 ms. When process A proceeds to close, and assuming process B is still alive, it would then go to 10 ms.

(There is also an undocumented NT API (`NtSetTimerResolution`) which timeBeginPeriod/timeEndPeriod internally uses which can go even lower than 1 ms, going down to 0.5 ms. In practice, there's little benefit between 0.5 ms and 1 ms, although maybe useful if you want to squeeze frame timing consistency down to its absolute limits)

jcm93:
Fixed case issue (`timeBeginPeriod` vs `TimeBeginPeriod`, etc.).

PR #1463, opened 2024-04-29
infra: update macOS build scripts and SDK, bump MoltenVK to 1.2.7

jcm93:
This PR does a few things:

- Updates ares to create macOS builds with the latest SDK. After https://github.com/ares-emulator/ares/commit/65264090b06853f3d831f507a3335d9c42e8a198 there are no longer any outstanding issues doing so. Importantly, we use Xcode 15.2, since prior versions have issues with weak linking. 
- Adds `macos-compat` to CI as a compatibility build that maintains macOS 10.9 to 10.13 compatibility. The `macos-latest` runner will target 10.13 and above. The only difference with this build is the SDK version used to build it (Xcode 13.4); macOS weak linking handles the rest.
- Increments the MoltenVK version to 1.2.7. The version ares was building required a Python library `distutils` that is no longer trivially available with newer Python versions.
- Refactors the `xcode-select` usage reserved for GitHub Actions out of the SDL script and up to the top level, before any dependency scripts are run. This is just tidying up.
- Improves the librashader build script warning to specify that you need to add a target with the nightly toolchain to build librashader (from @rasky)
- Unbreaks the build.

This PR has been tested and verified to build in a local branch, but to be safe, it would probably be good to run it here as well.


jcm93:
For some reason, setting our Xcode version to 15.2 on the `macos-latest` runner causes CMake to look for the macOS 13.1 SDK for the C/CXX compiler when building SDL (?), and then die when it cannot find it. I do not want to build with the default 15.0.1 Xcode, because I know that this will cause a startup crash on macOS 11 and 12. Setting the Xcode version to 14.3.1 (available on `macos-latest`) to build SDL, and then switching back afterward, does work, so I've pushed an update that does that.

edit: It appears as though setting the SDL deployment target to 10.13 instead of 10.11 is necessary to make CMake not search for only the 13.1 SDK. I will add a change to modify the argument on macos-latest shortly.

jcm93:
The weirdness with CMake dying after not being able to find the correct macOS SDK was due to the CMake cache from the build data we had been storing. The SDL cache step has been removed as well in this PR.

Regardless, we now set the deployment targets for SDL correctly depending on the runner.

PR #1472, opened 2024-05-02
CI: Update upload-artifact and download-artifact to use v4

shinra-electric:
- v2 will be unavailable from the end of June 2024
- v4 can be up to 98% faster than previous versions according to [Github's notification](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/)

This PR only changes the version number, and I'm submitting it to see if it works. 
It is possible that further changes are necessary, as there were some [breaking changes](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes), but I don't think any of them apply to ares. 

Testing: 
- See if CI passes
- See if warnings regarding using v2 of `upload-artifact` and `download-artifact` are gone (There will still be other warnings regarding `checkout` and `cache`)
- (Optional) Compare times for the tasks using `upload-artifact` and `download-artifact` with previous CI runs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -181,7 +181,7 @@ jobs:
         MACOS_NOTARIZATION_PASSWORD: ${{ secrets.MACOS_NOTARIZATION_PASSWORD }}
         MACOS_NOTARIZATION_TEAMID: ${{ secrets.MACOS_NOTARIZATION_TEAMID }}
     - name: Upload
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v4
       with:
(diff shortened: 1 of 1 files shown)


shinra-electric:
- Upload artifact step across 9 jobs:  
  - Ranged from 2~8 minutes on the v138 CI run. Total: 45m 48s
  - Ranged from 2~12 seconds on this PR. Total: 1m 8s
- Total saving: 44m 40s
(Doesn't mean it finishes 45 minutes earlier because they are run in parallel, but still impressive)

- Download artifact step: 
  - 1 hour 2 mins on v138
  - Not performed on the PR as there was no release step

PR #1478, opened 2024-05-03
CI: Update cache and checkout to v4

shinra-electric:
- Updates `checkout` from v2 to v4
- Updates `cache` from v3 to v4

Reasons for change: 
- Checkout@v2 runs on node 12, which has been [discontinued](https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/)
- Checkout@v3 runs on node 16 and has also reached its [end of life](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/), so skip it entirely. 
- Cache@v3 runs on node 16 as well, so update to v4 for the same reason.

Testing: 
- Ensure CI passes
- Check that warnings regarding `checkout` and `cache` are gone

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -90,7 +90,7 @@ jobs:
         keychain-name: ares-macos-keychain
         keychain-password: ${{ secrets.MACOS_KEYCHAIN_PASSWORD }}
     - name: Checkout source code
-      uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


shinra-electric:
CI warnings dropped from 36 to 9

PR #1515, opened 2024-06-05
Support default device in WASAPI driver

### LINKED ISSUE #1514: One Bug, One Request
Bug: Whenever I switch audio sources, Ares doesn't default to it. For example, under Windows, switching from headphones to speakers, the emulator will still manually need to be switched to speakers in the Drivers menu.

Request: A Simple "Game" directory in the Paths menu, that you can navigate from a default location when going to load a game.

reflectronic:
The WASAPI sound driver now supports a "Default" audio output device:

This uses the [Automatic Stream Routing](https://learn.microsoft.com/en-us/windows/win32/coreaudio/automatic-stream-routing) API that was introduced in Windows 10. A runtime check was added to ensure that the feature does not appear on older versions of Windows.

I did some limited testing and found the feature to work without issues.

Closes https://github.com/ares-emulator/ares/issues/1514


LukeUsher:
I will accept this for now; but I do think we should look at manually implementing the fallback too for people using systems prior to win10.

PR #1517, opened 2024-06-08
Fix cl build break

reflectronic:
Fix build break caused by #1515. Let's see what CI says...


invertego:
There are actually two build breaks here, one of which is not your fault. The (confusingly named) ares-windows-msvc builds, which use clang, are broken due to https://github.com/actions/runner-images/issues/10001

reflectronic:
I fixed the linker error, this patch should be good to go now.

Issue #752, opened 2022-09-14
[NES] GFX artefacts with 31-in-1 (Asia) (En) (Pirate)...


BenjaminSiskoo:
Emulator : Ares 1.29
Nightly build 2022-09-13 17:07:05 UTC
Options : Default
Game : 31-in-1 (Asia) (En) (Pirate)
System : NES

![image](https://user-images.githubusercontent.com/12481242/190103386-e7d4b361-1c6d-4556-a5fd-505915d9b591.png)

BenjaminSiskoo:
Fixed with the last release

PR #1521, opened 2024-06-11
add board unl-bmc-31-in-1 (mapper 229)

ZengGengSen:
add board unl-bmc-31-in-1 (mapper 229)
pass issue #752


BenjaminSiskoo:
thanks :)

PR #1538, opened 2024-06-18
CI: Update wpilibsuite signing action for macOS 

shinra-electric:
This updates the [import signing certificate](https://github.com/wpilibsuite/import-signing-certificate) to the latest version. 
It should silence several warnings in the actions builds. 

Since this is security related, it needs to be thoroughly reviewed.  

I noticed that the wpilibsuite repo does not get updated very often, sometimes years between commits. Since this is security related, it might be an idea to switch to an alternative tool that's updated more often. 
There are a few options including [Apple Actions](https://github.com/Apple-Actions/import-codesign-certs) which looks like it's kept up to date on a regular basis.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,7 +83,7 @@ jobs:
     - name: "macOS: Import Certificate"
       if: runner.os == 'macOS' && github.event_name != 'pull_request'
       continue-on-error: ${{ github.repository != 'ares-emulator/ares' }}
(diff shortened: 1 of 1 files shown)


LukeUsher:
For audit purposes, the changes between both commits can be viewed here: https://github.com/wpilibsuite/import-signing-certificate/compare/2ac4f44d28045073d23153256efbb4c4b2d8aa22...9e858a4e87d254c4f1d39632ff810d78ee061302

jcm93:
This should be closeable now that the build system/CI has been updated to use the latest version of [apple-actions/import-codesign-certs@v3](https://github.com/Apple-Actions/import-codesign-certs) to import the certificate, and signing happens via `xcodebuild`.

PR #1542, opened 2024-06-20
Enforce the "C" locale for `LC_NUMERIC` (fix #1526)

### LINKED ISSUE #1526: [bug] ares uses locale-aware formatting for configuration
**Describe the bug**
ares settings inconsistently use locale-aware formatting for configuration (`settings.bml`)

**To Reproduce**
1. Be on linux / posix-y OS (?) (I am on Kubuntu Linux)
2. Run ares with French numeric locale `LC_NUMERIC=fr_FR.UTF-8 ares`
3. Open audio settings, change the volume, close ares
4. Open ares' `settings.bml` (e.g. `~/.local/share/ares/settings.bml`)
5. Observe `Volume: 0,460000`
6. Start ares again (any locale), open audio settings
7. Observe Volume is 0%
8. Bonus: Close ares, open `settings.bml` again, observe `Volume: 0,000000`

The issue does not happen with running ares with English numeric locale like `LC_NUMERIC=en_US.UTF-8 ares`

**Expected behavior**
Volume should be saved and loaded properly.
Configuration should not use locale-aware formatting at all.

**Additional context**
Kubuntu Linux
ares v138, built from source

Dragorn421:
Fixes #1526

The default locale in a C program is the "C" locale ( POSIX https://pubs.opengroup.org/onlinepubs/9699919799/functions/setlocale.html , C89 standard ), which we want for string-number formatting and parsing

gtk ( https://gitlab.gnome.org/GNOME/gtk/-/blob/a274a5ff4b2a2f9cf9731cdddbb456a0758fcbcc/gtk/gtkmain.c#L435 ) and Qt ( 5 https://codebrowser.dev/qt5/qtbase/src/corelib/kernel/qcoreapplication.cpp.html#597 , 6 https://codebrowser.dev/qt6/qtbase/src/corelib/kernel/qcoreapplication.cpp.html#613 ) set the locale from the environment variables instead. (by passing "" as the locale, cf `setlocale`)

This PR sets the locale for `LC_NUMERIC` back to "C" with `setlocale` after initializing gtk/Qt

I tested this with gtk, saw no problem in the GUI with a rom named `/home/dragorn421/Desktop/c'est la fête/rom de noël __  やあ！.z64`

Qt codepath is not tested

### DIFF
--- a/hiro/gtk/application.cpp
+++ b/hiro/gtk/application.cpp
@@ -145,6 +145,12 @@ auto pApplication::initialize() -> void {
   char** argvp = argv;
 
   gtk_init(&argc, &argvp);
+
+  // gtk_init causes GTK to set the locale from the environment.
+  // Set the locale for LC_NUMERIC back to "C". It is expected to be "C" for
+  // the purpose of various string formatting and parsing operations.
+  setlocale(LC_NUMERIC, "C");
+
   GtkSettings* gtkSettings = gtk_settings_get_default();
 
   //allow buttons to show icons
--- a/hiro/qt/application.cpp
+++ b/hiro/qt/application.cpp
@@ -124,6 +124,11 @@ auto pApplication::initialize() -> void {
   static char** argvp = argv;
   qtApplication = new QApplication(argc, argvp);
 
+  // Creating the QApplication causes Qt to set the locale from the environment.
+  // Set the locale for LC_NUMERIC back to "C". It is expected to be "C" for
+  // the purpose of various string formatting and parsing operations.
+  setlocale(LC_NUMERIC, "C");
+
   pKeyboard::initialize();
 }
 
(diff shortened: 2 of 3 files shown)


Screwtapello:
As I understand it, forcing the C locale prevents UTF-8 support. I don't know offhand if ares or nall try to do anything in a case-insensitive way, or count the number of code-points (as opposed to bytes) in a string or whatever, but that might be regressed without at least the `C.UTF-8` locale, if not the user's actual configured locale.

LukeUsher:
Yeah I am quite reluctant to change the entire locale and not just the numeric. It’s a shame this isn’t thread safe or the original solution would be fine.

Dragorn421:
Changed to only set `LC_NUMERIC` to C, after gtk/qt init
Testing with gtk, still good

Issue #1528, opened 2024-06-14
[NES] Alwa's Awakening (World) (Aftermarket) (Unl) - Messed screen...


BenjaminSiskoo:
nightly release on 2024-06-12 16:38:44 UTC

![image](https://github.com/ares-emulator/ares/assets/12481242/d6e97917-7627-4d5c-895a-bb4cffe6b42c)

![image](https://github.com/ares-emulator/ares/assets/12481242/baae4982-cc3f-44f7-8589-af547e50717e)

BenjaminSiskoo:
with the last release : nightly release on 2024-06-15 22:01:18 UTC

There is a message at the start : 

![image](https://github.com/ares-emulator/ares/assets/12481242/89db462a-9c5f-4dec-a9ea-639f647d2f22)

invertego:
"PRG RAM MISSING!" can be explained by the fact that ares doesn't emulate PRG RAM for UXROM because there never was any. I guess this another emulator extension of sorts that some unlicensed games take advantage of. Notably, this game is also exceeding the PRG ROM limits of UXROM, but ares already emulates this extended capacity, [expressly for the benefit of unlicensed games](https://github.com/ares-emulator/ares/commit/98aee19e93eb9565613e1e4fbb563e7a40e200a2).

Nesdev Wiki reference: https://www.nesdev.org/wiki/PRG_RAM_circuit
> The [iNES](https://www.nesdev.org/wiki/INES) format implies 8 KiB of PRG RAM at $6000–$7FFF, which may or may not be battery backed, even for [discrete boards](https://www.nesdev.org/wiki/Category:Discrete_logic_mappers) such as [NROM](https://www.nesdev.org/wiki/NROM) and [UxROM](https://www.nesdev.org/wiki/UxROM) that never actually had RAM there. This inspired some people on the nesdev.org BBS to come up with circuits to add PRG RAM to the original boards, so that games relying on it can run on an NES.

LukeUsher:
Fixed in #1557

PR #1557, opened 2024-07-08
fc: add prg ram to uxrom and cnrom mappers

invertego:
Some homebrew/aftermarket software calls for UXROM w/ PRG (NV)RAM. The licensed title Hayauchi Super Igo pairs CNROM with SRAM.


invertego:
Fixes #1528

PR #1570, opened 2024-07-17
n64: don't generate code for nops

ghost:
Instead of generating code for a no-op SLL instruction, replace it with a native NOP instruction that is more compact. This has no observable performance impact but shrinks the generated code by 10 bytes for each NOP instruction. In libdragon's gldemo this makes 1.6 bytes or 0.6% per block, on average.

The big idea is to gradually get rid of all possible writes to r0 in code generation so that we don't need to clear r0 after every instruction in the epilogue. This is a small step towards that goal.

### DIFF
--- a/ares/n64/cpu/recompiler.cpp
+++ b/ares/n64/cpu/recompiler.cpp
@@ -93,6 +93,11 @@ auto CPU::Recompiler::emit(u32 vaddr, u32 address, bool singleInstruction) -> Bl
 #define n26 u32(instruction & 0x03ff'ffff)
 
 auto CPU::Recompiler::emitEXECUTE(u32 instruction) -> bool {
+  //SLL r0, r0, 0
+  if (instruction == 0) {
+    return 0;
+  }
+
   switch(instruction >> 26) {
 
   //SPECIAL


ghost:
Here's the measurement I used for the numbers above. I started gldemo and dumped the size of each compiled block in bytes on my x64 linux machine.
[nop_comparison.csv](https://github.com/user-attachments/files/16267468/nop_comparison.csv) (actually tab separated)

LukeUsher:
I may be missing something but why bother emitting an instruction at all?

ghost:
I was worried that a block consisting of only nops would misbehave, and that if the nop was a jump target, not emitting it would cause that branch instruction to jump to the wrong location.

But looking at the `CPU::Recompiler::emit` function, it seems to always emit at least the function entry boilerplate, the per-instruction epilog, the "test jump epilog", and a "jump epilog". So it's not a risk that we'll end up with a `Block*` pointing to an allocation of 0 bytes.

Regarding jumps, it seems to me that the recompiler never emits x64 code that jumps in the middle of a block anyway. Each MIPS jump is implemented by looking up a `Block*` of its target address and executing that block. Is my understanding correct? If that's the case then it should be fine to just not emit anything.

ghost:
I updated the PR to just skip the nops. Also changed the title of this PR. I ran libdragon's gldemo and n64-systemtest successfully.

ghost:
Closing since this behavior is included in https://github.com/ares-emulator/ares/pull/1580

PR #1579, opened 2024-07-26
ps1: Fix Dragon Ball GT Final Bout (fix found by @LukeUsher)

bachimus:


hackgrid:
Just preserving what Luke said about this: "Seek was only pausing audio playback, not data playback due to an oversight."

PR #1580, opened 2024-07-26
n64: clear r0 only when it was written to

ghost:
Currently a per-instruction epilogue is needed to make sure the read-only register r0 stays as the value 0. To improve performance of the generated code, we'd like to move this operation to be done only when needed.

This commit solves the above problem in two ways: simple instructions are never emitted if their destination register is r0, and more complex instructions now emit a clear operation right after. This way exceptions are still thrown when needed.

### DIFF
--- a/ares/n64/cpu/cpu.cpp
+++ b/ares/n64/cpu/cpu.cpp
@@ -145,9 +145,12 @@ auto CPU::instructionEpilogue() -> s32 {
   if constexpr(Accuracy::CPU::Recompiler) {
     //simulates timings without performing actual icache loads
     icache.step(ipu.pc, devirtualizeFast(ipu.pc));
+    assert(ipu.r[0].u64 == 0);
   }
 
-  ipu.r[0].u64 = 0;
+  if constexpr(Accuracy::CPU::Interpreter) {
+    ipu.r[0].u64 = 0;
+  }
 
   switch(branch.state) {
   case Branch::Step: ipu.pc += 4; return 0;
--- a/ares/n64/cpu/recompiler.cpp
+++ b/ares/n64/cpu/recompiler.cpp
@@ -92,6 +92,10 @@ auto CPU::Recompiler::emit(u32 vaddr, u32 address, bool singleInstruction) -> Bl
 #define n16 u16(instruction)
 #define n26 u32(instruction & 0x03ff'ffff)
 
+auto CPU::Recompiler::emitZeroClear(u32 n) -> void {
+  if(n == 0) mov64(mem(IpuReg(r[0])), imm(0));
+}
+
 auto CPU::Recompiler::emitEXECUTE(u32 instruction) -> bool {
   switch(instruction >> 26) {
 
@@ -159,11 +163,13 @@ auto CPU::Recompiler::emitEXECUTE(u32 instruction) -> bool {
     lea(reg(2), Rs);
     mov32(reg(3), imm(i16));
     call(&CPU::ADDI);
+    emitZeroClear(Rtn);
     return 0;
   }
 
   //ADDIU Rt,Rs,i16
   case 0x09: {
+    if(Rtn == 0) return 0;
     add32(reg(0), mem(Rs32), imm(i16));
     mov64_s32(reg(0), reg(0));
     mov64(mem(Rt), reg(0));
@@ -172,6 +178,7 @@ auto CPU::Recompiler::emitEXECUTE(u32 instruction) -> bool {
 
   //SLTI Rt,Rs,i16
   case 0x0a: {
+    if(Rtn == 0) return 0;
     cmp64(mem(Rs), imm(i16), set_slt);
     mov64_f(mem(Rt), flag_slt);
     return 0;
@@ -179,6 +186,7 @@ auto CPU::Recompiler::emitEXECUTE(u32 instruction) -> bool {
 
   //SLTIU Rt,Rs,i16
   case 0x0b: {
+    if(Rtn == 0) return 0;
     cmp64(mem(Rs), imm(i16), set_ult);
     mov64_f(mem(Rt), flag_ult);
     return 0;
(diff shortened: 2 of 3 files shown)


ghost:
This change now passes n64-systemtest and I consider it ready for review.

ares/n64/cpu/recompiler.cpp
```diff
@@ -92,6 +93,10 @@ auto CPU::Recompiler::emit(u32 vaddr, u32 address, bool singleInstruction) -> Bl
 #define n16 u16(instruction)
 #define n26 u32(instruction & 0x03ff'ffff)
 
+auto CPU::Recompiler::emitZeroClear(u32 n) -> void {
+  if (n == 0) mov64(mem(IpuReg(r[0])), imm(0));
```
LukeUsher: nit: we tend to omit spaces between if and the braces, eg
`if(n == 0)`

There are other occurrences of this here too

LukeUsher (commented):
Just a minor nitpick from me

ares/n64/cpu/cpu.cpp
```diff
@@ -147,8 +147,6 @@ auto CPU::instructionEpilogue() -> s32 {
     icache.step(ipu.pc, devirtualizeFast(ipu.pc));
   }
 
-  ipu.r[0].u64 = 0;
```
LukeUsher, line 150: Actually: what about when the interpreter enabled? I didn't spot any code to clear r0 in that path

ghost: A good catch. I wrapped it in a constexpr check like the icache test above it.

ghost:
Passes n64-systemtest now with recompiler and interpreter. libdragon's gldemo also works.

invertego:
Sorry for missing out on most of the discussion before this change, but once we have some basic instruction analysis for the primary CPU like we do for the RSP (and I believe we'll want this anyway for a variety of reasons), it will become possible to accomplish this task with a much less invasive change to the recompiler. I'll send out a PR for the RSP to demonstrate.

rasky:
Yes but we are trying to do small improvements over time. We are trying to remove the epilogue, and getting close to that. We can't block changes behind large things that nobody is working on.

I think this change is very reasonable. We can revisit it later when we have an instruction analysis pass.

invertego:
I started working on it last weekend (collecting data), which is why I was pointing out instruction decoding inconsistencies. I just haven't had any time to work on it during the week.

invertego:
I don't mean to say this PR can't go in as-is, but I want to give fair warning of the potential impact from upcoming changes.

The approach I had in mind is demonstrated here in the RSP: #1581

The approach taken in this PR will generate slightly less code (compared to the approach in my PR) in some cases when R0 is used as a destination, but this seems to be extremely rare in practice (you can fail to clear it entirely and some games still run just fine), so I'm mostly concerned with 1) performance in the cases R0 is not assigned to, which is almost all the time and 2) correctness in the rare cases it is assigned.

ares/n64/cpu/cpu.cpp
```diff
@@ -147,7 +147,9 @@ auto CPU::instructionEpilogue() -> s32 {
...
```
invertego, line 150: Consider `assert(ipu.r[0].u64 == 0);` in the recompiler code path. This is a careful dance it's good to be on the safe side in debug builds.

ares/n64/cpu/recompiler.cpp
```diff
@@ -32,6 +32,7 @@ auto CPU::Recompiler::emit(u32 vaddr, u32 address, bool singleInstruction) -> Bl
 
   auto block = (Block*)allocator.acquire(sizeof(Block));
   beginFunction(3);
+  emitZeroClear(0);
```
invertego: Is this needed? Shouldn't this always be zero on entry?

invertego (changes requested):
LH must be handled as well.

rasky (approved):
For me the PR is fine and should go in, after the suggested changes

ares/n64/cpu/recompiler.cpp
ghost: You're right. I had missed the `for(auto& r : ipu.r) r.u64 = 0;` line in `CPU::power`.

ares/n64/cpu/cpu.cpp
ghost: I'll add that

ghost:
Thank you for the review, @invertego . Did I understand correctly that you'd prefer also the CPU to use something like `OpInfo op0 = self.decoderEXECUTE(instruction);` as your PR does on the RSP, but such functionality is not available yet? I do think checking for the rare r0 writes would be cleaner to do outside the `emitEXECUTE` function. I did have a conservative "analysis" function [like this](https://pastebin.com/raw/0va7CJUe) earlier so a full decoder wouldn't be necessary, but perhaps this negates the simplicity of your approach?

When it comes to performance, as you said, the r0 writes are exceedingly rare, so in practice only in the case of no-op `sll r0, r0, 0` there is any difference. That can be handled with a single `if (!instruction) return` anyway. So yeah correctness and code quality should take priority here.

Finally, I don't mind reverting this change when the analyser is ready and switching to that later. I just want to unblock other work on the recompiler in the meantime.

---

I fixed the mentioned issues and checked that the recompiler and interpreter both pass n64-systemtest.
**Edit:** Updated the commit message and PR title.

invertego:
Thanks @kannoneer. I believe your PR is good to go.

Re: future work, I now have a branch with a decoder like the one we have for the RSP:
https://github.com/invertego/ares/tree/n64_cpu_decoder

I'm not sure how much of that detail we'll ultimately need, though. I'm also not thrilled with the idea of yet another copy of basic instruction decoding logic that is already duplicated in three places: the interpreter, recompiler, and disassembler. This is already an issue with the RSP, though. Some consolidation would be nice, though it would take some refactoring.

invertego (approved):

PR #1584, opened 2024-07-31
gba: reset prefetch buffer on ROM accesses from DMA

png183:
Addresses todo item from #1377


png183:
Changed to draft - current version has a bug that can cause invalid prefetch buffer reads

png183:
Should be fixed now - will remove draft status once I run some more tests

png183:
Okay, it should be safe to merge now

PR #1588, opened 2024-08-02
ruby OSS: improve naming and usage of _bufferSize variable

bsdcode:
_bufferSize is used to store the number of bytes that can be written without blocking. It is not the size of the application buffer. Therefore rename this variable to _nBlkBytes. Use the variable _bufferSize as a constant for the size of the application buffer instead.


bsdcode:
I'm working a little bit more on the OSS backend. Therefore I marked this PR as a draft now.

bsdcode:
Renamed branch. Will open new PR when ready. Sorry for the noise.

PR #1595, opened 2024-08-07
Separate gameinput from hotkeys with hotkey prefix binding

bsdcode:
Currently there's no separation between gameinput bindings (e.g. button A) and hotkey bindings (e.g. Pause Emulation). This can be a problem especially for gamepad only users playing on the couch in front of the TV. After assigning all the buttons for the gameinput there are probably not many buttons on the gamepad left to assign freely to frequently used hotkey functions, e.g. Toggle Fullscreen, Save/Load State, Decrement/Increment State Slot, Pause Emulation, Reset System, Quit Emulator.

Introduce a new Prefix hotkey and separate the gameinput from the hotkeys.

This effectively provides two input modes: if the Prefix hotkey has any bindings, the mode is ON, otherwise it's OFF. The user gets informed in which mode the emulator is currently in when adding the first Prefix binding or when removing the last Prefix binding in the Hotkeys settings window. This serves to avoid confusion for the user if he assigns a binding to the Prefix hotkey by accident.

In the OFF state the emulator behaves exactly as before, i.e. gameinputs and hotkeys are executed without restrictions. In the ON state the gameinputs and hotkeys are executed depending on if the Prefix binding was pressed or not in advance. A message "Prefix ON/OFF" is displayed shortly in the statusbar whenever the Prefix hotkey is pressed, indicating if the prefix is active or not.

If the prefix is not active, only the gameinputs are executed, hotkeys aren't processed. If the prefix is active, only the hotkeys are executed, gameinputs aren't processed.

Example: Prefix hotkey is set to the "Home" button, Pause Emulation hotkey is set to the "A" button and Capture Screenshot is set to the "B" button on the gamepad. In order to pause the emulation, take a screenshot, unpause the emulation and then continue the gameplay the user inputs the following sequence:

Home button -> A -> B -> A -> Home button

### DIFF
--- a/desktop-ui/emulator/emulator.cpp
+++ b/desktop-ui/emulator/emulator.cpp
@@ -200,7 +200,9 @@ auto Emulator::input(ares::Node::Input::Input input) -> void {
 
   auto port = ares::Node::parent(device);
   if(!port) return;
-  
+
+  bool disabledInput = inputManager.hasPrefix && program.prefixed;
+
   for(auto& inputPort : ports) {
     if(inputPort.name != port->name()) continue;
     for(auto& inputDevice : inputPort.devices) {
@@ -208,10 +210,12 @@ auto Emulator::input(ares::Node::Input::Input input) -> void {
       for(auto& inputNode : inputDevice.inputs) {
         if(inputNode.name != input->name()) continue;
         if(auto button = input->cast<ares::Node::Input::Button>()) {
+          if(disabledInput) return;
           auto pressed = inputNode.mapping->pressed();
           return button->setValue(pressed);
         }
         if(auto axis = input->cast<ares::Node::Input::Axis>()) {
+          if(disabledInput) return;
           auto value = inputNode.mapping->value();
           return axis->setValue(value);
         }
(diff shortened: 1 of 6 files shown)


jcm93:
As useful as this is, I personally think it would be preferable to have a more standard implementation supporting multiple-key keybinds / keybinds with modifiers in general for all keys, rather than a specific, hard-coded single-key prefix option like this.

bsdcode:
The prefix hotkey has a different function than general multi-keybindings. More in the line of accessing emulator functions, e.g. accessing the menu with the PS button on Playstation while playing. Of course the scope for ares is different, ares doesn't have a home menu and doesn't make all emulator options available via hotkeys. Naming of the prefix hotkey and its location in the settings window could be changed/improved if desired.

But yes, I see what you mean. Making emulator actions available by multi-keybindings is similar, and multi-keybindings are probably a desired feature for all keybindings anyway. I totally understand if my PR is not accepted by the project in favor for a multi-keybinding implementation.

bsdcode:
I thought about it more. I still think that the concepts are different, but my analogy to a Home screen is not fitting. It's more in the line of a big picture mode. But I agree now, this prefix implementation is just not the correct way to handle either concepts.

PR #1598, opened 2024-08-11
n64: remove branch state machine from epilogue

invertego:
This will bloat generated code in the short term, but some of that can be elided with future optimizations.

Passes [n64-systemtest](https://github.com/lemmy-64/n64-systemtest) with both the interpreter and recompiler.


invertego:
Rebased on cleanup commit from @rasky.

PR #1604, opened 2024-08-16
Enforce newlines at end-of-file

jcm93:
Enforce newlines at end-of-file in the codebase.

While every modern compiler is able to deal with not having EOF newlines, it is still a good idea to enforce this.

ParaLLEl-RDP has a few files without EOF newlines, so I surrounded the n64 Vulkan `include` with a diagnostic pragma to ignore it. This avoids having to deal with compiler errors if ParaLLEl-RDP is updated, pending newlines added upstream.


jcm93:
Think I missed a couple; draft mode while I check.

edit: fixed.

PR #1605, opened 2024-08-16
nall/arithmetic: ignore -Winfinite-recursion warning in square()

jcm93:
More code cleanup for the new build system along with https://github.com/ares-emulator/ares/pull/1602, https://github.com/ares-emulator/ares/pull/1603, and https://github.com/ares-emulator/ares/pull/1604.

This one is fairly mysterious. clang thinks that all paths through this particular signature of the `square()` function call itself. This area of nall is a mass of templating and operator overloads, and figuring out exactly what's going on here would be a good one to unravel for a large-brained, C++-enjoying individual.

For now, just ignore this warning locally so that we can compile the rest of ares while still warning about infinite recursion elsewhere.

### DIFF
--- a/nall/arithmetic/natural.hpp
+++ b/nall/arithmetic/natural.hpp
@@ -167,6 +167,8 @@ alwaysinline auto bits(Pair value) -> u32 {
   }
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Winfinite-recursion"
 //Bits * Bits => Bits
 inline auto square(const Pair& lhs) -> Pair {
   static const Type Mask = (Type(0) - 1) >> HalfBits;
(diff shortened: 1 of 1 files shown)


Morilli:
This has been fixed locally in bizhawk for bsnes by applying a diff like:
```diff
diff --git a/nall/arithmetic/unsigned.hpp b/nall/arithmetic/unsigned.hpp
index 35e8a34b9..5bf89027a 100644
--- a/nall/arithmetic/unsigned.hpp
+++ b/nall/arithmetic/unsigned.hpp
@@ -18,7 +18,7 @@ inline auto mul(T lhs, U rhs) -> uintmax {
   return lhs * rhs;
 }
 
-template<typename T, enable_if_t<is_unsigned<T>::value>>
+template<typename T, typename = enable_if_t<is_unsigned<T>::value>>
 inline auto square(T value) -> uintmax {
   return value * value;
 }
```
[source](https://github.com/TASEmulators/BizHawk/blob/96b4de92af8a112c0b1dabfd8091ef160808f436/waterbox/bsnescore/bsnes/nall/arithmetic/unsigned.hpp)

I'm not too familiar with c++ templates, but this looks like the correct fix to me. In general it seems like not putting `typename = ` when the intent is clearly to constraint the type `T` is causing confusion.

PR #1619, opened 2024-08-22
build, n64: fix build

png183:
Removed `__VA_ARGS__` from line 6 of interpreter.cpp, since it appears to be unused and was causing compiler issues. Tested on Ubuntu with g++ and clang++ with both interpreter and recompiler running a default build of n64-systemtest.

Needs testing to make sure msvc still compiles correctly, and preferably also testing from someone more familiar with the N64 core in case I've overlooked something.

### DIFF
--- a/ares/n64/cpu/interpreter.cpp
+++ b/ares/n64/cpu/interpreter.cpp
@@ -3,7 +3,7 @@
 #define RT ipu.r[RTn]
 #define RS ipu.r[RSn]
 
-#define jp(id, name, ...) case id: return decoder##name(instruction)
+#define jp(id, name)      case id: return decoder##name(instruction)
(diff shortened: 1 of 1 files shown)


invertego:
Sorry, my mistake. I did successfully build that PR with clang, but it alters the preprocessor behavior for *-msvc targets.

Your fix should be fine across all compilers for the purposes of unbreaking the build.

I think it would be even better to:
A) Drop the "..." entirely from the macro parameters so if someone later tries to pass any, they aren't silently discarded.
or
B) Replace `__VA_ARGS__` with `##__VA_ARGS__`. This was a part of rasky's original commit that I accidentally reverted.

...but you should feel no obligation to spend more time on this. If it goes in as-is, I can follow up on it later.

PR #1635, opened 2024-09-02
n64: refactor devirtualize and memory ops, force JIT only with instruction cache

rasky:
This change contains a large refactoring with no functional changes, plus a small functional change that will help speeding up the JIT.

The refactoring allows CPU::devirtualize to return all the information regarding the memory access. This in turn allows to simplify all memory ops (which contained lots of duplicated code) to use devirtualize, reducing code duplication.

The functional change is that we now use JIT only when running code from icache (which is 99.9999% of the times). Running code without icache is extremely slow on real hardware and only happens in specific situations (eg: during boot when RDRAM is not initialized). By limiting the JIT to run from icache, we open the door to implement proper icache support in the JIT and finish remove the instruction epilogue with its slow "instruction cache stepper" (which in addition to being slow, is also inaccurate and make us fail cache test ROMs).

### DIFF
--- a/ares/n64/cpu/cpu.hpp
+++ b/ares/n64/cpu/cpu.hpp
@@ -110,29 +110,29 @@ struct CPU : Thread {
   struct InstructionCache {
     CPU& self;
     struct Line;
-    auto line(u32 vaddr) -> Line& { return lines[vaddr >> 5 & 0x1ff]; }
+    auto line(u64 vaddr) -> Line& { return lines[vaddr >> 5 & 0x1ff]; }
 
     //used by the recompiler to simulate instruction cache fetch timing
-    auto step(u32 vaddr, u32 address) -> void {
+    auto step(u64 vaddr, u32 paddr) -> void {
       auto& line = this->line(vaddr);
-      if(!line.hit(address)) {
+      if(!line.hit(paddr)) {
         self.step(48 * 2);
         line.valid = 1;
-        line.tag   = address & ~0x0000'0fff;
+        line.tag   = paddr & ~0x0000'0fff;
       } else {
         self.step(1 * 2);
       }
     }
 
     //used by the interpreter to fully emulate the instruction cache
-    auto fetch(u32 vaddr, u32 address, CPU& cpu) -> u32 {
+    auto fetch(u64 vaddr, u32 paddr, CPU& cpu) -> u32 {
       auto& line = this->line(vaddr);
-      if(!line.hit(address)) {
-        line.fill(address, cpu);
+      if(!line.hit(paddr)) {
+        line.fill(paddr, cpu);
       } else {
         cpu.step(1 * 2);
       }
-      return line.read(address);
+      return line.read(paddr);
     }
 
     auto power(bool reset) -> void {
@@ -147,11 +147,11 @@ struct CPU : Thread {
 
     //16KB
     struct Line {
-      auto hit(u32 address) const -> bool { return valid && tag == (address & ~0x0000'0fff); }
-      auto fill(u32 address, CPU& cpu) -> void {
+      auto hit(u32 paddr) const -> bool { return valid && tag == (paddr & ~0x0000'0fff); }
+      auto fill(u32 paddr, CPU& cpu) -> void {
         cpu.step(48 * 2);
         valid = 1;
-        tag   = address & ~0x0000'0fff;
+        tag   = paddr & ~0x0000'0fff;
         cpu.busReadBurst<ICache>(tag | index, words);
       }
 
@@ -160,7 +160,7 @@ struct CPU : Thread {
         cpu.busWriteBurst<ICache>(tag | index, words);
       }
 
-      auto read(u32 address) const -> u32 { return words[address >> 2 & 7]; }
+      auto read(u32 paddr) const -> u32 { return words[paddr >> 2 & 7]; }
 
       bool valid;
       u32  tag;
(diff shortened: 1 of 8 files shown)


ares/n64/cpu/cpu.hpp
invertego, line 1: Is passing the 64-bit vaddr a functional change or is it for consistency? At a glance it doesn't seem like the upper 32 bits are being used here.

On another note, I like the improved clarity of variable names that disambiguate virtual and physical addresses.

ares/n64/cpu/memory.cpp
```diff
@@ -88,23 +88,27 @@ auto CPU::segment(u64 vaddr) -> Context::Segment {
...
-    addressException(vaddr);
-    exception.addressLoad();
-    return nothing;
+    if(raiseExceptions) {
+      addressException(vaddr);
+      exception.addressLoad();
+    }
+    return PhysAccess{false};
   case Context::Segment::Mapped:
-    if(auto match = tlb.load(vaddr)) return match.address & context.physMask;
```
invertego, line 99: The masking by context.physMask disappeared. Was it needed in the first place, or has the need gone away?

ares/n64/cpu/interpreter-ipu.cpp
```diff
@@ -118,8 +118,9 @@ auto CPU::BREAK() -> void {
 }
 
 auto CPU::CACHE(u8 operation, cr64& rs, s16 imm) -> void {
-  auto access = devirtualize(rs.u64 + imm);
+  auto access = devirtualize<Read>(rs.u64 + imm);
   if (!access) return;
```
invertego, line 122: This can skip the vaddrAlignedError check, which seems like a change in behavior. Is this correct?

Similar questions apply to LL and LLD.

ares/n64/cpu/cpu.hpp
```diff
@@ -277,17 +279,24 @@ struct CPU : Thread {
...
-  auto fetch(u64 vaddr) -> maybe<u32>;
+  auto fetch(PhysAccess access) -> maybe<u32>;
   template<u32 Size> auto busWrite(u32 address, u64 data) -> void;
   template<u32 Size> auto busRead(u32 address) -> u64;
   template<u32 Size> auto busWriteBurst(u32 address, u32 *data) -> void;
   template<u32 Size> auto busReadBurst(u32 address, u32 *data) -> void;
-  template<u32 Size> auto read(u64 vaddr) -> maybe<u64>;
-  template<u32 Size> auto write(u64 vaddr, u64 data, bool alignedError=true) -> bool;
+  template<u32 Size> auto read(PhysAccess access) -> maybe<u64>;
+  template<u32 Size> auto write(PhysAccess access, u64 data, bool alignedError=true) -> bool;
```
invertego: Nit: spaces around '=' here and a few lines below

ares/n64/cpu/memory.cpp
```diff
@@ -168,132 +169,40 @@ inline auto CPU::busReadBurst(u32 address, u32 *data) -> void {
...
-    step(1 * 2);
-    return busRead<Word>(vaddr & 0x1fff'ffff);
-  case Context::Segment::Direct32:
-    step(1 * 2);
-    return busRead<Word>(vaddr & 0xffff'ffff);
-  }
-
-  unreachable;
+auto CPU::fetch(PhysAccess access) -> maybe<u32> {
+  step(1 * 2);
```
invertego, line 181: This doesn't appear to preserve the original stepping behavior. The original code deferred to InstructionCache::fetch to step the appropriate amount (in cases where it is called). Now we will end up with an extra step.

ares/n64/cpu/memory.cpp
```diff
@@ -168,132 +169,40 @@ inline auto CPU::busReadBurst(u32 address, u32 *data) -> void {
...
-  case Context::Segment::Direct:
-    step(1 * 2);
-    return busRead<Size>(vaddr & 0x1fff'ffff);
-  case Context::Segment::Direct32:
-    step(1 * 2);
-    return busRead<Size>(vaddr & 0xffff'ffff);
-  }
-
-  unreachable;
+auto CPU::read(PhysAccess access) -> maybe<u64> {
```
invertego, line 188: It looks like we lost the step() calls entirely here. Same applies to CPU::write.

invertego (changes requested):
I have just a few questions/concerns that stem from places where appears behavior has changed, as noted in my comments.

ares/n64/cpu/cpu.hpp
rasky: It's just for consistency. I'd like to standardize on using `u64` for all virtual addresses to avoid confusion, uniform logging, etc.

ares/n64/cpu/memory.cpp
rasky: I think it was not needed. It was written by Near years ago, I think they originally misunderstood how 32-bit vs 64-bit TLB addressing work. I took the chance to remove them and n64-systemtest still passes. 

context.physMask is only used in devirtualizeFast now which I'm not touching as I'd like to remove it soon. After that, we can remove the variable.

ares/n64/cpu/interpreter-ipu.cpp
rasky: CPU::CACHE calls vaddrAlignedError explicitly just after.
LL/LLD calls CPU::read that also calls vaddrAlignedError.

ares/n64/cpu/cpu.hpp
rasky: ok

ares/n64/cpu/memory.cpp
rasky: True, I'll remove this then.

ares/n64/cpu/memory.cpp
rasky: I purposedly removed this. I don't see how it can be correct. In addition to the standard stepping in fetch, a memory load or write per-se do not cause an extra cycle of stall. Reads and writes have two cycle latency, and that can cause a stall in case of register conflicts, but forcing 1 cycle stall for all reads and writes is surely wrong.

Obviously this means that with this PR, our CPU will run "slower" than before, which in turn might affect framerate a bit, but that's life I guess.

ares/n64/cpu/interpreter-ipu.cpp
invertego: Yes, but it can be skipped by the early return in CACHE, and the read call in LL/LLD is also inside a conditional. Before this PR, the alignment check ran at the top of devirtualize() before anything else.

ares/n64/cpu/memory.cpp
invertego: All right. We should call out that this change affects timing, and it will be no surprise if we see an impact in the usual suspects (e.g. DK64).

ares/n64/cpu/memory.cpp
invertego: Removing it entirely means the non-cached cases that used to have a step now have none. Does the rationale in your reply on the CPU::read() comment thread (justifying the removal) apply here as well?

rasky: Unfortunately this becomes a giant snowball pretty fast. If you can suggest a simple way to maintain the current timing behavior in the PR I can do that, otherwise we will have to accept small changes. The timing is wrong anyway. The concept that the cache has an impact on timing is wrong. Timing on the CPU is either stalls or the bus being held during memory accesses (and depending on what access), cache has nothing to do with it. I tried fixing it more but I end up needing to implement more correct timings across the board, and I don't have numbers for those, and this in turn has surprising behaviors on random games.

rasky: I pushed a commit that should get us closer to before, let me know if it works for you

invertego: Yeah, that seems less disruptive.

ares/n64/cpu/interpreter-ipu.cpp
rasky: Thanks, this is indeed a bug not covered by n64-systemtest. I've verified on hardware that address errors come always before TLB errors (and [added a test](https://github.com/lemmy-64/n64-systemtest/pull/80)).

To fix this, I've had to add a new template parameter to multiplex `devirtualize()` also by access size (as aligned errors depend on the access size). I've also had to reinstate the hack for which CPU writes can request not to emit align errors (used by SWR -- and should be used by SDR -- while we don't properly support masked writes in the SysAD bus implementation).

invertego: Thanks for digging into it!

invertego (approved):

PR #1637, opened 2024-09-05
Added Pencil II support.

cdoty:
Created a Pencil II core based on the ColecoVision core. The timings are different as the system uses a 10.738 mhz crystal instead of the 7 mhz on the Colecovision.

### DIFF
--- a/ares/GNUmakefile
+++ b/ares/GNUmakefile
@@ -73,6 +73,10 @@ ifneq ($(filter $(cores),cv),)
   include $(ares.path)/cv/GNUmakefile
 endif
 
+ifneq ($(filter $(cores),pencil2),)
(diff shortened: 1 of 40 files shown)


LukeUsher:
Given how similar these platforms are, I can't help but think it would it be better to make it an alternate configuration of the ColecoVision core rather than an entirely new one.

Take a look at how Pocket Challenge V2 is an alternate configuration in the WonderSwan core,

Thoughts/feedback welcome

Also this seems to lack support for the keyboard? Given that the Pencil II is a computer, don't you think that would be important to add?

cdoty:
That would be better. Let me have a look and I'll see if I can adapt it.

LukeUsher:
Did you manage to take a look at this (merged cores)?

PR #1639, opened 2024-09-06
Allow arbitrary and possibly unaligned writes to RDRAM through gdb

cemulate:
We want to allow arbitrary writes through the GDB server to the RDRAM area. This was done in #1620, but that version has a subtle issue where the server will still refuse to write or change writes that happen to be of size `Byte|Half|Word|Dual` if the addressed is not aligned to that size.
This PR fixes the issue by checking if the write is to RDRAM first, allowing arbitrary and possibly unaligned writes. Otherwise, the write is handled through the normal machinery where these issues may be semantically important.

### DIFF
--- a/ares/n64/system/system.cpp
+++ b/ares/n64/system/system.cpp
@@ -174,20 +174,20 @@ auto System::initDebugHooks() -> void {
       switch(data.size()) {
         case Byte: 
           value = (u64)data[0];
-          cpu.write<Byte>(address, value, false);
+          cpu.writeDebug<Byte>(address, value);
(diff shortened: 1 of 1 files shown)


LukeUsher:
Looks like the build is failing here for windows (x64 and arm64) and macOS.

cemulate:
Missed an actual change (`CPU::write` -> `CPU::writeDebug`) during the rebase since large portions were shifted; for some reason my local (linux) build still succeeded. This should fix the rest (I don't have easy access to those platforms to test).

PR #1640, opened 2024-09-08
Add N64 recompiler block hashes & inline 64-bit ops

ghost:
Reworking recompiler's JIT block accesses to be more flexible so that we can bake more runtime info to them. This way runtime checks can be avoided in the generated code, making inlining the implementations more attractive since we don't need to generate calls to e.g. kernel mode checks for every 64-bit op. This is expected to increase recompiled code's performance eventually.

I expect these changes to have a small negative performance impact but haven't measured it.

Commit messages:

>We'd like the recompiler to take the execution context such as kernel
    mode into account when compiling blocks. That's why it's necessary to
    identify blocks not just by address but all the information used at
    compile time. This is done by computing a 32-bit key and using that as
    a block's identifier instead of the last six physical address bits like
    was done before.
    
  >  Since we have now 32-bit instead of 6-bit keys, the block() function
    hashes the keys before mapping them to one of the 64 pool rows. The hash
    function was chosen arbitrarily to be better than a simple multiplicative
    hash and is likely not the best choice for this exact task.
>* Pass JITContext down to leaf emit functions.
>* Emit inline implementations of basic 64-bit operations.
> * Use block compile-time information to elide kernel mode checks of
      the now inlined operations.

### DIFF
--- a/ares/n64/cpu/context.cpp
+++ b/ares/n64/cpu/context.cpp
@@ -18,6 +18,8 @@ auto CPU::Context::setMode() -> void {
     break;
   }
 
+  self.recompiler.jitContext.update(self);
+
   if(bits == 32) {
     physMask = 0x1fff'ffff;
     segment[0] = Segment::Mapped;
--- a/ares/n64/cpu/cpu.cpp
+++ b/ares/n64/cpu/cpu.cpp
@@ -112,7 +112,7 @@ auto CPU::instruction() -> void {
 
   if(Accuracy::CPU::Recompiler && recompiler.enabled && access.cache) {
     if(vaddrAlignedError<Word>(access.vaddr, false)) return;
-    auto block = recompiler.block(ipu.pc, access.paddr, GDB::server.hasBreakpoints());
+    auto block = recompiler.block(ipu.pc, access.paddr);
     if(block) {
       block->execute(*this);
       return;
--- a/ares/n64/cpu/cpu.hpp
+++ b/ares/n64/cpu/cpu.hpp
@@ -883,13 +883,34 @@ struct CPU : Thread {
       u8* code;
     };
 
+    struct PoolRow {
+      Block* block;
+      u32 stateBits;
+    };
+
     struct Pool {
-      Block* blocks[1 << 6];
+      PoolRow* rows[1 << 6];
+    };
+
+    struct JITContext {
+      bool singleInstruction;
+      Context::Endian endian;
+      Context::Mode mode;
+      bool cop1Enabled;
+      bool floatingPointMode;
+      bool is64bit;
+
+      u32  stateBits; //the above state, compressed
+
+      auto update(const CPU& cpu) -> void;
+      auto toBits() const -> u32;
     };
 
     auto reset() -> void {
       pools.resize(1 << 21);  //2_MiB * sizeof(void*) == 16_MiB
       std::ranges::fill(pools, nullptr);
+      blockCache.clear();
+      poolBlockKeys.clear();
     }
 
     auto invalidate(u32 address) -> void {
@@ -905,13 +926,24 @@ struct CPU : Thread {
       auto pool = pools[address >> 8 & 0x1fffff];
       if(!pool) return;
       memory::jitprotect(false);
-      pool->blocks[address >> 2 & 0x3f] = nullptr;
+      pool->rows[address >> 2 & 0x3f] = nullptr;
       memory::jitprotect(true);
       #endif
     }
 
     auto invalidatePool(u32 address) -> void {
-      pools[address >> 8 & 0x1fffff] = nullptr;
+      u32 pool = address >> 8 & 0x1fffff;
+
+      if (pools[pool]) {
+        auto range = poolBlockKeys.equal_range(pool);
+        for (auto it = range.first; it != range.second; ++it) {
+          u64 cacheKey = it->second;
+          blockCache.erase(cacheKey);
+        }
+      }
+
+      pools[pool] = nullptr;
+
     }
 
     auto invalidateRange(u32 address, u32 length) -> void {
@@ -921,10 +953,13 @@ struct CPU : Thread {
     }
 
     auto pool(u32 address) -> Pool*;
-    auto block(u64 vaddr, u32 address, bool singleInstruction = false) -> Block*;
+    auto blockCacheKey(u32 address, u32 state) -> u64;
+    auto block(u64 vaddr, u32 address) -> Block*;
 
-    auto emit(u64 vaddr, u32 address, bool singleInstruction = false) -> Block*;
+    auto emit(u64 vaddr, u32 address) -> Block*;
+    auto emitOverflowCheck() -> sljit_jump*;
     auto emitZeroClear(u32 n) -> void;
+    auto checkDualAllowed() -> bool;
     auto emitEXECUTE(u32 instruction) -> bool;
     auto emitSPECIAL(u32 instruction) -> bool;
     auto emitREGIMM(u32 instruction) -> bool;
@@ -936,7 +971,18 @@ struct CPU : Thread {
     bool callInstructionPrologue = false;
     bump_allocator allocator;
     std::vector<Pool*> pools;
+    std::unordered_map<u64, Block*> blockCache;
+    std::unordered_multimap<u32, u64> poolBlockKeys;
+    JITContext jitContext;
   } recompiler{*this};
+
+  auto arithmeticOverflowException() -> void {
+    exception.arithmeticOverflow();
+  }
+
+  auto reservedInstructionException() -> void {
+    exception.reservedInstruction();
+  }
 
   struct Disassembler {
     CPU& self;
--- a/ares/n64/cpu/recompiler.cpp
+++ b/ares/n64/cpu/recompiler.cpp
@@ -1,3 +1,4 @@
+
 auto CPU::Recompiler::pool(u32 address) -> Pool* {
   auto& pool = pools[address >> 8 & 0x1fffff];
   if(!pool) {
@@ -9,14 +10,69 @@ auto CPU::Recompiler::pool(u32 address) -> Pool* {
   return pool;
 }
 
-auto CPU::Recompiler::block(u64 vaddr, u32 address, bool singleInstruction) -> Block* {
-  if(auto block = pool(address)->blocks[address >> 2 & 0x3f]) return block;
-  auto block = emit(vaddr, address, singleInstruction);
-  if(block) {
-    pool(address)->blocks[address >> 2 & 0x3f] = block;
+auto CPU::Recompiler::blockCacheKey(u32 address, u32 state) -> u64 {
+  return u64(address) | (u64(state) << 32);
+}
+
+auto CPU::Recompiler::block(u64 vaddr, u32 address) -> Block* {
+  u32 state = jitContext.stateBits;
+  u32 idx = (address >> 2) & 0x3f;
+  assert(idx < sizeof(Pool::rows)/sizeof(Pool::rows[0]));
+
+  {
+    Pool* p = pool(address);
+    if (auto row = p->rows[idx]) {
+      if (row->stateBits == state) {
+        assert(row->block);
+        return row->block;
+      } else {
+        auto blockIt = blockCache.find(blockCacheKey(address, state));
+        if (blockIt != blockCache.end()) {
+          row->block = blockIt->second;
+          row->stateBits = state;
+          return row->block;
+        }
+      }
+    }
+  }
+
+  auto block = emit(vaddr, address);
+  if (block) {
+    Pool* p = pool(address);
+    auto row = (PoolRow*)allocator.acquire(sizeof(PoolRow));
+    row->block = block;
+    row->stateBits = state;
+    p->rows[idx] = row;
+
+    u32 poolIdx = address >> 8 & 0x1fffff;
+    u64 key = blockCacheKey(address, state);
+    blockCache[key] = block;
+    poolBlockKeys.insert({poolIdx, key});
     memory::jitprotect(true);
   }
   return block;
+}
+
+auto CPU::Recompiler::JITContext::update(const CPU& cpu) -> void {
+  singleInstruction = GDB::server.hasBreakpoints();
+  endian = Context::Endian(cpu.context.endian);
+  mode = Context::Mode(cpu.context.mode);
+  cop1Enabled = cpu.scc.status.enable.coprocessor1 > 0;
+  floatingPointMode = cpu.scc.status.floatingPointMode > 0;
+  is64bit = cpu.context.bits == 64;
+
+  stateBits = toBits();
+}
+
+auto CPU::Recompiler::JITContext::toBits() const -> u32 {
+  u32 bits = 1; // first bit always set to make 0 invalid
+  bits |= singleInstruction ? 1 << 1 : 0;
+  bits |= endian ? 1 << 2 : 0;
+  bits |= (mode & 0x03) << 3;
+  bits |= cop1Enabled ? 1 << 5 : 0;
+  // bits |= floatingPointMode ? 1 << 6 : 0;
+  bits |= is64bit ? 1 << 7 : 0;
+  return bits;
 }
 
 #define IpuBase        offsetof(IPU, r[16])
@@ -27,7 +83,7 @@ auto CPU::Recompiler::block(u64 vaddr, u32 address, bool singleInstruction) -> B
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Winvalid-offsetof"
 #endif
-auto CPU::Recompiler::emit(u64 vaddr, u32 address, bool singleInstruction) -> Block* {
+auto CPU::Recompiler::emit(u64 vaddr, u32 address) -> Block* {
   if(unlikely(allocator.available() < 1_MiB)) {
     print("CPU allocator flush\n");
     allocator.release();
@@ -39,6 +95,7 @@ auto CPU::Recompiler::emit(u64 vaddr, u32 address, bool singleInstruction) -> Bl
     return nullptr;
 
   bool abort = false;
+
   beginFunction(3);
 
   Thread thread;
@@ -78,7 +135,7 @@ auto CPU::Recompiler::emit(u64 vaddr, u32 address, bool singleInstruction) -> Bl
     vaddr += 4;
     address += 4;
     jumpToSelf += 4;
-    if(hasBranched || (address & 0xfc) == 0 || singleInstruction) break;  //block boundary
+    if(hasBranched || (address & 0xfc) == 0 || jitContext.singleInstruction) break;  //block boundary
     hasBranched = branched;
     jumpEpilog(flag_nz);
   }
@@ -89,7 +146,7 @@ auto CPU::Recompiler::emit(u64 vaddr, u32 address, bool singleInstruction) -> Bl
   auto block = (Block*)allocator.acquire(sizeof(Block));
   block->code = endFunction();
 
-//print(hex(PC, 8L), " ", instructions, " ", size(), "\n");
+  //print(hex(PC, 8L), " ", instructions, " ", size(), "\n");
   return block;
 }
 #if defined(COMPILER_CLANG) || defined(COMPILER_GCC)
@@ -124,6 +181,24 @@ auto CPU::Recompiler::emit(u64 vaddr, u32 address, bool singleInstruction) -> Bl
 
 auto CPU::Recompiler::emitZeroClear(u32 n) -> void {
   if(n == 0) mov64(mem(IpuReg(r[0])), imm(0));
+}
+
+auto CPU::Recompiler::emitOverflowCheck() -> sljit_jump* {
+    // If overflow flag set: throw an exception, skip the instruction via the 'end' label.
+    auto didntOverflow = jump(flag_no);
+    call(&CPU::arithmeticOverflowException);
+    auto end = jump();
+    setLabel(didntOverflow);
+    return end;
+}
+
+auto CPU::Recompiler::checkDualAllowed() -> bool {
+  if (jitContext.mode != Context::Mode::Kernel && !jitContext.is64bit) {
+    call(&CPU::reservedInstructionException);
+    return false;
+  }
+
+  return true;
 }
 
 auto CPU::Recompiler::emitEXECUTE(u32 instruction) -> bool {
@@ -282,15 +357,21 @@ auto CPU::Recompiler::emitEXECUTE(u32 instruction) -> bool {
 
   //DADDI Rt,Rs,i16
   case 0x18: {
-    callf(&CPU::DADDI, mem(Rt), mem(Rs), imm(i16));
-    emitZeroClear(Rtn);
+    if (!checkDualAllowed()) return 1;
+    add64(reg(0), mem(Rs), imm(i16), set_o);
+    auto skip = emitOverflowCheck();
+    if(Rtn > 0) mov64(mem(Rt), reg(0));
+    setLabel(skip);
     return 0;
   }
 
   //DADDIU Rt,Rs,i16
   case 0x19: {
-    callf(&CPU::DADDIU, mem(Rt), mem(Rs), imm(i16));
-    emitZeroClear(Rtn);
+    if (!checkDualAllowed()) return 1;
+    if(Rtn > 0) {
+      add64(reg(0), mem(Rs), imm(i16));
+      mov64(mem(Rt), reg(0));
+    }
     return 0;
   }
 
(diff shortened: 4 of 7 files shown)


ares/n64/cpu/cpu.cpp
```diff
@@ -113,7 +113,15 @@ auto CPU::instruction() -> void {
...
-    auto block = recompiler.block(ipu.pc, access.paddr, GDB::server.hasBreakpoints());
+    auto block = recompiler.block(ipu.pc, access.paddr,
+        {
+        .singleInstruction = GDB::server.hasBreakpoints(),
+        .endian = Context::Endian(context.endian),
+        .mode = Context::Mode(context.mode),
+        .cop1Enabled = scc.status.enable.coprocessor1 > 0,
+        .floatingPointMode = scc.status.floatingPointMode > 0,
+        .is64bit = context.bits == 64,
+        });
```
rasky: I'm a bit concerned that we compose the context structure every block we run. I believe it might be better if we do this on changes instead. I understand it's harder to get right and can bring more bugs, but I believe it's going to be faster.

So let's say, have a function that recalculates the current context and its hash key and store them somewhere. At runtime, we just use the precalulcated context and the precalculated hash key.

Then, you need to call the function to recalculate the context in any codepath that can change one of the variables that affect it, os for instance `mtc0` of the status register is a prime suspect, and there will be others of course, but maybe not dozen of them.

ghost: Thank you for the quick review. I moved JITContext inside the existing Context struct, now it's `Context::JIT` and its contents + bit vector representation are computed only when `Context::setMode()` is called. This is more often than strictly needed, for example exceptions change the **mcc** vector state and call `setMode()` but should be better than before.

I added separate `update` and `toBits` member functions to `Context::JIT` for later debugging; they can be used to check for staleness bugs in debug builds.

ghost:
The `GDB::server.hasBreakpoints()` check is a bit problematic because right now it gets updated only when `Context::setMode` is called even though breakpoints might be added or removed at other times as well. I suppose it could be polled for every block perhaps? I considered it part of the execution context but before it wasn't taken into account at all when looking up blocks. To me it seems like breakpoints only worked after all Pool instances where flushed.

invertego:
Sorry for the delay. I will try to review this in the next few days.

ares/n64/cpu/recompiler.cpp
```diff
@@ -9,10 +9,34 @@ auto CPU::Recompiler::pool(u32 address) -> Pool* {
...
 
-auto CPU::Recompiler::block(u64 vaddr, u32 address, bool singleInstruction) -> Block* {
-  if(auto block = pool(address)->blocks[address >> 2 & 0x3f]) return block;
-  auto block = emit(vaddr, address, singleInstruction);
-  pool(address)->blocks[address >> 2 & 0x3f] = block;
+auto CPU::Recompiler::computePoolKey(u32 address, u32 jitBits) -> u32 {
+  return (address >> 2 & 0x3f) | (jitBits & ~0x3f);
+}
+
+auto CPU::Recompiler::computePoolRow(u32 key) -> u32 {
```
invertego: The n64 core already pulls in xxhash.h, so XXH32_avalanche might work here as well.

ares/n64/cpu/context.cpp
```diff
@@ -18,6 +18,9 @@ auto CPU::Context::setMode() -> void {
     break;
   }
 
+  jit.update(*this, self);
+  jitBits = jit.toBits();
```
invertego: Should update() be responsible for recalculating jitBits?

ares/n64/cpu/context.cpp
```diff
@@ -18,6 +18,9 @@ auto CPU::Context::setMode() -> void {
     break;
   }
 
+  jit.update(*this, self);
```
invertego: We also need to make sure this is updated after a save state load.

invertego: Also when the breakpoint count changes.

ares/n64/cpu/cpu.hpp
```diff
@@ -106,6 +118,8 @@ struct CPU : Thread {
     u32  mode;
     u32  bits;
     u32  segment[8];  //512_MiB chunks
+    u32  jitBits;
```
invertego: The rest of the recompiler state is in the recompiler object. Is there a reason you want to keep this state separate?

ares/n64/cpu/recompiler.cpp
```diff
@@ -950,41 +988,42 @@ auto CPU::Recompiler::emitSPECIAL(u32 instruction) -> bool {
...
 
   //DSUBU Rd,Rs,Rt
   case 0x2f: {
-    lea(reg(1), Rd);
-    lea(reg(2), Rs);
-    lea(reg(3), Rt);
-    call(&CPU::DSUBU);
-    emitZeroClear(Rdn);
+    if (!checkDualAllowed(ctx)) return 1;
+    sub64(reg(0), mem(Rs), mem(Rt), set_o);
```
invertego: set_o not needed

ares/n64/cpu/recompiler.cpp
```diff
@@ -284,21 +327,19 @@ auto CPU::Recompiler::emitEXECUTE(u32 instruction) -> bool {
...
 
   //DADDIU Rt,Rs,i16
   case 0x19: {
-    lea(reg(1), Rt);
-    lea(reg(2), Rs);
-    mov32(reg(3), imm(i16));
-    call(&CPU::DADDIU);
-    emitZeroClear(Rtn);
+    if (!checkDualAllowed(ctx)) return 1;
+    add64(reg(0), mem(Rs), imm(i16), set_o);
```
invertego: set_o not needed

ares/n64/cpu/recompiler.cpp
```diff
@@ -103,12 +127,31 @@ auto CPU::Recompiler::emitZeroClear(u32 n) -> void {
   if(n == 0) mov64(mem(IpuReg(r[0])), imm(0));
 }
 
-auto CPU::Recompiler::emitEXECUTE(u32 instruction) -> bool {
+auto CPU::Recompiler::emitOverflowCheck(reg temp) -> sljit_jump* {
+    // If overflow flag set: throw an exception, skip the instruction via the 'end' label.
+    mov32_f(temp, flag_o);
+    auto didntOverflow = cmp32_jump(temp, imm(0), flag_eq);
```
invertego:
```suggestion
    auto didntOverflow = jump(flag_no);
```

ares/n64/cpu/recompiler.cpp
```diff
@@ -103,12 +127,31 @@ auto CPU::Recompiler::emitZeroClear(u32 n) -> void {
...
+    call(&CPU::Exception::arithmeticOverflow, &cpu.exception);
```
invertego: Emitting calls this way causes all the parameters to be emitted as immediates. In general it's cheaper (in terms of code footprint) to calculate addresses that are passed as arguments (see instances of `lea` in this file). Although, for cold code paths like this, it would be even better to call a trampoline in CPU that takes no arguments (aside from the implicit this pointer).

ares/n64/cpu/recompiler.cpp
```diff
@@ -284,21 +327,19 @@ auto CPU::Recompiler::emitEXECUTE(u32 instruction) -> bool {
...
   //DADDIU Rt,Rs,i16
   case 0x19: {
-    lea(reg(1), Rt);
-    lea(reg(2), Rs);
-    mov32(reg(3), imm(i16));
-    call(&CPU::DADDIU);
-    emitZeroClear(Rtn);
+    if (!checkDualAllowed(ctx)) return 1;
+    add64(reg(0), mem(Rs), imm(i16), set_o);
+    if(Rtn > 0) mov64(mem(Rt), reg(0));
```
invertego: The add64 can also be skipped if Rtn is zero.

ares/n64/cpu/recompiler.cpp
```diff
@@ -950,41 +988,42 @@ auto CPU::Recompiler::emitSPECIAL(u32 instruction) -> bool {
...
   //DSUBU Rd,Rs,Rt
   case 0x2f: {
-    lea(reg(1), Rd);
-    lea(reg(2), Rs);
-    lea(reg(3), Rt);
-    call(&CPU::DSUBU);
-    emitZeroClear(Rdn);
+    if (!checkDualAllowed(ctx)) return 1;
+    sub64(reg(0), mem(Rs), mem(Rt), set_o);
+    if(Rdn > 0) mov64(mem(Rd), reg(0));
```
invertego: The sub64 can be skipped in Rdn is zero.

ares/n64/cpu/cpu.hpp
```diff
@@ -863,7 +877,11 @@ struct CPU : Thread {
     };
 
     struct Pool {
-      Block* blocks[1 << 6];
+      struct Row {
+        Block* block;
+        u32 tag;
+      };
+      Row rows[1 << 6];
```
invertego: How did you arrive at 64 rows? Were other numbers tried?

invertego (changes requested):
Looks good aside from the issues mentioned. How much more are you planning to do with this before taking it out of draft status?

ghost:
Noting here that I've transferred to a new GitHub account: @pekkavaa 
I'd like to continue ares JIT performance work at some point but maybe with a solid benchmarking setup first. Even small changes should get reliably measured.

**Edit:** And thanks for the review, invertego!

ares/n64/cpu/recompiler.cpp
pekkavaa: That's a good point. I added a trampoline instead:
```
  auto arithmeticOverflowException() -> void {
    exception.arithmeticOverflow();
  }
  ```

File: ares/n64/cpu/cpu.hpp (hunk 12 above)
Comment (pekkavaa) [reply ares/n64/cpu/cpu.hpp]: It's the same number of blocks the old Pool had:
```c++
struct Pool {
      Block* blocks[1 << 6];
    };
```
I didn't try any other sizes yet. Seems like it's too small though since I see a `CPU allocator flush` every few seconds when idling in SM64.

ares/n64/cpu/recompiler.cpp
pekkavaa: Thanks this made it a lot cleaner.

ares/n64/cpu/cpu.hpp
pekkavaa: That's an excellent point. I moved it all under Recompiler. The recompiler's code itself became neater because there's no need to pass the Context object around. Calling the `update` method is somewhat awkward now but I think it's passable. From serialization.cpp: `recompiler.jitContext.update(context, *this);`

ares/n64/cpu/context.cpp
pekkavaa: To address the save state load, I added this to serialization.cpp:
```c++
  if constexpr(Accuracy::CPU::Recompiler) {
    recompiler.reset();
    recompiler.jitContext.update(context, *this);
  }
```
I hope it's correct.

For breakpoints, the JIT context update is done on `emuCacheInvalidate` hook:
```c++
  if constexpr(Accuracy::CPU::Recompiler) {
    GDB::server.hooks.emuCacheInvalidate = [](u64 address) {
      cpu.recompiler.invalidate(address);
      cpu.recompiler.jitContext.update(cpu.context, cpu); // the number of breakpoints may have changed
    };
  }
}
```

ares/n64/cpu/context.cpp
pekkavaa: Yeah I moved it inside update()

ares/n64/cpu/recompiler.cpp
pekkavaa: Excellent. I'm using XXH32_avalanche instead.

pekkavaa:
Hello again :wave: I rebased the branch and addressed all the review comments. I also squashed the two commits together since it got annoying to move changes in between them. Things seem to work still and it passes N64 systemtest. The test ROM looks different now than what I recall though:
image

> How much more are you planning to do with this before taking it out of draft status?

I'd like to get an idea on the performance impact and then adjust at least the block pool size accordingly. Any ideas how to benchmark ares are welcome :)

ares/n64/cpu/recompiler.cpp
```diff
@@ -9,23 +9,65 @@ auto CPU::Recompiler::pool(u32 address) -> Pool* {
...
+  u32 key = computePoolKey(address, jitContext.stateBits);
+  u32 row = computePoolRow(key);
+
+  if (pool(address)->rows[row].tag == key) {
+    if (auto block = pool(address)->rows[row].block) {
+      return block;
+    }
+  }
+
+  // TODO need to call jitprotect(false) here?
```
invertego: Nope, emit() will unprotect the code cache (make it writable) if needed.

ares/n64/cpu/cpu.hpp
invertego: Well, originally it was directly indexing the instructions in the block. Now that it's indexing by hash, collisions are an issue. Two potential solutions:
1. Increase the row count. This will reduce the likelihood of collisions at the expense of memory (and data cache efficiency).
2. In the event of a mismatch, check if the block was previously compiled and if so, reuse it. The RSP and SH2 recompilers do this, so you can check those out for inspiration.

ares/n64/cpu/cpu.hpp
```diff
@@ -884,7 +884,25 @@ struct CPU : Thread {
...
+      bool singleInstruction;
+      Context::Endian endian;
+      Context::Mode mode;
+      bool cop1Enabled;
+      bool floatingPointMode;
+      bool is64bit;
+
+      u32  stateBits; //the above state, compressed
+
+      auto update(const Context& ctx, const CPU& cpu) -> void;
```
invertego: Slight simplification: just pass CPU& and access the context through cpu.context.

invertego (commented):
Looking better. Regarding performance tuning, there is no formal process in place, currently. I usually boot a few titles (with audio/video sync disabled) to look for noticeable impact on framerate, which is a pretty superficial way to analyze things. The starting area outside the castle in SM64 gets a lot of attention, but I can make no claim to it being representative of the performance of the catalog as a whole.

ares/n64/cpu/cpu.hpp
pekkavaa: Yeah 64 blocks is way too tiny so it must be constantly thrashing.

Looking at `rsp/recompiler.cpp`, I can see it has a two-level cache: a direct lookup from `context` via `address` (a direct-mapped cache) and then a second hash table lookup from `blocks`, with the key (address, size, self.pipeline). Apparently the first one guarantees pipeline correctness through the saved `block.pipeline` object. It's unclear to me how the second, `hashset<BlockHashPair> blocks`, guarantees that hash collisions don't result in invalid blocks being executed. Also, the set size doesn't seem to be capped.

Since RDRAM is much larger than 1024 words (RSP's IMEM size), the direct-mapped cache wouldn't obviously work. But perhaps it could be extended a simple hash table like N-way cache, like in CPUs? Like this:

```c++
    struct Pool {
      struct Row {
        struct Way {
          Block* block;
          u32 tag;
        } ways[2]; // <-- tune this size
      };
      Row rows[1 << 6]; // <-- tune this size
    };
```

And then construct the row index from low bits of the address, put high bits and the state bitvector to tag. Wouldn't need hashing and is somewhat resistant to conflicts. Checking both ways should be efficient.

I'm not sure if this is superior to a large hash table though. I'll test them both :)

pekkavaa: I now realize the "cache" idea I suggested is how it works right now with the "pool" of "blocks", just with a single cache "way".

pekkavaa:
I simplified the implementation. The cached CPU state has only the bits enabled that are actually used by the recompiler. I also got rid of the hash and store the tags in their own array.

Performance is slightly degraded in SM64 starting area: 139 VPS vs 144 VPS of master. I'd like to get it to be even or better before merging.

One idea I got was to reference blocks with an 32-bit offset instead of a full `Block*`. I got it working on top of master ([commit](https://github.com/pekkavaa/ares/commit/37e68dec51b3f3175a7d90495898a57d849d6113)) but it didn't improve the frame timings, but perhaps it would alleviate the now bloated `Pool` struct.

rasky:
This needs rebasing after the calling convention PR went in.

I'd still like to merge this even if it seems to regress on performance because it does two things that seem on the good path: 1) create the jitcontext thing which is mandatory for good performance
2) inline many ops

I also notice that there's always a performance downgrade when you start inlining stuff rather than calling external function. So we either assume all jits in other emulators are wrong, or this trend has to tip at some point and then somehow switch into an improvement. It might be that we still bloat blocks too much because of per-instruction overhead. Fixing that might eventually recover what seems lost performance now.

pekkavaa:
Rebased against master and confirmed that n64-systemtest still passes.

I looked at the implementation again now and see that the computed tag doesn't affect the index of a computed block. In practice this means the recompiler's cache will get thrashed if the tag bits change constantly. I recall commenting out the unused bits so it doesn't happen anymore, but recommend coming up with a smarter scheme in the future.

rasky:
Ok that might explain the reason why there is a speed pessimization rather than speedup. The context key should partecipare with the address as index for the cache, so that you basically LRU among blocks whose keys are (context,address).

BTW: another rebase now should green the CI.

pekkavaa:
I updated the logic, see the commit message. Basically one layer of indirection more for Blocks and a cache. The old Pool objects store the latest instance. Similar to what was discussed on Discord.

Performance is now _almost_ on par with master. I'm happy with it.

pekkavaa:
Also I took some "traces" of the block behavior. Here are some quick numbers (data in the same repo): https://github.com/pekkavaa/2025-09-20-recompiler-block-tracing/blob/master/docs/1_trace_stats.md

For example in SM64 castle grounds, there are 33.21k compiled block lookups per frame but only 4093 unique addresses. 36% of block cache hits are to the same address (idle loop?). So the keys are sparse and accesses unevenly distributed. This validates the old design (sparse array, very fast lookups) so I tried to stay close to it even with the state bits added.

One optimization idea: the `PoolRow` struct could be shrunk to 64 bits if the `Block* block` ptr is replaced with an `u32` offset.

ares/n64/cpu/recompiler.cpp
```diff
@@ -9,16 +10,68 @@ auto CPU::Recompiler::pool(u32 address) -> Pool* {
...
+
+  stateBits = toBits();
+}
+
+auto CPU::Recompiler::JITContext::toBits() const -> u32 {
+  u32 bits = 1; // first bit always set to make 0 invalid
+  bits |= singleInstruction ? 1 << 1 : 0;
+  bits |= endian ? 1 << 2 : 0;
+  bits |= (mode & 0x03) << 3;
+  bits |= cop1Enabled ? 1 << 4 : 0;
```
invertego: The `cop1Enabled` bit overlaps the upper bit of `mode`.

invertego: Including `cop1Enabled` is causing a lot of code to be compiled twice because the code that runs with cop1 disabled is effectively random. I believe this is because it gets disabled during thread context switches (to allow for lazy switching of FPU state).

A lot of this duplication will occur in blocks that use no FPU instructions, making it an entirely pointless expansion of the code cache. This could be addressed by ignoring the cop1 bit for such blocks.

This is effectively the only thing that blockCache helps with at the moment, and there only partially. It prevents code from being recompiled more than twice.

ares/n64/cpu/recompiler.cpp
```diff
@@ -9,16 +10,68 @@ auto CPU::Recompiler::pool(u32 address) -> Pool* {
...
+  assert(idx < sizeof(Pool::rows)/sizeof(Pool::rows[0]));
+
+  {
+    Pool* p = pool(address);
+    if (auto row = p->rows[idx]) {
+      if (row->stateBits == state) {
+        assert(row->block);
+        return row->block;
+      } else {
+        auto blockIt = blockCache.find(blockCacheKey(address, state));
```
invertego, line 29: This is reusing any previously compiled block with the same start address and the same CPU context. However, the instructions could be entirely different! This is why the RSP recompiler incorporates a hash of the instructions in the lookup.

invertego (changes requested):
At a minimum, the issue of blockCache lookup must be addressed.

ares/n64/cpu/recompiler.cpp
pekkavaa: Yikes. Yeah I forgot to clean up the block cache in `invalidatePool(address)`. I added an inverse mapping from a pool index to a list of cache keys but it's kind of cheesy. I'm not really happy with it. I wonder if there are other possible solutions other than a block content hash and this.

ares/n64/cpu/recompiler.cpp
pekkavaa: Fixed

pekkavaa: Tracking COP1 changes in the hash is indeed not useful right now. I've also understood that libultra flip flops the `cop1Enabled` flag on and off on thread switches.

I'll remove the COP1Enabled bit from the hash for now. It can be added back once its used during code generation. And perhaps the design can be revisited at that point too.

The block cache is not really meant to be hit often and it's functioning as a convenience to avoid manually writing "if kernel mode gets disabled, purge all blocks" logic. Right now it makes it possible to elide kernel and "dual mode enabled" checks in CPU::Recompiler::emit.

I actually measured that in SM64 the hit rate of the cache is about 99.185% (with COP1 bit included in context), which works out to  

```
(1-(99.185/100))*33210
= 270,6615
```

`unordered_map` lookups per frame (based on the 33.21k/frame lookups measurement from earlier). So performance-wise it's actually fine but the point about blocks being compiled twice still stands.

ares/n64/cpu/recompiler.cpp
pekkavaa: Yikes. Yeah I forgot to clean up the block cache in `invalidatePool(address)`. I added an inverse mapping from a pool index to a list of cache keys but it's kind of cheesy. I'm not really happy with it. I wonder if there are other possible solutions other than a block content hash and this.

ares/n64/cpu/recompiler.cpp
invertego: I also tested SM64, and the hit rate is 0% when the COP1 bit excluded. That's what I meant when I said that it's the only thing that it helps with.

ares/n64/cpu/recompiler.cpp
```diff
@@ -9,16 +10,71 @@ auto CPU::Recompiler::pool(u32 address) -> Pool* {
...
+  stateBits = toBits();
+}
+
+auto CPU::Recompiler::JITContext::toBits() const -> u32 {
+  u32 bits = 1; // first bit always set to make 0 invalid
+  bits |= singleInstruction ? 1 << 1 : 0;
+  bits |= endian ? 1 << 2 : 0;
+  bits |= (mode & 0x03) << 3;
+  bits |= cop1Enabled ? 1 << 5 : 0;
+  // bits |= floatingPointMode ? 1 << 6 : 0;
```
invertego, line 73: In a comment you mentioned you would exclude the cop1Enabled bit, but you removed the floatingPointMode bit instead. Was this a mistake?

ares/n64/cpu/recompiler.cpp
invertego: You could maintain a singly linked list of all blocks allocated for a given pool. The head pointer goes in `Pool` and next pointer goes in `Block`. You could walk the list in invalidatePool() instead of using poolBlockKeys. Honestly though, you could probably remove both blockCache and poolBlockKeys and just walk the list for lookup. The list shouldn't ever get that long.

ares/n64/cpu/cpu.hpp
```diff
@@ -883,13 +883,34 @@ struct CPU : Thread {
       u8* code;
     };
 
+    struct PoolRow {
+      Block* block;
+      u32 stateBits;
```
invertego, line 888: Consider moving stateBits to Block and getting rid of PoolRow to save an extra pointer indirection.

ares/n64/cpu/recompiler.cpp
pekkavaa: Sigh, yes. I'll fix it.

ares/n64/cpu/cpu.hpp
pekkavaa: Well it could be done but both the PoolRow and Block are allocated from the same bump allocator and usually are even neighboring, so it's not like the pointer indirection is random access to to heap.

ares/n64/cpu/recompiler.cpp
invertego: Adding to my last comment, you could actually keep the lists separate for each pool row to make the lookups even faster.

rasky:
Superseded by the new JIT

PR #1648, opened 2024-09-21
Fixing PowerFest '94 and Campus Challenge '92 for SNES

DerekTurtleRoe:
Hoping to get the last games fixed for SNES! Basically this is a draft PR, RFC.

Currently these modify how the DSP works and change the frequency of the oscillator to match newer information and the rest of the titles in the database, but those are subject to change.

Campus Challenge '92 seems to be fully functional, but PowerFest '94 seems to have issues regarding the DSP for Super Mario Kart. Super Mario Kart freezes using the latest nightly of ares.

On vanilla ares, neither of these boot.

bsnes and higan can boot these if they are in the appropriate format, and they seem to work fine there.

I tried to boot both into bsnes+ for more research, but I couldn't get that working, so it might be a core issue.

SNES9x can boot PowerFest '94, but not Campus Challenge '92, not sure why.

I would like to thank the ares Discord for help with this!

### DIFF
--- a/mia/Database/Super Famicom Boards.bml
+++ b/mia/Database/Super Famicom Boards.bml
@@ -664,7 +664,7 @@ board: BS-SA1-RAM
 
 board: EVENT-CC92
   memory type=RAM content=Save
-    map address=70-7d,f0-ff:0000-7fff mask=0x8000
+    map address=70-7d,f0-ff:0000-7fff mask=0x3fff
   processor manufacturer=NEC architecture=uPD78214
     identifier: Campus Challenge '92
     map address=c0,e0:0000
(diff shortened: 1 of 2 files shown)


FitzRoyX:
The entries should probably be moved to the //Prototypes section at the top of the bml. I would also rename //Prototypes to //Promos.

hackgrid:
Mario Kart works fine with PowerFest '94 and a concatted ROM with hash d7cf278767973c8aa66681dde46b5f6c61d4bbb0356b8b97d53b43cac750bf9f (program + slot-1 + slot-2 + slot-3 + dsp1.program + dsp1.data)

PR #1655, opened 2024-09-28
mia: try to load external sfc manifest if not found in internal db

hackgrid:
This allows fan translations like Heartthrob Memorial to use their own supplied manifest.bml file (e.g. for higan), without having to be included in the internal ares db or heuristics to be hardcoded for special cases.

### DIFF
--- a/mia/medium/super-famicom.cpp
+++ b/mia/medium/super-famicom.cpp
@@ -73,6 +73,14 @@ auto SuperFamicom::load(string location) -> bool {
   this->sha256   = Hash::SHA256(rom).digest();
   this->location = location;
   this->manifest = Medium::manifestDatabase(sha256);
+  
+  if(!manifest) {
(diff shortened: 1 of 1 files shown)


FitzRoyX:
All this means is that 10 years down the road, if the manifest format changes even slightly, the hack will stop working because you're making the broken manifest take precedence. This is the main problem with external mapping: we have no control over the patch, the websites that host them, etc. People are in the scene one day and MIA the next. We do, however, have total control over our dbs, and there is already an //Unlicensed section at the bottom of the sfc db.

hackgrid:
This is not true, the external manifest only gets loaded if there is no
entry in the internal db. So nothing lost or gained if a game is not in the
db and the format really changed.

FitzRoyX ***@***.***> schrieb am So., 29. Sept. 2024, 01:22:

LukeUsher:
Replaced by #1656

PR #1654, opened 2024-09-27
sfc: fix PowerFest '94 and Campus Challenge '92

hackgrid:
Based on database fixes and research from #1648 (thx @DerekTurtleRoe, remutro & SuperMikeMan100!)

Fixes #1361 

Only supports the merged one-file ROMs for these games (without appended firmware).
Also sets the default dip value for the known competition boards to 0x03 for 6 minutes (default mode used in these competitions)

See for dip settings:
https://www.snesmaps.com/maps/NintendoPowerFest94/NintendoPowerFest94.html
https://www.snesmaps.com/maps/1992Challenge/1992Challenge.html

Nitpick: The mia database for existing sfc games only contains sha256 hashes including all appended firmware - these two competition games use the hash WITHOUT firmware (=matches the no-intro hashes).

PR #1656, opened 2024-09-29
desktop-ui: enable loading rom folders from the cmd line

hackgrid:
Needs to merge #1654 first.

Enables loading folders instead files from cmd line.
Removes loading firmware files from rom folders in sfc mia (firmware data would be loaded twice, db hashes would not match anymore).

### DIFF
--- a/ares/sfc/coprocessor/dip/dip.hpp
+++ b/ares/sfc/coprocessor/dip/dip.hpp
@@ -8,7 +8,7 @@ struct DIP {
   //serialization.cpp
   auto serialize(serializer&) -> void;
 
-  n8 value = 0x00;
+  n8 value = 0x03; // default game time used for the competition is 6 minutes
 };
 
 extern DIP dip;
--- a/desktop-ui/desktop-ui.cpp
+++ b/desktop-ui/desktop-ui.cpp
@@ -138,7 +138,7 @@ auto nall::main(Arguments arguments) -> void {
 
   program.startGameLoad.reset();
   for(auto argument : arguments) {
-    if(file::exists(argument)) program.startGameLoad.append(argument);
+    if(file::exists(argument) || directory::exists(argument)) program.startGameLoad.append(argument);
   }
 
   Instances::presentation.construct();
--- a/mia/Database/Super Famicom Boards.bml
+++ b/mia/Database/Super Famicom Boards.bml
@@ -676,7 +676,7 @@ board: EVENT-CC92
       memory type=ROM content=Level-3
     dip
   processor manufacturer=NEC architecture=uPD7725
-    map address=20-3f,a0-bf:8000-ffff mask=0x7fff
+    map address=20-3f,a0-bf:8000-ffff mask=0x3fff
     memory type=ROM content=Program architecture=uPD7725
     memory type=ROM content=Data architecture=uPD7725
     memory type=RAM content=Data architecture=uPD7725
--- a/mia/Database/Super Famicom.bml
+++ b/mia/Database/Super Famicom.bml
@@ -287,6 +287,60 @@ game
       size: 0x80000
       content: Download
 
+//sha256 is without firmware
+game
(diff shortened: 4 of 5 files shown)


LukeUsher:
> Removes loading firmware files from rom folders in sfc mia (firmware data would be loaded twice, db hashes would not match anymore).

Could we skip loading the default firmware in this case and use the firmware shipped in the rom folder? While this would not work for complex carts that can’t be resolved with heuristics, it would work for using custom firmware for games in which heuristics can figure things out. 

A small use case I know; but being able to swap out the firmware on a game basis may have some value

hackgrid:
Sure, see my new commit.

But CC and PF will not boot anymore, if I use the supplied firmware in my rom folder.
I compared the internal ares firmware and the external higan? firmware, and they do not match.
So maybe the games do not boot because of a firmware mismatch.

Another potential reason could be that the heurestics for these two games fail in firmwareRomSize(), so it may work fine with other games.

If the rom folders do not contain external firmware files the games still work of course.

Screwtapello:
> I compared the internal ares firmware and the external higan? firmware, and they do not match.

To the best of my knowledge, the firmware chips have only been dumped once. I just recalculated the SHA256 hashes of all the firmware ares ships with, and the firmware pack I used in the higan days, and the only differences were the extra firmware blobs that ares ships with (for things like the SGB).

It's possible you're comparing a DSP1 blob (SHA256 91e87d11e1c30d172556bed2211cce2efa94ba595f58c5d264809ef4d363a97b) with a DSP1b blob (d789cb3c36b05c0b23b6c6f23be7aa37c6e78b6ee9ceac8d2d2aa9d8c4d35fa9), or possibly with the split program/data ROMs that higan likes—but if you concatenate those, they're exactly the same data as the combined ROMs.

hackgrid:
Ok, thx! Since the manifest only mentioned 

>       manufacturer: NEC
>       architecture: uPD7725
>       identifier: DSP1

I compared [dsp1.bin](https://github.com/ares-emulator/ares/blob/master/mia/Firmware/Super%20Famicom/dsp1.rom) 

to merged "dsp1.program.rom" + "dsp1.data.rom" file (sha256: d789cb3c36b05c0b23b6c6f23be7aa37c6e78b6ee9ceac8d2d2aa9d8c4d35fa9) from my campus challenge rom folder, which works completly fine in higan.

Most stuff is identical, but there are many differences as well.

So I guess it is DSP1b, since it matches your hash, but I wonder why it is just called "DSP1" in the manifest (and it works fine with ares then?).

hackgrid:
Loading external firmware actually worked before with CC and PF! I got mixed up I guess.
With these new changes, you can play both games with splitted roms with or without the firmware files.

Since the ares db only includes the no-intro hashes for both games, there will be a mismatch if load from rom folders containing the firmware files.

Because of that, I also now included #1655 with this PR, to support external manifests for single rom files and rom folders.
So if you simply copy the contents of the ares db for CC or PF 1:1 in a "manifest.bml" file in the corresponding rom folder (which also contains the dsp firmware files) it will work now :-)

Screwtapello:
> So I guess it is DSP1b, since it matches your hash, but I wonder why it is just called "DSP1" in the manifest (and it works fine with ares then?).

The DSP1b is a bugfixed version of the DSP1. The SNES internal header indicates that a game expects a "DSP1" chip but doesn't specify which variant. As a result, automatically-generated manifests tend to just give the filename as "DSP1" and  higan/bsnes defaults to DSP1b in that case.

That's not always the right call, though — for example, the Pilotwings attract sequence depends on bugs in the original DSP1, so if you run it with an original DSP1 chip it works fine, but with a DSP1b the plane crashes. Official, original Nintendo-manufactured carts exist with both variants. One of the reasons the verified game database exists, is so that there's an authoritative source that says "this game expects DSP1, that game expects DSP1b".

PR #1663, opened 2024-10-06
n64: remove call to epilogue function from the jit

rasky:


invertego:
Sorry for the delay. I will try to review this in the next few days.

ares/n64/cpu/recompiler.cpp
```diff
@@ -28,11 +30,16 @@ auto CPU::Recompiler::emit(u64 vaddr, u32 address, bool singleInstruction) -> Bl
     reset();
   }
 
-  auto block = (Block*)allocator.acquire(sizeof(Block));
+  // abort compilation of block asap if the instruction cache is not coherent
+  if(!self.icache.coherent(vaddr, address))
+    return nullptr;
+
+  bool abort = false;
```
invertego, line 37: `abort` is unused

ares/n64/cpu/cpu.cpp
```diff
@@ -132,11 +135,7 @@ auto CPU::instructionPrologue(u64 address, u32 instruction) -> void {
 
 template<bool Recompiled>
 auto CPU::instructionEpilogue() -> void {
-  if constexpr(Recompiled) {
-    //simulates timings without performing actual icache loads
-    icache.step(ipu.pc, devirtualizeFast(ipu.pc));
-    assert(ipu.r[0].u64 == 0);
```
invertego, line 138: A little concerned about losing this assert in debug builds. I guess we could replace it with a check in the JITed code. Not saying this has to be addressed in this PR, but I wanted to make note of it.

invertego (approved):

PR #1664, opened 2024-10-08
User configurable setting for N64 Controller Pak size

Cuyler36:
This PR adds support for user configurable N64 Controller Pak bank sizes. The motivation for this comes from Doubutsu no Mori. A NES emulator is included which can load Famicom/NES ROM images from the Controller Pak. The standard 1 bank 32KB Controller Pak is far too small for most games. Ares is the only N64 emulator that I've found which both runs the NES emulator and can load data off of the Controller Pak.

The N64 Controller Pak can support up to 62 banks of 32KB each. Nintendo did not utilize this feature as far as I'm aware, but Datel did make a 128KB 4-bank Controller Pak. Bank switching is done by writing to the Controller Pak with address line 15 enabled. The desired bank is sent to an IC on the Controller Pak as one block of bytes (32 bytes). The IC keeps track of the selected bank when further reads and writes happen.

The default should be 1 bank or 32KB of space which preserves compatibility with existing Controller Pak files.

### DIFF
--- a/ares/n64/controller/gamepad/gamepad.cpp
+++ b/ares/n64/controller/gamepad/gamepad.cpp
@@ -12,6 +12,8 @@ Gamepad::Gamepad(Node::Port parent) {
   port->setDisconnect([&] { return disconnect(); });
   port->setSupported({"Controller Pak", "Rumble Pak", "Transfer Pak"});
 
+  bank = 0;
+
   x           = node->append<Node::Input::Axis>  ("X-Axis");
   y           = node->append<Node::Input::Axis>  ("Y-Axis");
   up          = node->append<Node::Input::Button>("Up");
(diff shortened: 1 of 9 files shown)


jcm93:
UI-wise, a couple thoughts:

Presentation-wise, a text box for number input isn't always ideal, particularly when the minimum and maximum aren't immediately apparent. Have you considered a slider? Further, it might be more neat to present it as simply "Controller Pak Size" with a minimum of 32 KB and a maximum value of 1984 KB, if that's the maximum, rather than the user having to do math to see how big their pak would end up.

Would there be any drawbacks to even just presenting it also as something a bit simpler, i.e. a radio button with a couple size options, or even a checkbox to enable the largest size? Is there a good reason to need to express precise values?

Will lowering the size of the Controller Pak after running with a large size destroy Pak data? If so, it may be worth warning the user about that in some way if data is going to be lost.

Cuyler36:
My thoughts were that because this was such a niche use case that a text box with the # of banks would be acceptable. Most people wouldn't touch it due to being unnecessary for every other game that I'm aware of. If you did lower the size after formatting a larger Controller Pak it would only destroy data but the pak could still be recovered since all system pages are stored in bank 0.

That being said, I'm open to revising the UI. I think having the number of banks be granular is useful for people utilizing this for the intended purpose. And I agree that changing to KiB size is probably more appropriate for clarity. A slider would probably be the most efficient way to handle that although there would be a lot of snapping.

Jhynjhiruu:
Having the slider represent the size in KiB isn't necessarily ideal, since the relationship between the number of banks and the storage size isn't trivial (it's (banks - 1) * (32KiB - 256 bytes) + (32KiB - (2 * banks + 3) * 256 bytes), which... yeah). Perhaps the slider should be the number of banks, with a side label showing the amount of space the selected number gives?

Cuyler36:
I've reworked the settings menu to give four options based on officially produced offerings and the maximum allowed size. Controller Paks will be created if they don't exist upon connecting the N64 Controller. If a Controller Pak file already exists, that file's bank count will be used to temporarily override the configured bank count. A more robust Controller Pak management system might be worth considering in the future.

stgiga:
The slider MAY be useful in some cases, because you may never know whether a game can go to the full 62 banks.

Jhynjhiruu:
> The slider MAY be useful in some cases, because you may never know whether a game can go to the full 62 banks.  

All known games either only support 32KiB packs or can handle any size, so it would be for interest only.

LukeUsher:
There's a merge conflict here now that needs to be resolved if you want to take a look!

Cuyler36:
I should have time tomorrow to rebase this!

rasky:
@Cuyler36 if you tick this box, I can push the rebase of this PR (I've fixed the conflicts):
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

rasky:
I've worked a bit on this code base. I'll close this PR to open another one.

PR #1671, opened 2024-10-20
GNUmakefile: minor compiler and flags improvement

pstef:
With this applied, it's now possible to choose a different clang version, and pass the canonical GNU makefile variables CFLAGS and CXXFLAGS.

### DIFF
--- a/nall/GNUmakefile
+++ b/nall/GNUmakefile
@@ -128,13 +128,13 @@ endif
 
 # global compiler flags
 ifeq ($(cl),true)
-  flags.c      = -TC -std:c11
-  flags.cpp    = -TP -std:c++17 -EHsc
+  flags.c      = -TC -std:c11 $(CFLAGS)
+  flags.cpp    = -TP -std:c++17 -EHsc $(CXXFLAGS)
   flags       += -nologo -permissive- -utf-8 -W2 -Fd$(object.path)/
   options     += -nologo $(if $(findstring clang,$(compiler)),-fuse-ld=lld) -link
 else
-  flags.c      = -x c -std=c11
-  flags.cpp    = -x c++ -std=c++17
+  flags.c      = -x c -std=c11 $(CFLAGS)
+  flags.cpp    = -x c++ -std=c++17 $(CXXFLAGS)
   flags.objc   = -x objective-c -std=c11
   flags.objcpp = -x objective-c++ -std=c++17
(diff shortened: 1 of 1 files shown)


Mastergatto:
Hi, not an Ares developer here, but I'm looking forward to this PR with great interest, as I'm maintaining the AUR package. Although I'm not sure about the `CFLAGS`/`CXXFLAGS` part there.  

Did you made sure that `CFLAGS`/`CXXFLAGS` aren't being overridden by later and conflicting flags? For example, see here:
https://github.com/ares-emulator/ares/blob/a8274cd1277586bd2af221de19b9f4b4711c08e3/desktop-ui/GNUmakefile#L15-L24
This part only allows choice between `-march=native` and `-march=x86-64-v2`; and here I believe that it sets the flag later than `CFLAGS`/`CXXFLAGS` thus it would override whatever the user sets for `-march`, due to the compiler's policy.  And this is just one of a series of similar cases.

pstef:
I see it as a policy question and as such, beyond the scope of this PR.

If the project's decision is to set some flags explicitly and in a way that overwrites whatever flags the user might have passed, then I'm in no position to change that policy.

The way this patch works now is that I can pass whatever I think will serve me, and if it's something that is never set (or reset) later, then it's more flexible that way and brings me some convenience. But if it is reset then probably for good reason, so the project will have the final say.

If I'm really stubborn I can change that locally, hack, fork, etc.

That's at least how I see it, and I don't have strong feelings about it.

Mastergatto:
The thing is, the codebase passes a lot of compiler flags as you can see, so the usefulness of `CFLAGS`/`CXXFLAGS`, as laid out in this PR, would be more limited. It may work out for you, but it cannot be said the same for other users. In fact, some AUR users asked me about this `-march` thing.

In response, I have already created a (yet to be published) patch for the AUR package before this PR, essentially it passes `CFLAGS`/`CXXFLAGS` at a later stage, so that it does have always the last word.

Though, as it stands currently, I'm not fully convinced of the validity of my patch and I think there must be a better way to handle `CFLAGS`/`CXXFLAGS`. Maybe by rethinking the makefile, or by just waiting for the replacement of the current build system...

LukeUsher:
Closing as this is not particularly relevant after the move to cmake;

pstef:
Thanks for letting us know.

PR #1674, opened 2024-10-24
nall: Namespace noinline macro

jcm93:
This is a change spun off from continuing build system work.

There, we've found that precompiling the `ares/ares.hpp` header can speed up compile times for the emulator cores by 30-35%. Doing so, however, leads to a few collisions:

* Several files in parallel-rdp use the `Hash` type, which variously resolves to a `Hash` in its own namespace as well as nall's namespace.
* Also in the N64 core, nall's `noinline` macro conflicts with similar definitions present in `xxhash`.

For parallel-rdp, we can simply skip precompiled headers for the conflicting files; parallel-rdp does not use anything in the `ares/ares.hpp` header, so including it there is unnecessary anyway.

This change resolves the second set of collisions by putting nall's noinline preprocessor macro into its own namespace.

There are many macros remaining that are not namespaced. I did not proceed in namespacing them for this PR, because there are a lot of them, and it would be a fairly significant code style change that probably merits further discussion (not to mention, it will be a decent amount of work). However, in this instance, there is (or will be) a material benefit that probably trumps the minor style issue.

### DIFF
--- a/ares/component/processor/gsu/gsu.hpp
+++ b/ares/component/processor/gsu/gsu.hpp
@@ -75,8 +75,8 @@ struct GSU {
   auto serialize(serializer&) -> void;
 
   //disassembler.cpp
-  noinline auto disassembleInstruction() -> string;
-  noinline auto disassembleContext() -> string;
+  NALL_NOINLINE auto disassembleInstruction() -> string;
+  NALL_NOINLINE auto disassembleContext() -> string;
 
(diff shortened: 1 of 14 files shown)


jcm93:
This was verified to build (and work) locally but running on CI is probably a good idea because of some of the macro weirdness in the N64 interpreter.

PR #1681, opened 2024-10-28
Slower PS1 DMA. Fixes #1680

### LINKED ISSUE #1680: [PS1] Fear Effect - Stuck playing intro movie
**Describe the issue**

Fear Effect is stuck in an infinite loop while decoding the first introduction video. This is due to bad timing in the DMA MDEC code. The DMA callback supposed to notify that blocks are done being decoded, is called too early. The interrupt happens before the game starts waiting on a boolean value in a while loop. This causes the game to be trapped in this while loop.

**Rom Information**
Specify the impacted rom filename and any associated hashes (sha256 preferred);.
The easiest way to do this is to open the game in ares, and go to "Tools -> Manifest" and copy/paste the contents.

```
file: Fear Effect (Disc 1).cue (SLUS-00920)
sha1: e96eb6d83dd890103b3d3511cb84da73adab2cde
```

kivutar:
I'm not sure about the correctness of this patch.
It does fix Fear Effect, it might break other games.
I'll try to run other games to check for regressions.

### DIFF
--- a/ares/ps1/dma/dma.cpp
+++ b/ares/ps1/dma/dma.cpp
@@ -21,9 +21,9 @@ auto DMA::unload() -> void {
 
 auto DMA::main() -> void {
   for(u32 id : channelsByPriority) {
-    if(channels[id].step(16)) break;
+    if(channels[id].step(32)) break;
   }
-  step(16);
+  step(32);
 }
 
 auto DMA::step(u32 clocks) -> void {


kivutar:
I performed a few regression tests on:

* Final Fantasy Tactics (Japan)
* Suikoden II (USA)
* Alundra (USA) (Rev 1)
* Lunar 2 - Eternal Blue Complete (USA)
* Wild Arms (USA)

And they seem to still be able to decode videos.

LukeUsher:
This code changes from executing 16 cycles of DMA in 16 cycles worth of CPU time to executing 32 cycles of DMA in 32 cycles worth of CPU time; so the actual speed of execution is the same; there will however, be more time between blocks being processed as a side effect

This is likely hiding a more fundamental issue in how synchronisation/timing is handled overall in the PS1 core, rather than fixing the problem

kivutar:
It's very possible yeah. Here are the other informations I have, hoping that it can help you to understand the real issue:
the DMA callback is supposed to be invoked after a certain number of words have been transferred. According to my event log, it was triggered right away after the readDMAs were done. But mdec itself is supposed to consume time too. Looking at a decompilation of the game, it expect at least a few cpu cycles to happen before ABG_BlockCallback it called.

Here was the interesting part of the log:

```
  44838 MDEC::readDMA() -> 00000000
  44839 MDEC::readDMA() -> 00000000
  44840 MDEC::readDMA() -> 00000000
  44841 MDEC::readDMA() -> 00000000
  44842 self.channels[1].irq.flag & self.channels[1].irq.enable = 1
  44843 self.channels[1].irq.flag = 1
  44844 self.channels[1].irq.enable = 1
  44845 self.channels[1].irq.flag & self.channels[1].irq.enable = 0
  44846 self.channels[1].irq.flag = 0
  44847 self.channels[1].irq.enable = 1
  44848 ABG_BlockCallback
  44849 DAT_80082c61 = '^A';
  44850 DAT_80082c61 = 1; return;
  44851 do {} while (DAT_80082c61 == '^A');
  44852 do {} while (DAT_80082c61 == '^A');
  44853 do {} while (DAT_80082c61 == '^A');
  44854 do {} while (DAT_80082c61 == '^A');
  ```
  
  `ABG_BlockCallback` is expected after the game is in the `do {} while (DAT_80082c61 == '^A');` loop.
  
  I took a look at some other emulators but they are all very high level compared to ares. They schedule `ABG_BlockCallback` based on a calculation that involves a certain number of DMA words + a bias that depends on wether the game is PAL or NTSC..

LukeUsher:
The main issue here is that the DMA implementation is not fully aware of the other components; all hardware that can be read from/written to from DMA have signal lines to signify a ready state (either input fifo full, dma ready, or something along those lines)

The current ares DMA implementation does not use these signals, and the MDEC itself does not even implement these signals at this time; so it would appear they need to be implemented in the MDEC, and DMA needs a refactor

kivutar:
I see, I was unaware of this. Thanks for the pointers. I might come back with a more relevant PR later to implement the missing parts then. Don't feel obligated to keep this PR open.

PR #1688, opened 2024-11-07
nall, desktop-ui: Account for other *nix data locations

jcm93:
### Context

https://github.com/ares-emulator/ares/pull/1575 ended up defining the `sharedData` directory on Unix-like systems as `~/.local/share` if a build prefix was not defined. This is not exactly correct; the meaning of a "shared data" directory on a Unix-like system is variously `/usr/share`, `/usr/local/share`, or else the `share` directory of a given install prefix.

### Changes

This PR allows us to respect normal Unix conventions while also respecting the option to build and install with a given prefix. This is done by defining  two more functions in nall's `path`:
* `localSharedData()` for `/usr/local/share`
* `prefixSharedData()` for `${ARES_PREFIX}/share`

`sharedData()` is changed back to the old behavior and returns `/usr/share`.

Then, the `locate` logic in desktop-ui (and uh, mia) is adjusted to now search (in order):

1. the prefix's shared data if the prefix exists
2. local shared data
3. system shared data

### DIFF
--- a/desktop-ui/desktop-ui.cpp
+++ b/desktop-ui/desktop-ui.cpp
@@ -19,6 +19,18 @@ auto locate(const string& name) -> string {
   if(inode::exists(location)) return location;
 
   // 3. The shared data directory
+#if defined(PLATFORM_LINUX) || defined(PLATFORM_BSD)
+  /// Unix-like systems have multiple notions of a 'shared data' directory. First, check for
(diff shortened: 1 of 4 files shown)


thekovic:
Nice, this should improve the experience installing ares on my Kubuntu install, I think

PR #1722, opened 2024-12-15
n64: step rsp dma during rsp execution

invertego:
Instead of advancing RSP DMA only after each CPU basic block, allow it to run between each RSP block. This prevents the RSP from racing with the DMAs it initiates in the event of a long running CPU block (which can trigger multiple consecutive blocks of RSP execution).

This fixes hangs in Tarzan and possibly other games that use MusyX.

Another observable side effect of this change is that when the RSP is halted, it no longer risks falling increasingly behind the CPU, because ares will now step multiple times if needed instead of just once.

### DIFF
--- a/ares/n64/cpu/cpu.cpp
+++ b/ares/n64/cpu/cpu.cpp
@@ -67,7 +67,6 @@ auto CPU::synchronize() -> void {
 
   queue.step(clocks, [](u32 event) {
     switch(event) {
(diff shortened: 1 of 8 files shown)


rasky (approved):
Patch is fine 

I always find confusing that we keep clocks centered at 0, while they are 64 bit numbers anyway so we could just let them count up. Anyway this is just the status quo.

PR #1723, opened 2024-12-15
Implement the hex editor widget for Cocoa and enable memory editor on macOS

Cathrach:
This PR adds the hex editor widget to Cocoa hiro and enables the memory editor in Tools. The widget is a cell-based NSTableView with data source the core hex editor widget and supports editing of memory (but not of the encoded ANSI).

### DIFF
--- a/hiro/cocoa/widget/hex-edit.cpp
+++ b/hiro/cocoa/widget/hex-edit.cpp
@@ -5,43 +5,157 @@
 -(id) initWith:(hiro::mHexEdit&)hexEditReference {
   if(self = [super initWithFrame:NSMakeRect(0, 0, 0, 0)]) {
     hexEdit = &hexEditReference;
+    if (tableView = [[NSTableView alloc] initWithFrame:[self bounds]]) {
+      [tableView setDataSource:self];
+      [tableView setHeaderView:nil];
+      
+      // remove padding and stuff to fit all 16 columns
+      [tableView setIntercellSpacing:NSMakeSize(0, 0)];
+      [tableView setStyle:NSTableViewStylePlain];
+      NSTableColumn* addressCol = [[NSTableColumn alloc] initWithIdentifier:@"Address"];
+      addressCol.editable = NO;
+      [tableView addTableColumn:addressCol];
+      
+      for (int i = 0; i < hexEdit->columns(); i++) {
+        NSTableColumn* col = [[NSTableColumn alloc] initWithIdentifier:[NSString stringWithFormat:@"%d",
+                                                                        i]];
+        col.width = 20;
+        col.title = @"";
+        [tableView addTableColumn:col];
+      }
+      /*
+       NSTableColumn* hexCol = [[NSTableColumn alloc] initWithIdentifier:@"Hex"];
+       [hexCol setTitle:@"Hex"];
+       [tableView addTableColumn:hexCol];
+       */
+      
+      NSTableColumn* ansiCol = [[NSTableColumn alloc] initWithIdentifier:@"ANSI"];
+      ansiCol.editable = NO;
+      [tableView addTableColumn:ansiCol];
+      
+      // must programmatically do this so the table shows up.
+      [self setDocumentView:tableView];
+      [self setHasVerticalScroller:YES];
+    }
   }
   return self;
+}
+
+- (NSTableView *) tableView {
+  return tableView;
+}
+
+- (NSInteger) numberOfRowsInTableView:(NSTableView *)tableView {
+  return hexEdit->rows();
+}
+
+-(id) tableView:(NSTableView *) tableView
+objectValueForTableColumn:(NSTableColumn *) tableColumn
+            row:(NSInteger) row {
+  // return a string with the content of (row, column)
+  u32 address = hexEdit->address() + row * hexEdit->columns();
+  // address only
+  if ([[tableColumn identifier] isEqualToString:@"Address"]) {
+    return [NSString stringWithUTF8String:hex(address, 8L).data()];
+  } else if ([[tableColumn identifier] isEqualToString:@"ANSI"]){
+    // create the ansi string by concatenating
+    NSMutableString *output = [NSMutableString stringWithCapacity:hexEdit->columns()];
+    for (auto column : range(hexEdit->columns())) {
+      if (address < hexEdit->length()) {
+        u8 data = hexEdit->doRead(address++);
+        [output appendString:[NSString stringWithFormat:@"%c",
+                              (data >= 0x20 && data <= 0x7e ? (char)data : '.')]];
+      }
+    }
+    return output;
+  } else {
+    // return only the single byte at the correct position
+    NSInteger columnNumber = [tableColumn.identifier integerValue];
+    address += columnNumber;
+    if (address < hexEdit->length()) {
+      u8 data = hexEdit->doRead(address);
+      return [NSString stringWithUTF8String:hex(data, 2L).data()];
+    } else {
+      return @"  ";
+    }
+  }
+}
+
+-(void) tableView:(NSTableView *) tableView
+   setObjectValue:(id) object
+   forTableColumn:(NSTableColumn *) tableColumn
+              row:(NSInteger) row {
+  // when table is edited, modify underlying data source
+  NSInteger colNumber = [tableColumn.identifier integerValue];
+  u32 address = hexEdit->address() + row * hexEdit->columns() + colNumber;
+  
+  // unclear if this is needed, but just to make sure the cast is safe
+  if ([object isKindOfClass:[NSString class]]) {
+    if (address < hexEdit->length()) {
+      // only get the first 2 characters
+      NSString* newVal = [(NSString *)object substringToIndex:2];
+      NSScanner* hexScanner = [NSScanner scannerWithString:newVal];
+      unsigned int data = 0;
+      if ([hexScanner scanHexInt:&data]) {
+        // this.......is probably ok....???
+        hexEdit->doWrite(address, (u8)data);
+      }
+    }
+    
+  }
+  [tableView reloadData];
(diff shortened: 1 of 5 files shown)


hiro/cocoa/widget/hex-edit.cpp
```diff
@@ -5,43 +5,157 @@
...
-auto pHexEdit::construct() -> void {
-  cocoaView = cocoaHexEdit = [[CocoaHexEdit alloc] initWith:self()];
+- (NSTableView *) tableView {
+  return tableView;
 }
 
-auto pHexEdit::destruct() -> void {
-  [cocoaView removeFromSuperview];
+- (NSInteger) numberOfRowsInTableView:(NSTableView *)tableView {
+  return hexEdit->rows();
```
jcm93: Hmm; it looks like in other hex-edit implementations, `rows` is being dynamically set based on the amount of memory being looked at (`hexEdit->length()`). For Cocoa here, we aren't currently doing that, so this value appears to always remain at 16. I think we need to derive the number of rows from `hexEdit->length` in this method, or else set `hexEdit`'s `rows` somewhere else based on the total memory length.

Cathrach: Thank you – I was misunderstanding the functionality. Newest commit fixes this so rows = length / columns and properly implements address goto!

hiro/cocoa/widget/hex-edit.cpp
```diff
@@ -5,43 +5,163 @@
...
+  if ([[tableColumn identifier] isEqualToString:@"Address"]) {
+    return [NSString stringWithUTF8String:hex(address, 8L).data()];
+  } else if ([[tableColumn identifier] isEqualToString:@"ANSI"]){
+    // create the ansi string by concatenating
+    NSMutableString *output = [NSMutableString stringWithCapacity:hexEdit->columns()];
+    for (auto column : range(hexEdit->columns())) {
+      if (address < hexEdit->length()) {
+        u8 data = hexEdit->doRead(address++);
+        [output appendString:[NSString stringWithFormat:@"%c",
+                              (data >= 0x20 && data <= 0x7e ? (char)data : '.')]];
```
jcm93, line 75: This is something of a nit/bikeshedding, since the terminology isn't exposed to the user currently, but "ANSI" is sort of a Windows-ism, which just sort of generically means "the system's default encoding". On macOS, that term isn't really used, since basically everything is UTF-8. It looks like when we're interpolating the `char` here, we're using the UTF-8 system-default interpolation of the character, but just constrained to the range 32-127 of printable characters that have only one codepoint. Not worth changing the names of everything (unless you feel like it), but if the column header is ever exposed to the user, worth keeping in mind.

I also personally would prefer a whitespace instead of a `.`, since the actual `.` character is within the 32-127 range, but others might have other opinions.

hiro/cocoa/widget/hex-edit.cpp
```diff
@@ -5,43 +5,163 @@
 -(id) initWith:(hiro::mHexEdit&)hexEditReference {
   if(self = [super initWithFrame:NSMakeRect(0, 0, 0, 0)]) {
     hexEdit = &hexEditReference;
+    if (tableView = [[NSTableView alloc] initWithFrame:[self bounds]]) {
+      [tableView setDataSource:self];
+      [tableView setHeaderView:nil];
+      
+      // remove padding and stuff to fit all 16 columns
+      [tableView setIntercellSpacing:NSMakeSize(0, 0)];
+      [tableView setStyle:NSTableViewStylePlain];
```
jcm93: We currently target macOS 10.13, where explicit NSTableView styling isn't available, so this might crash on systems below macOS 11.0. You should be able to remove this line without any impact, since the automatic styling would be used which should still resolve to the default plain style.

hiro/cocoa/widget/hex-edit.cpp
```diff
@@ -5,43 +5,163 @@
...
+      NSTableColumn* addressCol = [[NSTableColumn alloc] initWithIdentifier:@"Address"];
```
jcm93, line 18:
```suggestion
      NSTableColumn* addressCol = [[NSTableColumn alloc] initWithIdentifier:@"Address"];
      [tableView setUsesAlternatingRowBackgroundColors:true];
```

I'd suggest leveraging the built-in functionality for alternating row backgrounds, since it should help out with legibility for a data-dense table.

hiro/cocoa/widget/hex-edit.cpp
```diff
@@ -5,43 +5,163 @@
...
+namespace hiro {
+  
+  auto pHexEdit::construct() -> void {
+    cocoaView = cocoaHexEdit = [[CocoaHexEdit alloc] initWith:self()];
+    pWidget::construct();
+    for (NSTableColumn *column in cocoaHexEdit.tableView.tableColumns) {
+      NSTextFieldCell *cell = column.dataCell;
+      
+      // Set the font for the data cell.
+      cell.font = [NSFont monospacedSystemFontOfSize:10 weight:NSFontWeightRegular];
```
jcm93: This is also only available on macOS 10.15 or above, so we should guard it with an availability check:
```suggestion
      if (@available(macOS 10.15, *)) {
        cell.font = [NSFont monospacedSystemFontOfSize:10 weight:NSFontWeightRegular];
      }
```

I forget how onerous implementing other fonts for cells below macOS 10.15 is; if it's annoying, I wouldn't worry too much about it. You may, however, need to widen the hex columns if the monospace font isn't available.

jcm93 (commented):
Thanks very much for the work here! One other general comment is that we may want a slightly larger gap between the last hex column and the column showing the encoded text, just so the separation is a bit more apparent.

jcm93:
One other thing; the memory editor is also current ifdef'd out in the menu bar's Tools menu;

https://github.com/ares-emulator/ares/blob/a752b85d6f60e38b8e58b6f92c8c4ddf1c43c020/desktop-ui/presentation/presentation.hpp#L70

Should remove that in the header there as well as the implementation file.

hiro/cocoa/widget/hex-edit.cpp
```diff
@@ -5,43 +5,167 @@
 -(id) initWith:(hiro::mHexEdit&)hexEditReference {
   if(self = [super initWithFrame:NSMakeRect(0, 0, 0, 0)]) {
     hexEdit = &hexEditReference;
+    if (tableView = [[NSTableView alloc] initWithFrame:[self bounds]]) {
+      [tableView setDataSource:self];
+      // [tableView setHeaderView:nil];
+      [tableView setUsesAlternatingRowBackgroundColors:true];
+
```
jcm93, line 14:
```suggestion
      [tableView setGridStyleMask:(NSTableViewDashedHorizontalGridLineMask | NSTableViewSolidVerticalGridLineMask)];
      [tableView setRowHeight:14.0];
      
```

Makes it a bit more compact and justifies the font within the grid better: Here's a comparison:

Screenshot 2024-12-15 at 6 30 00 PM
Screenshot 2024-12-15 at 6 32 02 PM

hiro/cocoa/widget/hex-edit.cpp
```diff
@@ -5,43 +5,167 @@
...
+      NSString* newVal = [(NSString *)object substringToIndex:2];
+      NSScanner* hexScanner = [NSScanner scannerWithString:newVal];
+      unsigned int data = 0;
+      if ([hexScanner scanHexInt:&data]) {
+        // this.......is probably ok....???
+        hexEdit->doWrite(address, (u8)data);
+      }
+    }
+    
+  }
```
jcm93: I don't think this check is needed. The only way we end up here is when the user enters text into an `NSTextFieldCell`, so the value should in all circumstances already be coerced to a string:

```suggestion
  if (address < hexEdit->length()) {
    // only get the first 2 characters
    NSString* newVal = [(NSString *)object substringToIndex:2];
    NSScanner* hexScanner = [NSScanner scannerWithString:newVal];
    unsigned int data = 0;
    if ([hexScanner scanHexInt:&data]) {
      // this.......is probably ok....???
      hexEdit->doWrite(address, (u8)data);
    }
  }
```

hiro/cocoa/widget/hex-edit.cpp
```diff
@@ -5,43 +5,167 @@
...
 
+    for (NSTableColumn *column in cocoaHexEdit.tableView.tableColumns) {
+        if (@available(macOS 10.15, *)) {
+        NSTextFieldCell *cell = column.dataCell;
+          NSTableHeaderCell *headerCell = column.headerCell;
+        
+        // Set the font for the data cell.
+        cell.font = [NSFont monospacedSystemFontOfSize:10 weight:NSFontWeightRegular];
+      }
+    }
```
jcm93: Just some messed up indentation.

```suggestion
    for (NSTableColumn *column in cocoaHexEdit.tableView.tableColumns) {
      if (@available(macOS 10.15, *)) {
        NSTextFieldCell *cell = column.dataCell;
        NSTableHeaderCell *headerCell = column.headerCell;
      
        // Set the font for the data cell.
        cell.font = [NSFont monospacedSystemFontOfSize:10 weight:NSFontWeightRegular];
      }
    }
```

jcm93 (commented):
Looking good! Just a couple more style suggestions / nits.

hiro/cocoa/widget/hex-edit.cpp
Cathrach: Thanks – I changed to a "Plain text" header. The `.` is for consistency with the windows/gtk versions; also, I think whitespace is difficult to count.

jcm93:
Looks good to me! If two-way editing is a part of the GTK/Windows version, I think it can be worried about later, this seems plenty good for a first iteration to me at least.

jcm93 (approved):

PR #1736, opened 2024-12-31
build: Finalize CMake merge

jcm93:
* Remove version overrides from CMakePresets.json
    * These were necessary for CI to run when CMake existed on the forked branch, but *should* no longer be necessary
* Disable release flows for the legacy build
    * Keep legacy builds running for the time being, but disable the release flow so it does not conflict with the new CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -179,7 +179,7 @@ jobs:
 
      
   release:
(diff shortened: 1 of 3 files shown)


jcm93:
Added one other small bugfix to this; exposing the `ARES_CORES` variable to the user by no longer marking it as INTERNAL. Pardon my sloppiness, just did not want to bother with a second PR for a trivial change in this instance.

PR #1782, opened 2025-01-19
Use DB entries to correct games with bad headers for the 32X

### LINKED ISSUE #1066: [32X] Rom list of PAL games with the wrong region in the header.
Title is self explanatory.

All games from the list ares launches at 60Hz.

Two version, SHA1 for public and SHA256 that ares uses internally. Open with notepad, or any other text editor that you like.

[Sega - 32X-SHA1.zip](https://github.com/ares-emulator/ares/files/11132855/Sega.-.32X-SHA1.zip)
[Sega - 32X-SHA256.zip](https://github.com/ares-emulator/ares/files/11132856/Sega.-.32X-SHA256.zip)

ares nightly 2023-03-31

remutro:
This is a more focused fix for the 32X to only modify games with known bad headers based on game hash instead of using ROM name.

Fixes: https://github.com/ares-emulator/ares/issues/1066

### DIFF
--- a/mia/medium/mega-32x.cpp
+++ b/mia/medium/mega-32x.cpp
@@ -143,13 +143,14 @@ auto Mega32X::analyze(vector<u8>& rom) -> string {
     regions.append("NTSC-J", "NTSC-U", "PAL");
   }
 
-  // FIFA Soccer 96 and Mortal Kombat II have an incorrect header, so force PAL based on name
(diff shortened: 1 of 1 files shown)


remutro:
This seems to cause audio issues in the PAL only game FIFA 96. Reverting behavior in a future PR.

PR #1785, opened 2025-01-20
mia, desktop-ui: Add basic error handling to ROM loading

jcm93:
> [!NOTE]
> Seeking testers on this to validate that errors appear correctly, are accurate, and we didn't break anything.

### Description

Adds the `LoadResult` type, used to identify and propagate upwards errors encountered when loading ROMs and systems.

Modifies the `mia` and `desktop-ui` load routines to use this type instead of generic booleans, so that if we have a load failure, we can propagate a precise and useful error to the user rather than a vague and possibly mismatched one.

Currently the scope of the error handling is just `mia` and `desktop-ui`; errors encountered when loading in `ares` itself are currently generic "`otherError`"s. `LoadResult` is placed in `mia/mia.hpp` so it is accessible to both `mia` and `desktop-ui`. If LoadResult is added to ares loading in the future, the LoadResult type may need to be moved.

### DIFF
--- a/desktop-ui/emulator/game-gear.cpp
+++ b/desktop-ui/emulator/game-gear.cpp
@@ -30,30 +30,24 @@ GameGear::GameGear() {
 auto GameGear::load() -> LoadResult {
   game = mia::Medium::create("Game Gear");
   string location = Emulator::load(game, configuration.game);
-  if(!location) return LoadResult(noFileSelected);
+  if(!location) return noFileSelected;
   LoadResult result = game->load(location);
-  if(result != LoadResult(successful)) return result;
+  if(result != successful) return result;
 
   auto region = Emulator::region();
 
   system = mia::System::create("Game Gear");
   result = system->load(firmware[0].location);
-  if(result != LoadResult(successful)) {
-    result.firmwareSystemName = "Game Gear";
-    result.firmwareType = firmware[0].type;
-    result.firmwareRegion = firmware[0].region;
-    result.result = noFirmware;
-    return result;
-  }
+  if(result != successful) return otherError;
 
-  if(!ares::MasterSystem::load(root, {"[Sega] Game Gear (", region, ")"})) return LoadResult(otherError);
+  if(!ares::MasterSystem::load(root, {"[Sega] Game Gear (", region, ")"})) return otherError;
 
   if(auto port = root->find<ares::Node::Port>("Cartridge Slot")) {
     port->allocate();
     port->connect();
   }
 
-  return LoadResult(successful);
+  return successful;
 }
 
 auto GameGear::save() -> bool {
--- a/desktop-ui/emulator/game-gear.cpp
+++ b/desktop-ui/emulator/game-gear.cpp
@@ -1,6 +1,6 @@
 struct GameGear : Emulator {
   GameGear();
-  auto load() -> bool override;
+  auto load() -> LoadResult override;
   auto save() -> bool override;
   auto pak(ares::Node::Object) -> shared_pointer<vfs::directory> override;
 };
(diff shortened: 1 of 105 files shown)


LukeUsher:
Since most of these errors are regarding ROM loading and manifest parsing; which mia is responsible for creating/handling; I believe these error types would be better suited as part of mia.hpp.

desktop-ui/emulator/game-gear.cpp
```diff
@@ -27,23 +27,33 @@ GameGear::GameGear() {
...
+  if(result != LoadResult(successful)) return result;
 
   auto region = Emulator::region();
 
   system = mia::System::create("Game Gear");
-  if(!system->load(firmware[0].location)) return false;
+  result = system->load(firmware[0].location);
+  if(result != LoadResult(successful)) {
+    result.firmwareSystemName = "Game Gear";
+    result.firmwareType = firmware[0].type;
```
LukeUsher: firmware is optional for Game Gear; as most as only later models have a firmware on real hardware, so if system->load fails, the reason must be something other than firmware.

desktop-ui/emulator/master-system.cpp
```diff
@@ -101,10 +104,17 @@ auto MasterSystem::load() -> bool {
   if(region == "NTSC-U") regionID = 0;
 
   system = mia::System::create("Master System");
-  if(!system->load(firmware[regionID].location)) return false;
-  if(!game->pak && !system->pak->read("bios.rom")) return false;
+  result = system->load(firmware[regionID].location);
+  if(result != LoadResult(successful)) {
```
LukeUsher: Like Game Gear; Firmware is also optional here

desktop-ui/emulator/mega-cd-32x.cpp
```diff
@@ -60,14 +63,21 @@ auto MegaCD32X::load() -> bool {
...
     if(emulator->name == "Mega CD") firmware = emulator->firmware;
   }
-  if(!firmware) return false;  //should never occur
+  if(!firmware) return LoadResult(noFirmware);  //should never occur
 
   system = mia::System::create("Mega CD 32X");
-  if(!system->load(firmware[regionID].location)) return errorFirmware(firmware[regionID], "Mega CD"), false;
+  result = system->load(firmware[regionID].location);
+  if(result != LoadResult(successful)) {
+    result.firmwareSystemName = "Mega CD 32X";
```
LukeUsher: This should use the same system name as Mega CD; it's not separate firmware

LukeUsher (commented):
Just a few notes

jcm93:
Taking this out of draft now that it's in a better state. 

https://github.com/ares-emulator/ares/blob/90860c24422e53861d71bc06ee15daaa2b127eed/mia/medium/arcade.cpp#L12

I am still not 100% on if this is correct/representative of the underlying behavior. Arcade seems to be the only system with an existing failure condition here. It seems like it's bailing out if the game entry wasn't in the database rather than using heuristics like the other cores. Just want to make sure this is accurate. And if so, maybe a more descriptive error message is appropriate? "The selected [game/Arcade system] is not currently supported by ares"?

LukeUsher:
> Taking this out of draft now that it's in a better state.
> 
> https://github.com/ares-emulator/ares/blob/90860c24422e53861d71bc06ee15daaa2b127eed/mia/medium/arcade.cpp#L12
> 
> I am still not 100% on if this is correct/representative of the underlying behavior. Arcade seems to be the only system with an existing failure condition here. It seems like it's bailing out if the game entry wasn't in the database rather than using heuristics like the other cores. Just want to make sure this is accurate. And if so, maybe a more descriptive error message is appropriate? "The selected [game/Arcade system] is not currently supported by ares"?

This is indeed correct; due to how arcade roms are formatted, ares does not know how to load them without a database entry specifying which individual roms need to be loaded where.

PR #1794, opened 2025-01-27
sfc: adjust overscan numbers

FitzRoyX:
These coord and offset numbers will hide all empty space (including blanked active lines) without cropping any good pixels on the sides. The output now matches old ares and bsnes. Whether the blanked lines should be shown can be revisited later (most users have no idea what these are and will hate them). SMW USA 1.0, SMW EUR 1.0, and SMW EUR 1.1 were used for testing as each variant triggers each of the 3 modes and the game draws border tiles all around.

Screenshots:

SMW USA 1.0

smw_u10_overscan

smw_u10_xhideoverscan

SMW EUR 1.0

smw_e10_overscan

smw_e10_xhideoverscan

SMW EUR 1.1

smw_e11_overscan

smw_e11_xhideoverscan

### DIFF
--- a/ares/sfc/ppu-performance/ppu.cpp
+++ b/ares/sfc/ppu-performance/ppu.cpp
@@ -89,19 +89,19 @@ auto PPU::main() -> void {
       screen->setSize(564, height() * 2);
       screen->setViewport(0, 0, 564, height() * 2);
     } else {
-      int x = 24;
-      int y = 16;
-      int w = 564 - 48;
-      int h = height() - 16;
+      int x = 26;
+      int y = 18;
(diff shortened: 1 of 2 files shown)


LukeUsher (commented):
While this specifically undoes the work done to make output closer to real hardware, I understand any borders are not desirable for most; so this will be merged for now while we discuss future development of display

ideally we want to be able to support three scenarios: the entire display (which is currently shown when overscan is enabled), the region that is usually shown on a CRT including correct offset and borders (currently shown when overscan disabled), or just active display (what this PR implemented for snes)

PR #1796, opened 2025-01-28
arm7tdmi: add instruction test harness

invertego:


tests/arm7tdmi/arm7tdmi.cpp
```diff
@@ -0,0 +1,522 @@
...
+  processor.fiq.spsr = is.SPSR[0];
+  processor.svc.spsr = is.SPSR[1];
+  processor.abt.spsr = is.SPSR[2];
+  processor.irq.spsr = is.SPSR[3];
+  processor.und.spsr = is.SPSR[4];
+
+  const bool thumb = processor.cpsr.t;
+  const u32 length = thumb ? 2 : 4;
+
+  //tests/v1/arm_mrs.json
```
invertego, line 184: Just wanted to note the obvious: it would be nicer if these exceptions were data-driven instead of hardcoded. However, the end goal is to improve the tests (and ares in the process) and get rid of them entirely.

Issue #1787, opened 2025-01-21
Memory Editor Changing Bytes is Problematic


RevenantBob:
**Describe the bug**
When using the memory editor for N64 trying to overwrite a byte types 2 characters instead of 1 for the key input. For example, If I want to set a byte to "C0" when I type C on the keyboard I get "CC".

**To Reproduce**
1. Load an N64 Rom
2. Open Memory Editor from Tools
3. Find a bye in RDRAM
4. Try to edit byte

**Expected behavior**
Typing a character should only overwrite 1 character, not two.

**Screenshots**

Before typing C I have cursor in front of "22".

![Image](https://github.com/user-attachments/assets/5360942d-3533-4112-8419-914e3f3328b9)

After typing C once, 22 became CC, when it should have become C2.

![Image](https://github.com/user-attachments/assets/f56e51f8-dab9-433e-bd53-66aa13f63ad2)

remutro:
I could have sworn there was already an open issue for this but I'm having trouble locating it now. However this is a known issue. If it helps, the issue is Windows specific (works properly on Linux and MacOS).

More info:

Basically in hiro/windows/application.cpp it is reading in key presses, where in the Linux & MacOS implementations of application they do not. The memory editor window itself, on all platforms, is also trapping key press events. The presence of both on Windows results in it capturing input twice, and thus outputting this double input capture (you can comment out this part in windows/application.cpp and the memory editor works fine). I'm just not really sure why it was done differently for Windows. I looked at the commit at one point (long time ago, done by Near) where that change happened, but I don't think it was obvious as to why it needed to be special cased on this platform, or maybe I just didn't spend enough time trying to figure it out. Needs to be reviewed in more detail.

invertego:
Fixed by #1798

PR #1798, opened 2025-01-28
hiro: fix double character inputs in Windows HexEdit widget

invertego:
At some point it might be worth going through the list of widgets to understand which ones require WM_KEYDOWN messages to be forwarded by Application_keyboardProc. For the time being, it's clear the HexEdit widget is not among them, so this change disables forwarding for that widget type.

As an added bonus, this change also makes the number pad functional.


invertego:
Fixes #1787

PR #1801, opened 2025-01-29
fc: adjust overscan numbers

FitzRoyX:
Overscan=off now shows active display centered and intact for both regions. The pal ppu micromasking behavior was corrected (253 vs 254) and commented out.

SMB2 EUR full frame
smb2_e_fullframe   

before/after

smb2before   ares-n

before/after
Untitledggg    Untitledhhh

### DIFF
--- a/ares/fc/ppu/ppu.cpp
+++ b/ares/fc/ppu/ppu.cpp
@@ -97,11 +97,14 @@ auto PPU::frame() -> void {
     screen->setViewport(0, 0, 283, displayHeight());
   } else {
     int x = 16;
-    int y = 8;
-    int width = 283 - 33;
-    int height = displayHeight() - 16;
+    int y = 0;
+    int width = 283 - 27;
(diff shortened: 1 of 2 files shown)


LukeUsher:
I am happy to accept the overscan changes in this PR

However, I would like to keep the blanking of the sides and top in PAL regions, though. My reasoning here is that while overscan is a property of the display and is subjective, the blanking of the top 1px, left 2px, and right 2px is an actual behaviour of the PAL PPU and is *not* subjective; there is no way real hardware would ever display these lines regardless of overscan/scaling.

if you want to keep the pixel aspect ratio/screenshot size the same, I'm happy for those lines to be left black, but I hold firm that we should keep them emulated, as this is how the PPU behaves.

FitzRoyX:
Yeah I had a feeling you'd say that. It would be nice to have a toggle for this in the video section at some point.

PR #1807, opened 2025-02-02
m32x+md fixes

### LINKED ISSUE #937: [Mega 32X] BC Racers. Sound Samples are not played back
Greetings. 

The game BC Racers has some sound samples that are played back. For example, when you run into a pedestrian or when you jump over a bridge. Both produce a noise. So far no 32X core emulates this game correctly in that regard, unfortunatly. 

Fastest way to reproduce is just to start and drive right into a pedestrian.

### LINKED ISSUE #1196: [MEGA CD-32X] Night Trap: video/sound stuttering
**Describe the issue**
After the Digital Pictures initial logo all video content (which is whole game) stutters.

**Rom Information**
http://redump.org/disc/17550/
http://redump.org/disc/17551/

ares nightly 2023-08-15

TascoDLX:
1. Revised pwm emulation, mostly to align with documentation. Not likely to affect much except for rare edge cases. Also, added a dc filter to prevent audio popping in certain situations (like window losing focus), but pops on init are correct behavior. Fixes #937 by implementing byte-wide fifo writes from the SH2 side.
2. Various fixups to 32x io access by the 68k. Fixes #1196 and missing music in Brutal.
3. Follow-up to a previous PR: cram dots were not limited properly, and line buffer adjustment was crashing 32x in certain cases (as reported in the discord).

### DIFF
--- a/ares/md/m32x/pwm.cpp
+++ b/ares/md/m32x/pwm.cpp
@@ -1,7 +1,6 @@
 auto M32X::PWM::load(Node::Object parent) -> void {
   stream = parent->append<Node::Audio::Stream>("PWM");
   stream->setChannels(2);
-  n12 clocks = cycle - 1;
   updateFrequency();
 }
 
@@ -11,7 +10,8 @@ auto M32X::PWM::unload(Node::Object parent) -> void {
 }
 
 auto M32X::PWM::main() -> void {
-    n12 clocks = cycle - 1;
+    int clocks = cycle > 0 ? cycle-1 : 4095;
+    if(clocks == 0) clocks++;
 
     //check if cycle rate has changed and update sample rate accordingly
     static n12 previousCycle = cycle;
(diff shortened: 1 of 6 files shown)


ares/md/m32x/pwm.cpp
```diff
@@ -11,7 +10,8 @@ auto M32X::PWM::unload(Node::Object parent) -> void {
 }
 
 auto M32X::PWM::main() -> void {
-    n12 clocks = cycle - 1;
+    int clocks = cycle > 0 ? cycle-1 : 4095;
```
LukeUsher: This change shouldn’t have been necessary, if clocks was 0, being an n5 it should have underflowed and wrapped to 4095 anyway

TascoDLX: You are correct. I had changed this at some point but I guess it just wound up back where it started.

TascoDLX:
Added fix to pixel alignment between H32 and H40 modes. This is also follow-up to the previous PR where borders were added.

PR #1825, opened 2025-02-09
nall: Resolve all file paths passed on command line

jcm93:
When loading a game from the command line with a relative path with a saves directory unspecified, `nall::Location` would fail to return a valid directory for the load location, because it expects an absolute path.


LukeUsher:
I feel like we should do the path resolution earlier in the chain; perhaps when command line arguments are first parsed since there may be other parts of the code that expect location to be absolute,

jcm93:
Agreed; changed. Seems to work fine in my testing so far.

PR #1842, opened 2025-02-18
Update sha256 for Wonder Boy in Monster World (USA, Europe) for proper eeprom assignment

remutro:
Looks like the has value we had for Wonder Boy in Monster World was outdated from the GoodGen era and was not current with the currently agreed dump of this game in No-Intro's DB. This resulted in a hang after pressing start at the main menu as EEPROM values were not being found. 

This update the sha256 hash with the currently accepted dump of this game. EEPROM values are now being found and assigned correctly and the game is now playable.

### DIFF
--- a/mia/medium/mega-drive.cpp
+++ b/mia/medium/mega-drive.cpp
@@ -615,7 +615,8 @@ auto MegaDrive::analyzeStorage(vector<u8>& rom, string hash) -> void {
   }
 
   //Wonder Boy in Monster World (USA, Europe)
-  if(hash == "8d905c863b73a1522f6ea734d630beae1b824b9eecb63f899534c02efbde20fd") {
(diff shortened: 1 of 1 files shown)


LukeUsher:
Is there a reason we are using the hash here as opposed to game id like we do for n64?

remutro:
They don't really have a game id code that is consistent like the N64 which you can trust because Nintendo made sure they were unique and valid. Sega cartridges  have a serial number field, but a lot of them seem to be garbage and controlled by the developer. For example, the serial for FIFA 96 (32x) is:   "Revision1.0". That's not a real serial Id, it's whatever junk they wanted to put in that field. If there were a real Id assigned, it would be preferred, but we don't have that here.

There are also hundreds of entries in this file all based on hash which I would imagine is all for the same reason as well.

PR #1845, opened 2025-02-19
ruby: Update to SDL3

jcm93:
Draft status pending ares-deps update and further testing.

In the input subsystem, changes are mostly cosmetic with renamed functions and joystick enumeration being slightly different.

In the audio subsystem, there are somewhat more significant changes. Audio queues are replaced by audio streams. The conversation between the API and the host to get a desired buffer size proceeds differently; the only way to request a specific buffer size is now via an SDL_HINT. It seems like the behavior here is quite platform-dependent. On macOS at least, it looks like we actually have more precise control over our buffer sizes with SDL3; buffer sizes no longer need be powers of 2. Performance seems to be fine; blocking behavior is somewhat different but not significantly.

### DIFF
--- a/cmake/finders/FindSDL.cmake
+++ b/cmake/finders/FindSDL.cmake
@@ -42,7 +42,7 @@ include(FindPackageHandleStandardArgs)
 
 find_package(PkgConfig QUIET)
 if(PKG_CONFIG_FOUND)
-  pkg_search_module(PC_SDL QUIET sdl2)
+  pkg_search_module(PC_SDL QUIET sdl3)
 endif()
 
 # SDL_set_soname: Set SONAME on imported library target
@@ -64,10 +64,18 @@ macro(SDL_set_soname)
   unset(_result)
 endmacro()
 
+find_library(
+  SDL_LIBRARY
+  NAMES SDL3 SDL3-3.0.0 SDL3-3.0
+  HINTS ${PC_SDL_LIBRARY_DIRS}
+  PATHS ${CMAKE_SOURCE_DIR}/.deps /usr/lib /usr/local/lib
+  DOC "SDL location"
+)
+
 find_path(
   SDL_INCLUDE_DIR
-  NAMES SDL.h SDL2/SDL.h
-  HINTS ${PC_SDL_INCLUDE_DIRS}
+  NAMES SDL.h SDL3/SDL.h
+  HINTS ${PC_SDL_INCLUDE_DIRS} ${SDL_LIBRARY}/..
   PATHS ${CMAKE_SOURCE_DIR}/.deps /usr/include /usr/local/include
   DOC "SDL include directory"
   # "$<$<PLATFORM_ID:Darwin>:NO_DEFAULT_PATH>"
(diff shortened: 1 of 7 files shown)


jcm93:
CI on this PR has functional builds that use SDL3; testing is welcome. This branch currently fetches updated macOS and Windows dependencies from my fork of `ares-deps`, which contains SDL3 as well as updated librashader and slang-shaders. To test on macOS and Windows, all that should be needed is re-running CMake and then rebuilding ares.

bsdcode:
Testing on FreeBSD. Audio subsystem working fine so far. Input subsystem doesn't pick up my gamepad, `SDL_GetJoysticks` returns 0 found joysticks. This seems to be a specific problem with SDL3 on FreeBSD and not related to ares. I'll report the problem to the SDL3 port maintainer on FreeBSD and to SDL upstream if appropriate.

jcm93:
Thanks for testing! Hopefully that can get figured out before long. Good to know.

jcm93:
I pushed an update as we were enumerating devices wrong which could have caused your issue, probably worth retesting again.

bsdcode:
Thanks, unfortunately this doesn't fix the issue. I can reproduce the problem with a minimal C program which just initializes the Joystick subsystem and then uses `SDL_GetJoysticks` to retrieve the number of the currently connected joysticks. An equivalent program using SDL2 works as expected. I opened a bug report on the FreeBSD bugzilla: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=284919

bsdcode:
Another user has just reported an issue with undetected gamepads on Windows upstream: https://github.com/libsdl-org/SDL/issues/12347

bsdcode:
I found the problems with the SDL3 port on FreeBSD. The build system doesn't pick up the installed dependencies and the libusb backend of the HIDAPI driver needs a small patch to work with FreeBSD's implementation of the libusb API. I submitted a patch to the maintainer. With these fixes the input system detects my gamepad and it's working fine now :)

jcm93:
Got it; thanks for chasing those down. I'm sure we can wait a bit for that to be fixed. We may or may not also want to wait for the next SDL3 release so we don't ship the 8BitDo Micro regression, since some ares users could easily have that controller.

For ares-deps on macOS and Windows we can discuss internally what to do; we could either ship an update that stays on SDL2, or just stay on the current ares-deps for the time being until the next SDL3 release.

jcm93:
Looks like a fix may be shipped for SDL3 on FreeBSD?

bsdcode:
> Looks like a fix may be shipped for SDL3 on FreeBSD?

Yes indeed. The port maintainer opened a PR upstream adressing the libusb issue (https://github.com/libsdl-org/SDL/pull/12502) and accepted my patch as a preliminary fix to the SDL3 port (https://cgit.freebsd.org/ports/commit/?id=6d5ea5b1e6a167cb9759d26df2cf32273be543aa).

So on the FreeBSD side we are safe and ready to go.

jcm93:
This should be ready to go. This and #1922 should be merged together, but the merge order should not matter.

PR #1849, opened 2025-02-20
feature: add ppu settings

ZengGengSen:
- split the emulator settings
- add ppu options

### DIFF
--- a/ares/fc/ppu/memory.cpp
+++ b/ares/fc/ppu/memory.cpp
@@ -82,7 +82,7 @@ auto PPU::writeIO(n16 address, n8 data) -> void {
   //PPUCTRL
   case 0:
     scroll.nametable = data.bit(0,1);
-    if (io.lx == 257 && rendering())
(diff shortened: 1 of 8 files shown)


LukeUsher:
We don't usually provide toggles for accuracy settings unless they have performance impact; are there hardware revisions where these issues do not occur? If so, implementing multiple PPU revisions like the SNES core does may be a better approach, if not, I'm not sure this is something that ares would or should take, but I am open to opinions.

FitzRoyX:
Some info first: the recently reported Akumajou "bug" first appeared in v142 when the PPU glitch emulation was added. if you enable "PPU first-write scroll glitch" in Mesen, you get output similar to what Ares gets.

Do all PPU revisions suffer from this bug or just some of them? I have no idea. https://www.nesdev.org/wiki/PPU_variants
But if there are, then it makes sense for it to be disabled by default the way Mesen does it.

And I agree that the settings shouldn't be here, this makes them inaccessible when a game isn't running. Would be better if they were in Configuration>Options or ini only.

PR #1851, opened 2025-02-21
ps1: gpu: Ensure hi >= lo when calling std::clamp()

thentenaar:
On Lunar Disc 1, when entering Berg (about two minutes in,) going into Alex's house or the barn to the right of it, winds up causing an assertion failure in std::clamp() on the screen transition. Could be a bug in the game where it sends the 0xe3/0xe4 coords in the wrong order, or if it's off by one calculating them. Not sure if the real hardware ignores it or does anything in particular as a result.

Adding some print() calls in ``GPU::Render::triangle()`` before the clamp, I see:
```
v0=(320,216), v1=(336,216) v2=(320,232)
vmin.x=320, x1=160 x2=159
vmin.y=216, y1=112 y2=111
vmax.x=336, x1=160 x2=159
vmax.y=232, y1=112 y2=111

/usr/lib/gcc/x86_64-pc-linux-gnu/14/include/g++-v14/bits/stl_algo.h:3626: constexpr const _Tp& std::clamp(const _Tp&, const _Tp&, const _Tp&) [with _Tp = int]: Assertion '!(__hi < __lo)' failed.
```

Thus, this patch ensures ``hi >= lo``.

### DIFF
--- a/ares/ps1/gpu/renderer.cpp
+++ b/ares/ps1/gpu/renderer.cpp
@@ -1,3 +1,5 @@
+#define clip(X, LO, HI) std::clamp((X), (((LO) <= (HI)) ? (LO) : (HI)), (((HI) >= (LO)) ? (HI) : (LO)))
+
 auto GPU::generateTables() -> void {
   static constexpr s8 table[4][4] = {
     {-4, +0, -3, +1},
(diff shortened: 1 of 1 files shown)


LukeUsher:
There's probably a more correct solution to this but it'l d for now to avoid the crash; sorry for the delay, I meant to investigate further but did not find the time just yet.

jcm93:
Just an aside, definitely not a fan of the macro function usage. Better would probably be something like:

```c++
constexpr auto clip = [](auto x, auto lo, auto hi) {
  return std::clamp((x), (((lo) <= (hi)) ? (lo) : (hi)), (((hi) >= (lo)) ? (hi) : (lo)));
};
```
Which is not only actually debuggable and devoid of macro side effects, but would also give you the ability to do things at compile time like 
```c++
static_assert(clip(3, 1, 2) == 2);
static_assert(clip(1, 2, 3) == 2);
static_assert(clip(2, 1, 3) == 2);
```
Since it's localized here and maybe likely to be rewritten, I guess it's not terribly consequential, but it would be definitely be nicer if we take advantage of modern language features like this instead of macro functions in general in the future, I guess.

PR #1856, opened 2025-02-22
gba: add option to enable Game Boy Player

png183:
Re-adds support for Game Boy Player rumble, with a UI toggle to enable/disable the feature.


LukeUsher:
Nice!

The reason this was changed initially was because the GBAVideo titles do not run under the Game Boy Player; but it is great to have it as an option.

Just a little thing: can we consider adding a 'deviceName' variable to the Emulator::GameBoyAdvance struct, so that rather than duplicated cases, we can just have

`deviceName = settings.gameBoyAdvance.player ? "Game Boy Player" : "Game Boy Advance";`

And then InputPort, GameBoyAdvance::load,  and the pac lookup can just use this string instead, avoiding the duplication?

String concat in ares is easy: `if(!ares::GameBoyAdvance::load(root, {"[Nintendo]", deviceName})) return otherError;` should work, for example

PR #1876, opened 2025-03-03
User configurable setting for N64 Controller Pak size

rasky:
Continuation of #1664 by @Cuyler36

This PR adds support for user configurable N64 Controller Pak bank sizes. The motivation for this comes from Doubutsu no Mori. A NES emulator is included which can load Famicom/NES ROM images from the Controller Pak. The standard 1 bank 32KB Controller Pak is far too small for most games. Ares is the only N64 emulator that I've found which both runs the NES emulator and can load data off of the Controller Pak.

The N64 Controller Pak can support up to 62 banks of 32KB each. Nintendo did not utilize this feature as far as I'm aware, but Datel did make a 128KB 4-bank Controller Pak. Bank switching is done by writing to the Controller Pak with address line 15 enabled. The desired bank is sent to an IC on the Controller Pak as one block of bytes (32 bytes). The IC keeps track of the selected bank when further reads and writes happen.

The default should be 1 bank or 32KB of space which preserves compatibility with existing Controller Pak files.

### DIFF
--- a/ares/n64/controller/gamepad/gamepad.cpp
+++ b/ares/n64/controller/gamepad/gamepad.cpp
@@ -12,6 +12,8 @@ Gamepad::Gamepad(Node::Port parent) {
   port->setDisconnect([&] { return disconnect(); });
   port->setSupported({"Controller Pak", "Rumble Pak", "Transfer Pak"});
 
+  bank = 0;
+
   x           = node->append<Node::Input::Axis>  ("X-Axis");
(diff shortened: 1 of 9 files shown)


rasky:
I've tested the PR and it appears to work correctly. I've fixed one bug related to the reply of standard cpaks to bank switching commands (it should be ignored without reporting an error).

I'm going to test it further for some dev work I'm doing, but I think this is still a good start that could be merged as-is.

PR #1885, opened 2025-03-08
n64: allow booting ROMs with non-standard PI DOM1 parms in the header

rasky:
All commercial ROMs in N64 have the value 0x80371240 in the first 4 bytes. This has long been used by emulators both as a magic number to "detect a N64 rom", and as a way to detect endianess of the dump.

In reality, the first byte (0x80) is unused/ignored by the console, while the other 3 (0x371240) are just parameters used to configure cartridge access (basically, ROM speed). There's absolutely nothing that prevents a ROM with other values to boot. Ares currently fails to honor this by refusing booting ROMs with different values, but this is just wrong.

To fix this, we change approach and use the CIC detection algorithm to also detect the endianess, given that it's based on a checksum match. If we can't find a match, we tried the two common types of byteswaps, hoping to find a match. Eventually, we fallback to assume it's a big-endian dump, and needs the common CIC-6102.

This means that analyze() will now only reject ROMs with size lower than 4 KiB, but that's correct. Given that 4 KiB are loaded by the IPL2, we assume that minimum size is required. Besides that, there is nothing else in the ROM that can be used to "detect" a N64 ROM, so it's correct to just let it boot and see what happens.

### DIFF
--- a/mia/medium/nintendo-64.cpp
+++ b/mia/medium/nintendo-64.cpp
@@ -4,6 +4,7 @@ struct Nintendo64 : Cartridge {
   auto load(string location) -> LoadResult override;
   auto save(string location) -> bool override;
   auto analyze(vector<u8>& rom) -> string;
+  auto cic_detect(array_view<u8> ipl3) -> string;
   auto ipl2checksum(u32 seed, array_view<u8> rom) -> u64;
 };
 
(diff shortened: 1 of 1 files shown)


mia/medium/nintendo-64.cpp
```diff
@@ -145,42 +146,95 @@ auto Nintendo64::save(string location) -> bool {
...
+        data[index + 0] = d2;
+        data[index + 1] = d3;
+        data[index + 2] = d0;
+        data[index + 3] = d1;
+      }
+
+      cic = cic_detect(array_view<u8>(&data[0x40], 0xfc0));
+      if (cic == "") {
+        //no match is found. Fallback to CIC 6102, big-endian.
+        printf("CIC FALLBACK\n");
```
sp1187: Debug print left, remove?

mia/medium/nintendo-64.cpp
```diff
@@ -145,42 +146,95 @@ auto Nintendo64::save(string location) -> bool {
...
     return {};
-  } else if((data[0] == 0x80 && data[1] == 0x37 && data[2] == 0x12 && data[3] == 0x40)
-         || (data[0] == 0x80 && data[1] == 0x27 && data[2] == 0x07 && data[3] == 0x40)) {   //64DD IPL
-    //big endian
-  } else if((data[0] == 0x37 && data[1] == 0x80 && data[2] == 0x40 && data[3] == 0x12)
-         || (data[0] == 0x27 && data[1] == 0x80 && data[2] == 0x40 && data[3] == 0x07)) {   //64DD IPL
-    //byte-swapped
+  } 
+  
+  //detect endianess of the ROM by checking the IPL3 checksum. We run the same
```
sp1187: endianess -> endianness

mia/medium/nintendo-64.cpp
rasky: done

mia/medium/nintendo-64.cpp
rasky: done

PR #1913, opened 2025-04-01
build: Enable CHD support via build option

jcm93:
Changes support for MAME's "compressed hunks of data" format to be conditional in the build, controlled by the `ARES_ENABLE_CHD` CMake build option. The option is enabled by default.

Converting CHD support into an optional build flag has two benefits. Firstly, we can produce slimmer builds more quickly with fewer dependencies if we don't need CHD support, which is only relevant for a few cores and those users with ROMs stored in the CHD format.

Secondly, if we are eyeing a move to `chd-rs` over `libchdr`, we will want to maintain support for building ares without a Rust toolchain. If we migrate to `chd-rs`, CHD support would possibly be disabled by default on Linux, so that a Rust toolchain is not required there. On macOS and Windows, ares-deps would likely handle vendoring of `chd-rs`.

### DIFF
--- a/cmake/common/helpers_common.cmake
+++ b/cmake/common/helpers_common.cmake
@@ -235,6 +235,17 @@ function(_extract_target_from_link_expression)
         set(${arg_dt_TARGET_VAR} "${gen_library}")
       endif()
     endif()
+  elseif(arg_dt_LIBRARY MATCHES "\\$<LINK_ONLY:[^>]+>")
+    # Non-specific target link expression found. Since no other expressions matched, consider the parameter following
+    # LINK_ONLY to be a CMake target.
+    string(REGEX REPLACE "\\$<LINK_ONLY:([^>]+)>" "\\1" gen_library "${arg_dt_LIBRARY}")
+
+    message(DEBUG "gen_library is ${gen_library}")
+
+    if(TARGET ${gen_library})
+      message(DEBUG "setting arg_dt_TARGET_VAR to ${gen_library}")
+      set(${arg_dt_TARGET_VAR} ${gen_library})
+    endif()
   else()
     # Unknown or unimplemented generator expression found. Abort script run to either add to ignore list or implement
     # detection.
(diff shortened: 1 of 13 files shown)


chyyran:
You could always swap between libchdr and chd-rs-capi without too much trouble. API compatibility between chd-rs-capi and libchdr is guaranteed, and ABI compatibility should be guaranteed as long as you use a suitable `libchdcorefile` implementation when building chd-rs.

As long as you don't link both libchdr **and** chd-rs-capi and mix and match their `chd_file*` pointers it should be compatible.

jcm93:
Yeah, supporting both is still definitely a viable option. It is mildly more complex than only needing to care about one or the other, and adds weight to the build logic, but might nevertheless still end up being the best path forward.

bsdcode:
Linking `tools/sourcery/sourcery` fails on FreeBSD with disabled `ARES_ENABLE_CHD`:
```
ld: error: undefined symbol: pthread_create
>>> referenced by nall.cpp
>>>               nall/nall/CMakeFiles/nall.dir/nall.cpp.o:(nall::TCP::Socket::open(unsigned int, bool))
>>> referenced by nall.cpp
>>>               nall/nall/CMakeFiles/nall.dir/nall.cpp.o:(nall::TCP::Socket::open(unsigned int, bool))
>>> referenced by nall.cpp
>>>               nall/nall/CMakeFiles/nall.dir/nall.cpp.o:(nall::TCP::Socket::open(unsigned int, bool))
```
As far as I know, other platforms (e.g. Linux) provide pthread symbols in libc, but this is not the case for FreeBSD. It worked before and with enabled `ARES_ENABLE_CHD` because `thirdparty/libchdr/deps/zstd-1.5.6/...` sets the dependency on pthread for linking.

Adding
```
find_package(Threads REQUIRED)
target_link_libraries(nall PRIVATE Threads::Threads)
```
to `nall/nall/cmake/os-freebsd.cmake` fixes this linker error on FreeBSD. But maybe it is a better idea to set this unconditionally in `nall/nall/CMakeLists.txt` because nall depends on thread support. What do you think?

jcm93:
Thanks for testing. I did check my FreeBSD VM and didn't run into the same issue for some reason. Regardless, you're correct that nall.cpp needs Threads, so we probably should indeed be explicitly defining the relationship there. Will update.

PR #1918, opened 2025-04-04
ym2612: fix LFO AM enabled bit not functioning

### LINKED ISSUE #1910: [Mega Drive] Kid Chameleon - Audio issues from v142 update
starting with the v142 update, some of the sound and music in the game sounds very "wrong" and "off" from how it's supposed to sound. this is immediately noticeable from the first note of the title theme upon booting the game. the electric guitar sound is all warbly and weird now. i've played this game many times on many different platforms and it never sounded this weird, so i'm fairly certain the game is not supposed to sound like this.

```
game
  sha256: bc08b0188fb715ae469806843a8418d4937d2098b66d04227668865b3d4d5baa
  name:   Kid Chameleon (USA, Europe)
  title:  Kid Chameleon (USA, Europe)
  label:  KID CHAMELEON
  label:  KID CHAMELEON
  serial: GM MK-1010 -00
  region: NTSC-J, NTSC-U, PAL
  board
    memory
      type: ROM
      size: 0x100000
      content: Program

```

jsgroth:
This should fix #1910 (incorrect instrument sounds in Kid Chameleon).

This also renames the `lfoEnable` field to `tremoloEnable` to more accurately reflect what it does, since it has no effect on LFO FM / vibrato.

### DIFF
--- a/ares/component/audio/ym2612/channel.cpp
+++ b/ares/component/audio/ym2612/channel.cpp
@@ -125,7 +125,7 @@ auto YM2612::Channel::Operator::updatePhase() -> void {
 
 auto YM2612::Channel::Operator::updateLevel() -> void {
   u32 lfo = ym2612.lfo.clock & 0x40 ? ym2612.lfo.clock & 0x3f : ~ym2612.lfo.clock & 0x3f;
-  u32 depth = tremolos[channel.tremolo];
+  u32 depth = tremolos[tremoloEnable * channel.tremolo];
 
   bool invert = ssg.attack != ssg.invert && envelope.state != Release;
   n10 value = ssg.enable && invert ? 0x200 - envelope.value : 0 + envelope.value;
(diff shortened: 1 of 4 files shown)


jsgroth:
First notes of Kid Chameleon before this change:
[kid-chameleon-before.webm](https://github.com/user-attachments/assets/4fab49c7-41ad-441e-9bef-67761a009182)

After:
[kid-chameleon-after.webm](https://github.com/user-attachments/assets/118b0c23-d918-49a9-8c50-050854bbab51)

FitzRoyX:
Cool, thanks for the fix! Might be a good idea to get @TascoDLX approval of this since he has probably looked at this code the most.

remutro:
Since this changes serialization, you will need to bump the serialization version in md/system/serialization.cpp, first line at the top of the file. You can either increase the decimal number, or set it to v143 or v144 at this point. But it has to be incremented in a meaningful way, otherwise it will attempt to load older save states and fail.

Update: submitted PR to bump serialization version number for this core.

jsgroth:
@remutro I don't believe this is an incompatible change to the save state serialization format since it is only a field rename and doesn't change the field type or the serialization field ordering. I tested by making a save state in a version with these changes and then trying to load it in a version without these changes, and it loaded successfully. Doing the reverse also works, saving state in older version and loading in newer version.

Edit: But either way, bumping the serialization version doesn't hurt in case I am wrong.

PR #1922, opened 2025-04-10
Update ares-deps to 2025-04-09 release

jcm93:
Updates the dynamic dependencies used by ares.

Includes:

* Stop vendoring the librashader header on Linux
* deps.windows: Update librashader to 0.6.3
* deps.macos: Update librashader to 0.6.3
* Update slang-shaders to ref 25311dc
* Pin Rust nightly to version contemporary with librashader version
* Establish macOS deployment targets
* deps.macos: Update SDL to 3.2.10
* deps.windows: Update SDL to 3.2.10

> [!NOTE]
> This will cause ares on macOS and Windows to stop successfully locating SDL from ares-deps when configuring until the merger of #1845. This and #1845 should be merged together, but the order does not matter.

### DIFF
--- a/deps.json
+++ b/deps.json
@@ -1,14 +1,14 @@
 {
     "dependencies": {
         "prebuilt": {
-            "version": "2024-12-30",
+            "version": "2025-04-09",
             "baseUrl": "https://github.com/ares-emulator/ares-deps/releases/download",
             "label": "Pre-Built ares-deps",
             "hashes": {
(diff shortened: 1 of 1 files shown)


jcm93:
Because of changes to SDL packaging, macOS and Windows users may need to freshly re-generate the existing dev environment with `cmake --preset <preset> --fresh` after this is merged.

PR #1946, opened 2025-04-20
hiro: Use GTK native file dialogs for portal compatibility

jcm93:
Per [flatpak documentation](https://docs.flatpak.org/en/latest/portals.html), if we want to use portals to grant scoped filesystem access based on files the user has opened, rather than blanket host filesystem access, we need to use GTK's 'native' file dialog.

Since the native file dialog does not seem to be available under GTK2, we put this functionality in a new file guarded behind HIRO_GTK=3 for the sake of backwards compatibility (note though that GTK2 has still has not been tested by myself under the new build system and isn't officially supported).

Verified that these dialogs work for picking files and folders under KDE on Fedora 41. This could probably use a review from another everyday Linux user though.

### DIFF
--- a/hiro/gtk/browser-window-native.cpp
+++ b/hiro/gtk/browser-window-native.cpp
@@ -1,0 +1,92 @@
+#if defined(Hiro_BrowserWindow)
+
+namespace hiro {
+
+static auto BrowserWindow_addFilters(GtkFileChooserNative* dialog, vector<string> filters) -> void {
+  for(auto& filter : filters) {
+    auto part = filter.split("|", 1L);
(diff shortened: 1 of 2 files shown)


bsdcode:
Not your everyday Linux user, but the dialogs work fine under i3 on FreeBSD ;)

remutro:
I would recommend @HeroponRikiBestest  perhaps give this a try and validate it works as expected. They are our resident mostly Linux tester and regularly comes across Linux specific issues so perhaps when they get a change they can give this PR a test to make sure there are no regressions from this merge?

PR #1948, opened 2025-04-21
ws: support 8-bit bus width ROMs

asiekierka:


ares/ws/cartridge/cartridge.hpp
```diff
@@ -32,6 +32,7 @@ struct Cartridge : IO, Thread {
     n1 rtc;
     n1 flash;
     n1 karnak;
+    n1 wordWidth;
```
remutro, line 35: I believe this needs to be added to the serialization file and have the serialization bumped for the WonderSwan.

remutro (dismissed):

ares/ws/cartridge/cartridge.hpp
asiekierka: No - it's always derived from the cartridge `.ws`/`.wsc` file, so these are not serialized/deserialized.

PR #1950, opened 2025-04-22
ps1: gpu: macro -> constexpr

thentenaar:
Although I have no aversion to macros personally, this is as suggested by @jcm93 in https://github.com/ares-emulator/ares/pull/1851#issuecomment-2821853853

### DIFF
--- a/ares/ps1/gpu/renderer.cpp
+++ b/ares/ps1/gpu/renderer.cpp
@@ -1,4 +1,7 @@
-#define clip(X, LO, HI) std::clamp((X), (((LO) <= (HI)) ? (LO) : (HI)), (((HI) >= (LO)) ? (HI) : (LO)))
+
+constexpr auto clip = [](auto x, auto lo, auto hi) {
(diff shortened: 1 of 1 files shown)


jcm93:
Thanks for following up with this.

Just a nit: indentation.

ares/ps1/gpu/renderer.cpp
```diff
@@ -1,4 +1,7 @@
-#define clip(X, LO, HI) std::clamp((X), (((LO) <= (HI)) ? (LO) : (HI)), (((HI) >= (LO)) ? (HI) : (LO)))
+
+constexpr auto clip = [](auto x, auto lo, auto hi) {
+	return std::clamp(x, lo <= hi ? lo : hi, hi >= lo ? hi : lo);
```
jcm93:
```suggestion
  return std::clamp(x, lo <= hi ? lo : hi, hi >= lo ? hi : lo);
```
See above.

PR #1995, opened 2025-05-04
build: Improve version handling

jcm93:
ares' build system tries to always have version information available for debuggability and build tracking. It depends upon the presence of Git to provide version information, since Git is the version control system used by the project.

When distributing tarballs of ares, however, no `.git` directory and no inherent version information is included in-tree, which causes the build to fail with a vague fatal error after failing to derive a version for itself.

Since we would like building a tarball to be at least possible with minimal fuss, we should provide some version information inside a tar archive served by GitHub. This PR uses Git's [export-subst](https://git-scm.com/docs/gitattributes#_export_subst) attribute to automatically bundle version information when `git-archive` is run against the repository, as is done when the GitHub API serves a tar archive of a given release, branch or commit. The build system will check for the presence of this file if it finds that there is no `.git` directory at the root of the repository.

This PR also improves the version handling in general to better support semantic version strings of the form `v123.x.y`, in case future releases of ares require hotfixes or minor versions.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# When archiving, export a version file containing tag/commit information
+cmake/common/.archive-version export-subst
--- a/cmake/common/.archive-version
+++ b/cmake/common/.archive-version
@@ -1,0 +1,1 @@
+$Format:%(describe:tags,exclude=nightly)$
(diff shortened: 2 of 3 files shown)


LukeUsher:
good to have tarballs functioning again.

PR #2006, opened 2025-05-08
build: Don't require tags, add unofficial versioning

jcm93:
### Unofficial versioning

Distinguish between "official" and "unofficial" versioning for builds. Official version strings are used for commits within the master branch; unofficial version strings are the default anywhere else.

  * Official versions use the unvarnished output of `git describe`, consisting of solely the tag for tagged versions, and derivatives of the tag for nightlies and elsewhere - e.g. `ares v144-14-g1a79a0aae`.

  * Unofficial versions use the format `<ref name> - <commit shorthash>(-modified)`, e.g. `build/version-fix-2 - fba4aa9`

This accomplishes two objectives: Disambiguate unofficial builds in more cases, and eliminate the annoyance of outdated version strings on forks with stale or no tags.

Unofficial builds will also now succeed even if there are no tags found.

### Canonical versioning

This PR also attempts to better define a canonical versioning system. Canonical versioning tries to extract an idiomatic version string (three nonnegative dot-separated integers, optionally followed by `-` and other information) from the latest tag matching the format "v[0-9]*". For example, the tag v144 becomes the canonical version `0.144.0`.

While this canonical version string is not displayed within ares, defining it is done to try to conform to best practices across platforms, and to lay the foundation for future work where we may depend on well-defined versioning for functionality or external use. Some examples include autoupdates or exporting ares targets as libraries. Platforms like macOS also generally expect and internally use these types of version strings.

This PR uses `0.144.0` rather than `144.0.0` because ares doesn't have a public API and there are no immediate plans to introduce one. A version below 1.0.0 is generally regarded to mean "things can change at any time".

This PR sets official builds to require a canonical version string to be able to be derived; unofficial ones do not. Currently this is done only to catch errors where tag information is not present when cutting releases.

### Version bundling

Lastly, this PR improves the handling of version information in artifact metadata (Windows resource files and macOS plists). The Windows resource file is rewritten to include the displayed version string, as well as program information and copyright. macOS plists were intended to always contain version information, but inadvertently did not due to an error, which this PR corrects.


Screwtapello:
> This PR uses 0.144.0 rather than 144.0.0 because ares doesn't have a public API and there are no immediate plans to introduce one. A version below 1.0.0 is generally regarded to mean "things can change at any time".

A note on bsnes' historical versioning: the first surviving version [seems to be](https://gitlab.com/higan/bsnes-history-kit/-/blob/934d5581e32832a160d071879de30dafdc264948/sources/tukuyomi/misc/bsnes_changelog.txt) v0.0.002. After v0.0.005, Near dropped the first "0." releasing v0.006. After v0.059, Near dropped the next "0." and released v060. This numbering continued through higan and ares, all the way to v144. Other widely-used projects using this versioning system include [less](https://www.greenwoodsoftware.com/less/) and [xterm](https://invisible-island.net/xterm/xterm.log.html).

As you note, SemVer isn't very useful for a project like ares that doesn't provide any kind of API, but if you wanted to map the version number into SemVer for familiarity's sake, I suggest "v0.0.144" (the original scheme), "v144.0.0" (Near's chosen scheme) or "v1.44.0" (how end-users typically misinterpret Near's scheme) would probably be more appropriate than "v0.144.0".

EDIT: That said, of course the ares project can decide to use whatever versioning scheme they like, they don't have to be stuck with the one they inherited.

jcm93:
FWIW @bsdcode, packaging v145, you will (probably) want to pass `-DARES_BUILD_OFFICIAL=YES` when configuring ares to get a version string based on the tag. An unofficial version string will be based on the branch and include extraneous info.

bsdcode:
Thank you very much for the info. I'm now using it in my local WIP port update, see https://github.com/bsdcode/freebsd-ports/tree/ports.

I also encountered a linker error with enabled `ARES_BUILD_OPTIONAL_TARGETS` option if Game Boy Advance, Genesis, Neo Geo and SNES are unset. Will open an issue for this.

PR #2018, opened 2025-05-16
Added ability to copy TextView via CMD+C

ScraperMan2002:
Previously, even if you highlight the text in the Properties tool you can't use CMD+C to copy the actual text. You had to copy manually via right clicking and selecting copy. Now, with these changes, you can use the shortcut CMD+C


ScraperMan2002:
Actually, I changed my mind. Please also help me remove all instances of "TestAddCode" because it's not right. It's a personal code to test if my coding works or not.

ScraperMan2002:
Oh and shoutout to yam, this person helped me code a lot of the update.

PR #2023, opened 2025-05-17
Added CMD+C and CMD+A capabilities to Ares

ScraperMan2002:
On MacOS, previously highlighting text on the Properties TextView wouldn't allow you to copy them with CMD+C and clicking on the text and pressing CMD+A wouldn't allow you to select all of the text. This solves the problem.

P.S.: in the future perhaps we should consider moving the code that handles menu item creation to the cocoa/Application.cpp portion. This way the window wouldn't be responsible for generating menu items.


.gitignore
```diff
@@ -48,3 +48,4 @@ tests/m68000/tests
 # Exclude unused legacy zlib header that is renamed by CMake
 thirdparty/libchdr/deps/zlib-*/zconf.h
 thirdparty/libchdr/deps/zlib-*/zconf.h.included
+ares/build*
```
jcm93: Is there some reason you found this to be necessary? The documentation and build system recommend to not build ares in-tree; rather, you should establish a build directory at the base of the repository and invoke CMake from that location. Using the `macos` CMake preset from the root of the repository will have the same effect and should not make any further exclusions in the .gitignore necessary.

ares/ares/.DS_Store
jcm93, line 1: A recent commit to ares master made these .DS_Store files ignored by version control. I would recommend re-staging your work so that these files are not included.

hiro/cocoa/widget/text-edit.cpp
```diff
@@ -17,6 +17,15 @@
   return self;
 }
 
+- (BOOL) validateMenuItem:(NSMenuItem *) menuItem {
```
jcm93: It turns out that this validation implementation is not actually necessary; the NSTextView control itself provides a default implementation and will automatically validate typical actions present in the Edit menu.

hiro/cocoa/widget/text-edit.hpp
```diff
@@ -1,10 +1,11 @@
 #if defined(Hiro_TextEdit)
 
-@interface CocoaTextEdit : NSScrollView <NSTextViewDelegate> {
+@interface CocoaTextEdit : NSScrollView <NSTextViewDelegate, NSMenuItemValidation> {
```
jcm93: This protocol conformance isn't necessary, per above; both changes to this file can be removed.

hiro/cocoa/window.cpp
```diff
@@ -76,6 +76,19 @@
     item.keyEquivalent = @"q";
     [rootMenu addItem:item];
 
+    item = [[NSMenuItem alloc] initWithTitle:[NSString stringWithFormat:@"Edit"] action:nil keyEquivalent:@""];
```
jcm93: As discussed off-thread, if you are willing to take this on, I think it may be better to move the entire handling for menu creation up to the level of `hiro/cocoa/application.cpp`. By being here, the same redundant work is performed four times for no benefit. If you don't want to handle this now, it's probably fine to just make your changes here and fix that later.

hiro/cocoa/window.cpp
```diff
@@ -76,6 +76,19 @@
...
+    [menuBar addItem:item];
+      
+    item = [[NSMenuItem alloc] initWithTitle:[NSString stringWithFormat:@"Copy"] action:@selector(copy:) keyEquivalent:@"c"];
+    item.keyEquivalentModifierMask = NSEventModifierFlagCommand;
+    [editMenu addItem:item];
+
+    item = [[NSMenuItem alloc] initWithTitle:[NSString stringWithFormat:@"Select All"] action:@selector(selectAll:) keyEquivalent:@"a"];
+    item.keyEquivalentModifierMask = NSEventModifierFlagCommand;
+    [editMenu addItem:item];
+
```
jcm93: I would recommend creating items for other the standard Edit commands, Cut and Paste. One area these could already be used is the memory editor.

The typical ordering for applications is Cut, Copy, Paste, and lastly Select All.

hiro/cocoa/window.hpp
```diff
@@ -1,10 +1,11 @@
 #if defined(Hiro_Window)
 
-@interface CocoaWindow : NSWindow <NSWindowDelegate> {
+@interface CocoaWindow : NSWindow <NSWindowDelegate, NSMenuDelegate> {
```
jcm93: This protocol conformance is not necessary and can be removed.

jcm93 (changes requested):
Would like to see the remaining standard Edit menu commands (Cut/Paste) implemented and other mentioned changes made. Refactoring would be a bonus but that could be deferred to later.

Ideally the commits would be cleaned up, rebased and squashed as well, but as long as the final diff is appropriate, a maintainer may be able to squash for you.

.gitignore
ScraperMan2002: I found this necessary because I personally build ares via a build folder like build, sometimes build_macos, and I don't want those folders to get included in my git pushes.

ares/ares/.DS_Store
ScraperMan2002: I see. I'm also very new to GitHub so I don't know much of anything. Restaging huh?

.gitignore
jcm93: Build folders are already excluded in the gitignore. You have added an exclusion for a build directory inside the ares subdirectory of the source tree, not the root of the repository.

hiro/cocoa/widget/text-edit.cpp
ScraperMan2002: I understand what you mean, but because I am that unfamiliar with Cocoa, I don't know where exactly to put this all in Applications.cpp. I get your intentions, but I don't know the method. That's all.

.gitignore
ScraperMan2002: I see.

hiro/cocoa/widget/text-edit.cpp
ScraperMan2002: Not only that, I also would like to know whether the same should be done with the other GUI makers like QT, windows, etc.

jcm93: This comment thread is about the validation method which can be removed. I think you meant to reply to a different comment.

If you don't want to refactor the menu handling into `application.cpp` that's fine, it can be done later.

This menu handling as far as I know is all specific to Cocoa; I don't think you need to worry about other backends with any of this functionality.

hiro/cocoa/window.cpp
ScraperMan2002: Sorry, the previous comment I made was at the wrong location. This layout can confuse first timers.

I understand what you mean, but because I am that unfamiliar with Cocoa, I don't know where exactly to put this all in Applications.cpp. I get your intentions, but I don't know the method. That's all.

PR #2024, opened 2025-05-17
hiro/cocoa: Add Edit Menu, refactor menu creation

ScraperMan2002:
I added the Cut, Copy, Paste, Select All in the Edit menu item, and I also created their corresponding shortcuts. I also moved menu intialization away from Windows to Application, but I must warn you: I have no idea at all why all of this works, this is like Black Magic to me. It does work, but I must warn you that there may be bugs but I'm too tired and too broken currently to find all of them. The person monitoring this pull request has been warned.


hiro/cocoa/application.cpp
```diff
@@ -115,6 +212,7 @@ auto pApplication::initialize() -> void {
     cocoaDelegate = [[CocoaDelegate alloc] init];
     [NSApp setDelegate:cocoaDelegate];
   }
+  [cocoaDelegate constructMenu];
```
jcm93: This should probably go inside the autorelease pool block immediately above.

hiro/cocoa/application.cpp
```diff
@@ -2,6 +2,103 @@
...
+    [NSApp setServicesMenu:servicesMenu];
+
+    item = [[NSMenuItem alloc] initWithTitle:[NSString stringWithFormat:@"Hide %@", applicationName] action:@selector(hide:) keyEquivalent:@""];
+    [item setTarget:NSApp];
+    item.keyEquivalentModifierMask = NSEventModifierFlagCommand;
+    item.keyEquivalent = @"h";
+    [rootMenu addItem:item];
+
+    item = [[NSMenuItem alloc] initWithTitle:@"Hide Others" action:@selector(hideOtherApplications:) keyEquivalent:@""];
+    [item setTarget:NSApp];
```
jcm93, line 48: Redundant; one of these lines can be removed.

jcm93 (changes requested):
This looks basically fine to me pending the notes below. We also need to get rid of those .DS_Store files still.

Ideally also the commits would all be squashed together into one with a cleaner message, but that might be able to be handled when/if merged.

hiro/cocoa/application.cpp
ScraperMan2002: I see. I also see that the system does these checks as well. I guess they see whether or not the Github commit works well and fulfills some sort of requirement, but I see mine failing like Ubuntu, but I don't even know why. I didn't even touch any of the stuff in Ubuntu, yet it is still failing. This seems a bit confusing.

hiro/cocoa/application.cpp
ScraperMan2002: You mean "[item setTarget:NSApp];" right?

jcm93 (approved):
Should be fine. Would be preferable if delivered as one commit with a proper title, such as:

`hiro/cocoa: Add Edit Menu, refactor menu creation`

but that's up to maintainers whether to accept as is.

LukeUsher:
Collapsed into one commit with a squash merge

PR #2026, opened 2025-05-18
n64: Improve labels for internal resolution upscaling options

thekovic:
The SD/HD/UHD options represent scaling the internal N64 resolution to 1x/2x/4x. Even for a title running at 640x480, this results in 1280x960 and 2560x1920 which doesn't match the meaning of HD/UHD. Games running at lower resolutions give even more out of whack results. New label communicates more explicitly what the option actually does.

I didn't change the string used to serialize this option in `settings.bml`. Feedback welcome on whether I should do that.

### DIFF
--- a/desktop-ui/settings/video.cpp
+++ b/desktop-ui/settings/video.cpp
@@ -92,23 +92,23 @@ auto VideoSettings::construct() -> void {
   weaveDeinterlacingLayout.setAlignment(1).setPadding(12_sx, 0);
   weaveDeinterlacingHint.setText("Doubles the perceived vertical resolution; incompatible with supersampling").setFont(Font().setSize(7.0)).setForegroundColor(SystemColor::Sublabel);
 
-  renderQualitySD.setText("SD Quality").onActivate([&] {
+  renderQuality1x.setText("1x Native").onActivate([&] {
     settings.video.quality = "SD";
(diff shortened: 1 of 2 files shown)


desktop-ui/settings/video.cpp
```diff
@@ -92,23 +92,23 @@ auto VideoSettings::construct() -> void {
   weaveDeinterlacingLayout.setAlignment(1).setPadding(12_sx, 0);
   weaveDeinterlacingHint.setText("Doubles the perceived vertical resolution; incompatible with supersampling").setFont(Font().setSize(7.0)).setForegroundColor(SystemColor::Sublabel);
 
-  renderQualitySD.setText("SD Quality").onActivate([&] {
+  renderQuality1x.setText("1x Native").onActivate([&] {
```
jcm93, line 95: Is there a reason for this to be "1x Native" rather than just "Native"?

thekovic: 1. Symmetry with the other options.
2. Absolute explicitness. Even in case someone doesn't understand the meaning of the term "native", they can understand the "1x".

jcm93: I personally think that the meaning of "native" by itself would be clear from context, being right next to "2x native" and "4x native", and that "native" by itself is more elegant/simple, but that's just me.

PR #2035, opened 2025-05-25
Changed Configuration Window size so text doesn't get cut off, and added memory editor ability to swich between binary, oct, and hex view

ScraperMan2002:
Screenshot 2025-05-25 at 13 32 17

Originally, as per this image on MacOS, important text detailing the function of an option gets cut off, and you aren't allowed to resize the Configuration window. So the size of the window was changed so such text doesn't get cut off anymore.


ScraperMan2002:
I also added memory editor ability to swich between binary, oct, and hex view. This is only available currently on MacOS version, so I don't know what else to do. I need someone to help me with the other GUI versions.

PR #2061, opened 2025-06-24
ps1: fix dma interrupt flag calculation

invertego:
The topmost bit of the DICR (DMA Interrupt) register should not influenced by the per-channel interrupt enable bits. This flaw was masked in fast boot mode, as it allows title code to begin execution in a pristine state unlike what the BIOS leaves behind.

With this change, the titles that were "fixed" by enabling fast boot mode can now be launched with it disabled.


invertego:
Reference: https://github.com/psx-spx/psx-spx.github.io/pull/71

LukeUsher:
Nice spot!

Issue #1892, opened 2025-03-10
[GBA] [Feature Request] Real Time Clock Support for games that require them to function.


ExperimentT020:
**Is your feature request related to a problem? Please describe.**
Currently Ares is completely lacking Real Time Clock support for GBA titles that expect it to be present for either specific features, ingame progression or for saving purposes. Several titles are affected to different levels depending on how they use the RTC, with some beatable but others completely unplayable without it.

**Describe the solution you'd like**
Implementing S-3511A RTC support for EEPROM, FLASH 512K and FLASH 1M save types that request it, using the 4 digit game ID to determine which titles require it.

**Describe alternatives you've considered**
The only alternative is to use another GBA emulator such as mGBA, NanoBoy Advance, Mesen, SkyEmu, NO$GBA or VBA-M for RTC titles.

**Additional context**
Affected Titles:

**EEPROM: Uses external RTC with extra IO**
- Boktai - The Sun Is in Your Hand/ Bokura no Taiyō: Solar Action RPG (All Versions) (In tandem with the UV Sensor)
- Boktai 2 - Solar Boy Django/Zoku Bokura no Taiyō: Taiyō Shōnen Django (All Versions) (In tandem with the UV Sensor)
- Shin Bokura no Taiyō: Gyakushū no Sabata (In tandem with the UV Sensor)

**- Flash 512K + RTC**
- Rockman EXE 4.5 Real Operation (Required for game progression using a ingame calendar system) (See #1883) 

**- Flash 1M + RTC**

- Pokemon Emerald (All Versions/Revisions) (Used for ingame events, evolutions, RNG generation, berry farming)
- Pokemon Ruby (All Versions/Revisions) (Used for ingame events, evolutions, RNG generation, berry farming)
- Pokemon Saphhire (All Versions/Revisions) (Used for ingame events, evolutions, RNG generation, berry farming)
- Sennen Kazoku (All Revisions) (Required for saving and ingame progression) (See #1890) 

RTC data sheet: https://www.rockby.com.au/DSheets/10110.pdf

remutro:
CasualPokePlayer provided a link to a cleaner version of the PDF:  https://www.rockby.com.au/DSheets/10110.pdf

ExperimentT020:
RTC support now implemented. Closing issue.

PR #2072, opened 2025-06-28
gba: initial RTC implementation

png183:
Adds RTC support to Game Boy Advance cartridges. While the GBA RTC chip is identical to the one already emulated by the WonderSwan core, both consoles use very different interfaces, so for now they are implemented separately. Currently marked as draft, pending testing and feedback.


ExperimentT020:
For those wishing to test: https://github.com/ares-emulator/ares/issues/1892 has the shortlist of titles that require RTC support and what they use it for.

asiekierka:
Ideally, the WonderSwan's interface *should* be a higher-level abstraction on what the GBA uses - the 384 KHz cartridge clock is used to write bits in one at a time, after all. So if you're working on this level, I'd move to make the RTC module itself an `s3511a` shareable component.

png183:
> Ideally, the WonderSwan's interface _should_ be a higher-level abstraction on what the GBA uses - the 384 KHz cartridge clock is used to write bits in one at a time, after all. So if you're working on this level, I'd move to make the RTC module itself an `s3511a` shareable component.

Addressed as of most recent commit

Issue #2068, opened 2025-06-27
[PS1] Marionette Company 2 Chu (Japan) - Missing graphics during dialogue


thelastangryman1907:
Marionette Company 2 Chu (Japan) has a 3D rotating gear effect during dialogue scenes. It's missing from Ares and as far as I know doesn't work in any emulator. It doesn't even work with PS2 backwards compatibility. 

Here's pictures from real hardware.

![Image](https://github.com/user-attachments/assets/c96bd621-02b0-46d4-9123-0bbd5ce46b0c)

![Image](https://github.com/user-attachments/assets/c3f5fdab-5a03-482a-ae25-72eb8542b088)

Here's a picture of my TV running it on real hardware for good measure.

![Image](https://github.com/user-attachments/assets/6312701c-5697-4db9-85e0-5d300c6f0c32)

And here's Ares.

![Image](https://github.com/user-attachments/assets/fbc73895-60a3-491d-a13c-2482df0502c0)

![Image](https://github.com/user-attachments/assets/95e696e1-5300-4d17-a6bb-667691f8965d)

 name:   Marionette Company 2 Chu! (Japan)
  title:  Marionette Company 2 Chu! (Japan)
  region: NTSC-J

invertego:
The function at 0x80040f0c is responsible for drawing the gear polygon. It won't actually draw unless some of the bytes in the range 0x1f8001b0-0x1f8001bf are nonzero. Some other rendering functions write to that range, but none of them are actually called by the title, at least as far as entering gameplay, from what I can see.

My guess is that the title is accidentally relying on the power-on state of scratchpad memory, which on real hardware is random. Most emulators (and I'm guessing the PS2) will zero-initialize most memory for consistency.

invertego:
Fixed with #2076

PR #2076, opened 2025-06-29
ps1: randomly initialize memory on boot

invertego:
Randomly initialize CPU RAM, CPU I-cache, CPU scratchpad, GPU VRAM, and SPU RAM.


invertego:
Motivated by https://github.com/ares-emulator/ares/issues/2068#issuecomment-3016407207

Opening as a draft because there is room for discussion on how to best approach this. Some questions worth considering:
- Should we initialize more than the scratchpad?
- Should we initialize with a nonzero constant instead of random bytes?
- Should we use a constant random seed for reproducibility?

As written, this PR takes a maximalist approach, initializing all types of memory with a dynamic seed.

asiekierka:
The last time I had a discussion about random memory initialization, the consensus was "fixed reproducible state in homebrew mode, mimic console behaviour in non-homebrew mode".

thelastangryman1907:
@asiekierka Shouldn't homebrew also just do what the console does though? If you develop things based on emulators instead of real hardware they break when the emulator gets more accurate. Which is why half of the old SNES romhacks don't work.

rasky:
Homebrew mode is a homebrew developing / debugging 
mode. Having a fixed seed in homebrew mode means that you still get random memory contents, just that they are fixed at each run. When you develop, your memory layout changes and shifts often as you add and remove variables and the linker rearranges the symbols so you will still see seemingly random behaviors during iteration cycles if you come to rely to uninitialized memory. But if you instead debug using gdb or a builtin debugger (when Ares has one), you'll see the same behavior which helps debugging.

So I still think it's the best balance for developers.

@invertego just to double check, is it known / tested that all those memory areas have random contents? Was the randomness ever dumped and inspected? I'm asking because on N64, RDRAM has random contents but those random contents are "mostly" long sequences of 1s or 0s, which is quite different from what a PRNG generates

asiekierka:
> Shouldn't homebrew also just do what the console does though?

Yes, which is why you're expected to *not* run homebrew in Homebrew Mode, as Rasky said.

Screwtapello:
I'm not sure about ares, but at least bsnes has an "Entropy" option which can be set to:

- None: always returns 0
- High: full PRNG (seems to be PCG-based)
- Low: like high, but initialises arrays with a lower-entropy pattern intended to model the power-on state of SNES RAM
    - this is the default

The RNG is always seeded from `clock()` in the C standard library.

Screwtapello:
Actually looking at the PR now, I see ares *does* have a similar thing and you're using it in "High" mode. Probably all RAM should be randomised at power-on, and the "Low" mode would probably be a good default.

invertego:
I've been asking around, but some original research may be required if we want to more closely match original hardware with something more accurate than uniform randomness.

I did give low entropy mode a try, and the output is extremely regular. Visualized as 24bpp RGB, it looks like this:

low_entropy

invertego:
The latest commit uses a constant random seed in homebrew mode - still random, but reproducibly so. The specific constants I put in came from the reference implementation of PCG (the PRNG ares is using).

LukeUsher:
This is fine :tm:

Titles shouldn't be relying on memory state anyway, and given the amount of iterations the ps1 went through, the chances are there are differences in behaviour between models anyway. so some randomness is expected.

PR #2085, opened 2025-07-06
Add "Await GDB Client" boot option

Dragorn421:
This PR adds a "Await GDB Client" option located at `Settings > Boot Options > Await GDB Client`

When enabled, this option pauses the program on startup (like the "Launch Tracer" (renamed from "Launch Debugger" in this PR) boot option), then automatically unpauses when a gdb client connects

The purpose is to allow debugging a program "from the beginning" rather than gdb connecting after the program already started.

Feel free to let me know if this can be implemented better!

discussed on discord from https://discord.com/channels/976404869386747954/976463759935696977/1391268016959913994

### DIFF
--- a/desktop-ui/presentation/presentation.cpp
+++ b/desktop-ui/presentation/presentation.cpp
@@ -77,6 +77,9 @@ Presentation::Presentation() {
   launchDebugger.setText("Launch Debugger").setChecked(settings.boot.debugger).onToggle([&] {
     settings.boot.debugger = launchDebugger.checked();
   });
+  waitGDB.setText("Wait GDB").setChecked(settings.boot.waitGDB).onToggle([&] {
+    settings.boot.waitGDB = waitGDB.checked();
+  });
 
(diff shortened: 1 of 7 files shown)


jcm93:
I think naming might be able to be a bit clearer here?

Maybe the "Launch Debugger" option should be renamed as a part of this, since it predates the gdb stub and is a bit of a misnomer now anyway ("Launch Tracer")?

And just grammatically, we may want to keep things correct/consistent and describe the action that enabling the setting does; rather than "wait GDB," maybe "Await GDB Client"? or "Await GDB Connection"?

Dragorn421:
We could also rework "launch debugger" and "wait gdb" into three checkboxes:

- Pause on startup
- Open tracers on startup
- Resume on GDB connected

?

jcm93:
I personally think that, since these are menu options rather than normal options inside a configuration window, we should keep the state as simple as possible, and not have any interdependencies.

jshield:
@jcm93 can we get this merged?

PR #2086, opened 2025-07-06
desktop-ui: Threading behavior improvements

jcm93:
Generally improves performance and UI responsiveness on Windows, as well as efficiency on all platforms. Also fixes a possible data race when modifying some emulator color settings.

### Details

The first iteration of the threading rework used a synchronization strategy where a thread that wanted to modify the emulator state would simply try to acquire a lock on a single mutex owned by the emulator worker thread. The emulator worker thread would `usleep(1)` (`Sleep(0)` on Win32) at the beginning of each run loop to allow other threads an opportunity to grab the mutex and perform necessary work.

On Linux and macOS, this strategy was sufficient, if primitive; the UI thread generally did not have issues acquiring the mutex when it was ceded. On Windows, the story is different. `Sleep(0)` guarantees that a thread will give up its time to other threads on the system that are ready to run. It provides no guarantees about what threads those might be.

One of these threads would be the main thread waiting to acquire the program mutex. In practice, however, the main thread would not reliably acquire the mutex and be able to perform its work in a timely fashion. This would manifest as long delays when, for example, dragging and dropping a title onto the ares window; or assigning inputs, unloading titles, changing drivers, etc. A more complex/robust synchronization strategy was ultimately needed to provide more guarantees that threads that need to modify the program state would be able to perform work immediately when needed.

### `Program::Guard`

This PR introduces a new ares-specific synchronization object, designed around the use of a condition variable and other primitives inside `Program`, to manage "interrupts" to the emulator run loop that will affect its state.

Rather than non-program threads simply 'busy-waiting' to acquire the program mutex, the new sequence of events goes something like this:

1. A thread that wants to modify the emulator program state (Thread A) creates a `Program::Guard` object. Creating a guard sets the `interruptWaiting` atomic boolean to true, and uses the `programConditionVariable` condition variable to wait on the condition where another boolean, `interruptWorking`, is true.
2. At the beginning of its next run loop, the emulator thread will check if `interruptWaiting` is true, and if so, it will (safely) set `interruptWorking` to true. It will then signal the `programConditionVariable`, and initiate a new wait on the same condition variable, this time on the condition `interruptWorking == false`.
3. Thread A, with its condition variable now signaled via `interruptWorking = true`, performs its work.
4. Once the created `Program::Guard` goes out of scope, it is destroyed, which will set `interruptWaiting` and `interruptWorking` to false, and signal the `programConditionVariable`.
5. The emulator run loop thread, with the condition variable signaled, proceeds with the next run loop.

This paradigm allows threads to both wait more efficiently as well as perform their necessary work in a much more timely fashion.

Currently, these synchronization primitives are only exposed through this RAII object, but in the future they could also be exposed via separate methods, blocks, etc., if needed.

### `hiro` Improvements

While analyzing performance, it was noticed that the GUI run loop was running unthrottled on Windows and Linux. This  would perform a wholly unnecessary amount of work, potentially refreshing thousands of times per second, and performing expensive amounts of drawing of the status bar. On Linux, this was unlikely to affect emulator performance, but on Windows, this amount of drawing could have a significant effect on performance due to the locking behavior of `SendMessage`.

This PR sets a uniform 8ms between GUI run loops on all platforms, which should dramatically affect Windows CPU idle and general performance without impacting responsiveness. Future improvements can be made here to improve the performance of ares "idle" state, without any title loaded, as well as potential better synchronization between the GUI and emulator run loops, especially around the status bar, tools window and input polling.

The emulator run loop remains un-throttled, responsible for suspending itself as needed, either within drivers or within the run loop for idle scenarios.

### DIFF
--- a/desktop-ui/program/program.hpp
+++ b/desktop-ui/program/program.hpp
@@ -3,6 +3,7 @@ struct Program : ares::Platform {
   auto main() -> void;
   auto emulatorRunLoop(uintptr_t) -> void;
   auto quit() -> void;
+  auto waitForInterrupts() -> void;
 
   //platform.cpp
   auto attach(ares::Node::Object) -> void override;
@@ -97,21 +98,40 @@ struct Program : ares::Platform {
     string text;
   } message;
 
+  class Guard;
+
   vector<Message> messages;
   string configuration;
   atomic<u64> vblanksPerSecond = 0;
-  /// The emulator run loop mutex. The emulator thread will hold a lock on this mutex for the duration of its run loop, including while the thread is suspended. The UI thread should only acquire this mutex when absolutely necessary, as there will be a severe UI responsiveness penalty acquiring it.
-  std::recursive_mutex programMutex;
-  
-  /// Mutex used to manage access to the status message queue.
-  std::recursive_mutex messageMutex;
-  
+
+  bool _isRunning = false;
+
   /// Mutex used to manage access to the input system. Polling occurs on the main thread while the results are read by the emulation thread.
   std::recursive_mutex inputMutex;
   
 private:
   atomic<bool> _quitting = false;
   atomic<bool> _needsResize = false;
+  
+  /// Mutex used to manage access to the status message queue.
+  std::recursive_mutex _messageMutex;
+
+  /// Mutex used to manage interrupts to the emulator run loop. Acquired when setting either of the `_interruptWaiting` or `_interruptWorking` booleans.
+  std::mutex _programMutex;
+  /// The emulator run loop condition variable. This variable is signaled to indicate other threads may access the program mutex, as well as to indicate that the emulator run loop can continue.
+  std::condition_variable _programConditionVariable;
+  /// Boolean indicating whether another thread is waiting to modify the emulator program state.
+  std::atomic<bool> _interruptWaiting = false;
+  /// Boolean indicating whether a non-emulator worker thread is in the process of modifying the emulator program state.
+  bool _interruptWorking = false;
+  /// Counter used to allow for possible recursive creation of `Program::Guard` instances (which can happen in some UI paths).
+  u32 _interruptDepth = 0;
+  /// Thread-local variable used to allow the emulator worker thread to create `Program::Guard` instances without causing a deadlock.
+  static inline thread_local bool _programThread = false;
+  /// Debug accessor for `_programThread` since debuggers cannot read TLS values correctly.
+  NALL_USED auto getProgramThreadValue() -> bool {
+    return _programThread;
+  }
 };
 
 extern Program program;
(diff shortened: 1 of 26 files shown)


hiro/gtk/application.cpp
```diff
@@ -27,10 +27,11 @@ auto pApplication::modal() -> bool {
 auto pApplication::run() -> void {
   while(!Application::state().quit) {
     Application::doMain();
+    usleep(8 * 1000);
     processEvents();
     //avoid spinlooping the thread when there is no main loop ...
     //when there is one, Application::onMain() is expected to sleep when possible instead
```
LukeUsher: nit: doesn't this comment need updating now that we always sleep after doMain()?

hiro/windows/application.cpp
```diff
@@ -22,6 +22,7 @@ auto pApplication::run() -> void {
     if(Application::state().onMain) {
       //doMain() is responsible for sleeping the thread where practical
```
LukeUsher: nit: same as above;

desktop-ui/program/program.hpp
```diff
@@ -97,21 +99,36 @@ struct Program : ares::Platform {
...
+  /// The emulator run loop condition variable. This variable is signaled to indicate other threads may access the program mutex, as well as to indicate that the emulator run loop can continue.
+  std::condition_variable _programConditionVariable;
+  /// Boolean indicating whether another thread is waiting to modify the emulator program state.
+  std::atomic<bool> _interruptWaiting = false;
+  /// Boolean indicating whether a non-emulator worker thread is in the process of modifying the emulator program state.
+  bool _interruptWorking = false;
+  /// Counter used to allow for possible recursive creation of `Program::Guard` instances (which can happen in some UI paths).
+  u32 _interruptDepth = 0;
+  /// Thread-local variable used to allow the emulator worker thread to create `Program::Guard` instances without causing a deadlock.
+  static inline thread_local bool _programThread;
```
LukeUsher: Would it be worth expanding nall:thread to have a function to get the thread id or handle, and instead storing the thread ID of the program thread rather than a thread local bool?

Maybe a later thing,

hiro/gtk/application.cpp
jcm93: Indeed, addressed.

desktop-ui/program/program.hpp
jcm93: I think this may be appropriate down the line if we made some other API that needs that functionality, but since "checking if a thread is some particular thread" is not *typically* what we want to find ourselves doing, I think it may be better to keep this private to the sync object for now.

I was also worried that our TLS value was not zero-initializing properly in other threads, but this was a debugger artifact. It should be sufficient to simply set `_programThread = true` when creating the emulator run loop, and the program guard will function as expected without other threads needing to worry about setting the value.

PR #2113, opened 2025-07-16
n64: Reinitialize paraLLEl-RDP on system reset

### LINKED ISSUE #2100: N64 Save States Broken
After creating an N64 save state (i.e Slot 1), when the slot is then loaded it reboots the game. Have tried other consoles and they continue to work with v145.

1. Make a save state using tools>save state>slot 1-9
2. Load that same save state slot
3. Proceed to watch the game reboot with no progress recorded

As shown on the other systems, the N64 moment saved to a slot ought be a loaded when requested

This feature continues to work for me using v144. All firmware is installed as always and settings are duplicated between v144 & v145

jcm93:
The threading rework moved Vulkan initialization to a deferred routine inside the emulator run loop, because paraLLEl-RDP expects to be called into from the same thread that initialized it.

Meanwhile, 625df94 addressed a deadlock in paraLLEl-RDP that could occur if a reset or power on was sent from a thread that was not the thread that initialized it. It addressed the problem by deferring the `power` call until the next emulator run loop.

Unfortunately, that broke savestates in the N64 core. During unserialization, a power command is sent to the system directly before restoring the components' state. This combined with the new behavior would cause the system to reset immediately after loading a savestate.

This PR in effect reverts the previous fix and introduces a hopefully more robust solution: Unloading Vulkan whenever `power` is called on the system, and queueing it to be completely re-initialized on the next run loop.

> [!NOTE]
> This has been tested on macOS, but should undergo more testing on other platforms to verify it works fine.

Fixes #2100

### DIFF
--- a/ares/n64/system/system.cpp
+++ b/ares/n64/system/system.cpp
@@ -81,9 +81,6 @@ auto System::run() -> void {
     vulkan.load(node);
     _vulkanNeedsLoad = false;
   }
-  if(_needsPower) {
-    _power(_reset);
-  }
   cpu.main();
 }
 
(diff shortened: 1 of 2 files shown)


bsdcode:
Verified on FreeBSD, it's working fine.

remutro:
Verified on Windows & Linux.

PR #2121, opened 2025-07-19
reduce cmake_minimum_required

omarandlorraine:
This pull request lets the project build on Debian Bookworm which ships with CMake 3.25.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.28...4.0)
+cmake_minimum_required(VERSION 3.25...4.0)
 
 include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/common/bootstrap.cmake" NO_POLICY_SCOPE)
 


jcm93:
The build system makes use of and/or relies on a number of features and behaviors introduced between CMake 3.25 and 3.28. The 3.28 target was set intentionally as the minimum version for that reason.

It may be possible to build ares using CMake 3.25, particularly on Linux where fewer of the new behaviors are as important, but I would not consider that supported. The [build documentation](https://github.com/ares-emulator/ares/wiki/Build-Instructions-For-Linux) recommends using the binary distribution or apt repository hosted by Kitware for environments where CMake 3.28 is not backported or otherwise available.

(You can also pass `-DCMAKE_POLICY_VERSION_MINIMUM=3.25` when invoking CMake to achieve the same behavior as this PR without modifying the `cmake_minimum_required()` call. If you run into issues doing this though, you would again be on your own.)

LukeUsher:
Agreed, the current requirement is not arbitrarily chosen but required on some of our target platforms, so we won't be merging this.

PR #2132, opened 2025-07-28
Implement aspect correction mode

rasky:
Currently, Ares has an "Aspect Correction" checkbox that allows to disable the PAR (pixel aspect ratio) configuration made by each core, to basically display the framebuffer without any additional scaling, with host PC aspect ratio.

In addition to this, it is common for people to request an "anamorphic 16:9" option, to correctly display games the way they would appear on a 16:9 TV when stretched. Sometimes, homebrew games are even designed to be viewed like this.

This PR changes the aspect mode from a boolean to a 3-value option. The standard option is now named "Standard (4:3 TV)", and a new option called "Anamorphic (16:9 TV)" is added.

Screenshot 2025-07-28 alle 11 09 31


rasky:
Change a bit the naming after private feedback by Luke to avoid TV references, a larger change might be required before we can go there

PR #2133, opened 2025-08-01
optimized 64-bit System V function calls

pekkavaa:
The commit messages say it all:

---

When ares switched to SLJIT, some tiny overhead was introduced in every call from guest to host code on 64-bit Linux. This was due to the argument register mapping differing between SLJIT's own convention and the System V AMD64 ABI.
This commit makes the recompiler put function arguments directly in the correct registers.

This commit introduces a new 'callf' template function that accepts recompiler's 'mem' and 'imm' argument types. It abstracts away the register setup so that different host platforms can implement their optimized function calls. This also results in slightly shorter code in the recompiler (see the followup commit).

The SLJIT_CALL_REG_ARG type of call is actually meant for SLJIT's own, optimized calling convention, but since it assumes function arguments are placed in correct registers, it works here. The only other option I could think would've been emitting x64 machine code directly.

In the n64-systemtest ROM, this change shaves off 6% and 10% of the respective VR4300 and RSP  median block sizes:
  
  ```
    VR4300 block size statistics
  
              min   max   mean     median
     before   113   6369  893.93   573
     after    104   5937  881.17   537
  
    RSP block size statistics
  
              min   max   mean     median
     before   97   19196  3941.95  5336
     after    88   17044  3414.06  4606
```  

  Also, on a benchmark with an RSP-intensive homebrew libdragon ROM, this
  change gave about 1% actual runtime speedup.

image

### DIFF
--- a/ares/n64/cpu/recompiler.cpp
+++ b/ares/n64/cpu/recompiler.cpp
@@ -53,29 +53,23 @@ auto CPU::Recompiler::emit(u64 vaddr, u32 address, bool singleInstruction) -> Bl
     mov64(PipelineReg(pc), reg(0));
     add64(PipelineReg(nextpc), reg(0), imm(4));
     if(callInstructionPrologue) {
-      mov64(reg(1), imm(vaddr));
-      mov32(reg(2), imm(instruction));
-      call(&CPU::instructionPrologue);
+      callf(&CPU::instructionPrologue, imm64(vaddr), imm(instruction));
     }
     if(numInsn == 0 || (vaddr&0x1f)==0){
       //abort compilation of block if the instruction cache is not coherent
       if(!self.icache.coherent(vaddr, address)) {
         resetCompiler();
-        return nullptr;  
+        return nullptr;
       }
-      mov64(reg(1), imm(vaddr));
-      mov32(reg(2), imm(address));
-      call(&CPU::jitFetch);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by pekkavaa** [comment]
This is ready for review. n64-systemtest results match the master branch:
```
n64-systemtest 2.1.0 (base=1 timing=0 cycle=0 cp0-hazards=0)
Finished in 3.98s. Base: Failed 1 of 3650 tests (99% success rate)

Slowest tests: COP1: CVT.L, ROUND.L, TRUNC.L, FLOOR.L, CEIL.L (0.17s), COP1: CVT.W, ROUND.W, TRUNC.W, FLOOR.W, CEIL.W (0.17s), RSP VRSQ (all 16 bit values) (0.13s), RSP VRCP (all 16 bit values) (0.12s), SUB.D (randomized - quick) (0.11s)
```

The failing case is `Test 'RDP STATUS: Flags during a run' failed: Time out waiting for RDP status 0xa8. RDP status at timeout: 0xe8` which is not a new issue.

pekkavaa:
I'm also open for alternative names for the `callf` function. I just took that one because it's short and could be said to represent "call fast".

pekkavaa:
I had to change the `setup_arg` template function to check unknown types a bit differently because `static_assert` is evaluated before instantiation https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2593r1.html
Now build works also on gcc.

PR #2138, opened 2025-08-06
n64: fix readback of PI_CART_ADDR after direct I/O

rasky:


thekovic:
A description would have been nice to document what the issue was and what was fixed.

rasky:
I've just discovered PI_CART_ADDR gets updated after a direct I/O read/write (that is a CPU access to the memory mapped area of the PI bus). n64brew too was documenting it only with respect to PI DMA transfers but actually it does get updated after a non-DMA transfer as well.

PR #2139, opened 2025-08-07
n64: improve PI DMA accuracy with stricter timings

rasky:
This commit splits the PI DMA-Write transfers (that is, loads from ROM)
in multiple blocks, following what happens on real hardware. Moreover,
it tries to more accurately reflect timing that each block takes. This
is easier to be done at the block granularity, otherwise the timing
algorithm would need to figure it out again the whole block splitting
logic.

From the point of view of the running application, now a ROM con see
the two hardware registers PI_CART_ADDR and PI_RDRAM_ADDR increment
one block at a time during the transfer, which more accurately reflect
what happens on real hardware.

This commit is validated with n64_pi_dma_test test ROM
(https://github.com/rasky/n64_pi_dma_test) which we now pass in full,
including timing tests with the default 10% tolerance.

Screenshot 2025-08-07 alle 14 58 42

### DIFF
--- a/nall/nall/priority-queue.hpp
+++ b/nall/nall/priority-queue.hpp
@@ -26,10 +26,12 @@ struct priority_queue<T[Size]> {
 
   template<typename F>
   auto step(u32 clocks, const F& callback) -> void {
-    clock += clocks;
-    while(size && ge(clock, heap[0].clock)) {
+    u32 baseClock = clock;
+    while(size && ge(baseClock+clocks, heap[0].clock)) {
+      clock = heap[0].clock;
       if(auto event = remove()) callback(*event);
     }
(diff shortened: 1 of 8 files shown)


nall/nall/priority-queue.hpp
```diff
@@ -26,10 +26,12 @@ struct priority_queue<T[Size]> {
 
   template<typename F>
   auto step(u32 clocks, const F& callback) -> void {
-    clock += clocks;
-    while(size && ge(clock, heap[0].clock)) {
+    u32 baseClock = clock;
```
invertego, line 29: Nitpick:
```suggestion
    u32 nextClock = clock + clocks;
```
would be a little more concise in usage below.

PR #2146, opened 2025-08-12
Adding Pioneer LaserActive MegaLD support to Ares

RogerSanders:
This pull request adds support for MegaLD titles to Ares. Due to the fact MegaLD titles are stored on Laserdiscs, this brings with it some unique requirements around data formats. Discussion has already happened in several places on Discord, but for visibility, here is some of that repeated to capture the main points:

I've introduced a new .mmi (Mixed Media Image) format. This is basically a renamed zip file with a json schema document in the root. This format can be used for more complex media, like the Laserdisc format, where we have multiple distinct streams of data, some digital, some with an analog base, and multiple sides/volumes to the one media. The idea is that it enough structure to support multiple "media" entries in the one set (IE, physical optical disc/floppy disk/hard disk/tape/etc), with multiple "streams" per media item, allowing, for example, for analog audio/analog video/digital audio, in the case of LD's. Within the json payload, there are relative paths within the archive to find the data for each stream.

Right now, I've made that work with minimal changes in Ares by making mmi files loadable as a "CD" format, with a couple of minor changes to mia/medium and nall/nall/vfs/cdrom.hpp. These changes are logically wrong though, as an MMI image effectively "is" a "Virtual File System" in its own right, bunding up a bunch of independent payloads, including multiple CD images, digitized analog audio, digitized analog video, etc, with a json metadata file at the root defining the grouping/structure of the whole thing. It seems to me, the vfs system in Ares is ideally and uniquely suited for being able to expose these mmi files easily and directly, but at this point after discussion on Discord, I'm submitting things as-is for now, and this may be addressed later after merge.

For the analog video itself within the MMI files, they're stored in a .qon file, decoding for which is done using the header-only qon library I created, which I've pulled into thirdparty. I've also pulled in nlohmann/json into third to parse the json payload in the MMI file, since it makes decoding a complex structure just a couple of lines of code, and the existing nall::json library is very basic and unused in the main codebase. Discussion can be had about whether that changes, but I thought it'd be good to get the code into peer review to talk about that further.

It's important to note I did significant enhancements to nall::zip, to add support for Zip64 archives (so I could go over 2GB - these laserdisc games average about 30GB for a single-side CAV title). I also added the ability to return an array_view to files in "store" compression mode. This was essential for the QON video files, since they're a 30GB file, we don't want to try and decompress them to memory, and they're already compressed internally, so adding deflate compression on top doesn't help. Since Ares memmaps zip files, this allows this very large analog video file, which needs to be able to be rapidly accessed, to be directly addressed as if it was a large array in memory, without actually fully loading it into memory. This is key to making everything work.

As for the actual emulation support, it's not feature complete, but it's pretty far along, and probably robust at this point for prime-time I think. I'll continue to fix issues and add support for the few remaining hardware features as I convert more titles and test them in Ares.


LukeUsher:
There are multiple places where nall containers could be used over std; usually we'd prefer to use the containers in nall unless there is some performance benefit for otherwise.

eg; you use std::vector, but we have nall::vector
Smart pointers and std::optional (as nall::maybe) have nall equivalents too.

But most (all?) of this can be improved over time, the same thing with properly handling containers in mia and not in the emulator core.

Please don't forget to update the LICENSE file in the project root to include the license text for any third-party libraries that this work includes

Overall: great work!

RogerSanders:
Looks like I've got problems with including the implementation from the header only library. I was confused by the unity build structure, couldn't figure out the right place. I'm off to bed, but if someone has a suggestion of a cpp module I can add `#define QOI2_IMPLEMENTATION` and `#define QON_IMPLEMENTATION` in, which isn't inside a namespace, that can be used before `#include <qon/qon.h>` and `#include <qon/qoi2.h>`, I'd appreciate it. I can just create a standalone module which does only that if it's the best way, just let me know where and how would be the best way to do it.

RogerSanders:
> There are multiple places where nall containers could be used over std; usually we'd prefer to use the containers in nall unless there is some performance benefit for otherwise.
> 
> eg; you use std::vector, but we have nall::vector Smart pointers and std::optional (as nall::maybe) have nall equivalents too.
> 
> But most (all?) of this can be improved over time, the same thing with properly handling containers in mia and not in the emulator core.

That's a good point. I meant to do a pass back over that, but forgot. I did try and use nall containers initially, but at some point I got into a "I just want to make this work" mode and didn't want to search for nall equivalents or how to use them. I'll take a look tomorrow, see what I can move over.

> Please don't forget to update the LICENSE file in the project root to include the license text for any third-party libraries that this work includes

Thanks, missed that. Everything I've pulled in is MIT licensed, but I can see you like to include every library license in full. I'll add it in.

> Overall: great work!

Thanks. Took a bit to get here, but I'm glad to see it running games.

LukeUsher:
I attempted to rebase and squash into a branch at https://github.com/ares-emulator/ares/tree/megald; 

You can check out this branch and do any future work as pull-requests against that branch at this time, so we have a clean base, then when we're happy, we can squash-merge that branch into master and have a shippable build.

LukeUsher:
development migrated into https://github.com/ares-emulator/ares/tree/megald; so we can all collaborate with some cleanup and getting ready for upstream.

Please submit future work as pull requests to that branch, and once everyone is happy, we will merge that branch into master.

PR #2149, opened 2025-08-14
nall: remove memory ownership from string_view

invertego:
Previously, string_view had a constuctor that allocated a new string and forwarded all arguments to the string constructor. Among other things, this allowed construction of string_view objects using any of the various stringify overloads. However, there were drawbacks:

1.  The constructor string_view(const char*, u32) shadowed the forwarding constructor, leading to very different behavior. In the forwarding version, the first parameter would be treated as a null-terminated string, and the second parameter would be formatted as a string and concatenated with the first.
2.  Having both owning and non-owning modes makes it difficult for someone reading the code to reason about lifetimes, as it depends on how the object was constructed. It also deviates from the behavior of the identically-named standard C++ string_view.


invertego:
Removed the mmi.hpp part of the change so there would be no conflicts with #2150. The cleanup can be done after one or both PRs are completed.

PR #2151, opened 2025-08-14
nall: clean up string conversions in cue/mmi decode

invertego:
Now that the usage issues with string_view have been resolved it can be safely used to wrap a vector<u8>. I'll note that passing a vector<u8> directly to the string constructor would work as well (thanks to a stringify overload) but I'm preserving the wishes of the author of the MegaLD code to keep the type conversions explicit.

I'll also note that we're still constructing a string from the string_view when we call JSON::unserialize(). Ideally, it would take a string_view instead of a const string& and we could avoid the copy. Unfortunately, all of the markup decoders currently require the input to be null-terminated, which is only guaranteed by the string class.


invertego:
This is just tidying up after #2150 to utilize the improvements made in #2149.

PR #2158, opened 2025-08-19
ld: Fixed a double buffering issue which caused single-frame errors

RogerSanders:
Saw a few single-frame transient glitches with Myst which were caused by an error in the recent video double buffering I added. This changes savestates, so I wanted to get it addressed before the new release drops.

### DIFF
--- a/ares/md/mcd/mcd.hpp
+++ b/ares/md/mcd/mcd.hpp
@@ -437,8 +437,8 @@ struct MCD : M68000, Thread {
       s32 currentVideoFrameIndex;
       n1 currentVideoFrameLeadIn;
       n1 currentVideoFrameLeadOut;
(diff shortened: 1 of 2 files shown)


remutro:
You mentioned this changes save state, but I don't see an increase in save state version number here. Usually even nightly builds see a version increase when save state format changes, otherwise we attempt to load incompatible save states. Each core has it's own versioning, so it doesn't need to be tied to an official release. As long as the string is different (and increasing for human benefit), even if it becomes 144.1, 144.2, etc. with each change is fine.  String is the first line in ares/md/system/serialization.cpp (I don't think we distinguish between MD/CD/32X and now LD).

LukeUsher:
I'll bump save state version after merge

PR #2161, opened 2025-08-21
ld: Fix overscan mode, and address some redraw and playback issues.

RogerSanders:


ares/ares/node/video/screen.cpp
```diff
@@ -349,12 +350,17 @@ auto Screen::refresh() -> void {
 }
 
 auto Screen::lookupPalette(u32 index) -> u32 {
-  return _palette[index];
+  return (_palette ? _palette[index] : 0);
```
jcm93: Per comments off-thread, this should not be necessary after 778a32700d8c93fb4d689475283b20a9ea341a85; I would remove this so we can more easily detect if there is some other issue here.

RogerSanders: Change reverted

PR #2166, opened 2025-08-22
Implement Bio Sensor accessory for Nintendo 64

meeq:
**Stacked on top of https://github.com/ares-emulator/ares/pull/2173**

Bio Sensor is an obscure peripheral that reads heart beat pulses using infrared so that games can react to a player's heart-rate. This emulated version of Bio Sensor allows you to set a desired heart-rate, defaulting to 60 BPM.

This has been validated using the [LibDragon Bio Sensor test ROM](https://github.com/DragonMinded/libdragon/blob/preview/examples/biosensortest/biosensortest.c):  [biosensortest.z64.zip](https://github.com/user-attachments/files/21976041/biosensortest.z64.zip)

Screenshot 2025-08-25 at 15 45 51

I was also able to validate this implementation against Tetris 64 (the only official game that supports Bio Sensor) by hacking the values in `PIF::estimateTiming()`. 

Validated? | Description | Screenshot
--- | --- | ---
✅  | Single-Player "Bio Tetris" shows pulse with one controller and one Bio Sensor connected: | Screenshot 2025-08-25 at 15 14 30
✅  | Two-player "Bio Tetris" works with two controllers and two Bio Sensors: | Screenshot 2025-08-25 at 15 10 27
✅  | Sensor Check screen works with a single controller and Bio Sensor; readings are accurate: | Screenshot 2025-08-25 at 15 09 40
❌  | Sensor Check screen does not work with more than one Bio Sensor; only the sensor in port 1 is detected and readings are indeterminate | Screenshot 2025-08-25 at 15 17 19

I believe that the outstanding issue with "Sensor Check" for multiple Bio Sensors is related to PIF timings.

PR #2173, opened 2025-08-25
Improve N64 controller accessory handling

meeq:
**Extracted from https://github.com/ares-emulator/ares/pull/2166**

1. Properly reports the 0x03 status when a new N64 controller accessory has been inserted, which tells the software to re-detect the accessory type.
2. Properly prevents N64 accessory read/write commands to a newly-inserted Pak until identify is called.
3. Improves the N64 accessory read/write commands to accurately handle non-standard data lengths.

## Detecting an accessory change

N64 software's ability to determine when an accessory has changed relies on either:
* Controller disconnect is reported, which requires at least one "status" or "controller read" command between unplugging the current controller and plugging in the new one.
* 0x02 (Pak Absent) status is reported, which requires at least one "status" command between removing the current accessory and connecting the new one.
* 0x03 (Pak Inserted) status is reported, which allows detection of instantaneous accessory swaps.

Validated with the LibDragon joypadtest example ROM: [joypadtest.z64.zip](https://github.com/user-attachments/files/21976967/joypadtest.z64.zip)
To exercise, switch between "Controller Pak" and "Rumble Pak" in the "Controller Port 1" > "Pak" sub-menu. The detected accessory should change.

## Non-standard accessory read/write length

[The console behavior findings are posted in the N64brew Discord](https://discord.com/channels/205520502922543113/1336835549901885482/1411034121731309729)

## Menu improvements

This PR includes some additional cleanup of the `Nintendo64::portMenu` function:
* Hide "Pak" submenu when Mouse is connected
* Hide "Transfer Pak" option when Game Boy core is disabled
* Refactor "set currently enabled pak" logic
* Switch Paks without disconnecting + reconnecting Gamepad

### DIFF
--- a/ares/n64/controller/gamepad/gamepad.cpp
+++ b/ares/n64/controller/gamepad/gamepad.cpp
@@ -184,56 +184,68 @@ auto Gamepad::comm(n8 send, n8 recv, n8 input[], n8 output[]) -> n2 {
   //read pak
   if(input[0] == 0x02 && send >= 3 && recv >= 1) {
     u16 address = (input[1] << 8 | input[2] << 0) & ~31;
-    //check address CRC and DETECT latch state
-    if(pif.addressCRC(address) != (n5)input[2] || pakDetectLatched) {
-      //zero out pak read data if address CRC is invalid or pak detect is latched
-      for(u32 index : range(recv - 1)) output[index] = 0x00;
-      output[recv - 1] = 0xFF; //invalid pak state
+
+    //NUS-CNT only touches the first 33 bytes of output (32 data + 1 data CRC);
+    //reads of less than 32 bytes are technically possible, but with no data CRC
+    n8 recv_data_len = min(recv, 32);
+    for(u32 index : range(min(recv, 33))) output[index] = 0x00; //zero out up to 33 bytes
+
+    //check if no pak is connected, DETECT is latched, or address CRC is invalid
+    b1 data_crc_no_pak = 0;
+    if(!slot || pakDetectLatched || pif.addressCRC(address) != (n5)input[2]) {
+      data_crc_no_pak = 1;
       valid = 1;
-      goto comm_done;
+      goto read_pak_data_crc;
     }
 
     //controller pak
     if(ram) {
-      for(u32 index : range(recv - 1)) {
+      for(u32 index : range(recv_data_len)) {
         // read into current bank
         if(address <= 0x7FFF) output[index] = ram.read<Byte>(bank * 32_KiB + address);
         else output[index] = 0;
         address++;
       }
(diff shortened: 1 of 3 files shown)


rasky:
> 0x03 (Pak Changed) status is reported, which allows detection of instantaneous accessory swaps.

I was wondering what this value meant on the hardware side, as obviously there's nothing like an "instantaneous accessory swap", physically speaking.

N64brew says that the accessory status byte is a bitfield:

> Bitfield:
> 0x04: checksum error in previous command
> 0x02: No Pak installed
> 0x01: Pak installed

So while not very explicit, my guess is that this is really a bitfield with the following meaning:

* 0x01: is a pak currently installed (level triggered)
* 0x02: was a pak disconnected since last time status was requested (sticky bit, edge triggered)
* 0x04: did a checksum error happen since last time status was requested (sticky bit, edge triggered)

So I don't think 0x3 "pak changed" actually exists. You can say that 0x03 can be interpreted as "pak changed" because it is a combination of "a pak is installed now" and "a pak was disconnected", but probably the Ares change should reflect what the hardware does instead, and keep a sticky bit for "pak disconnected", which is returned as independent bit 0x02 one time, and then gets cleared.

meeq:
There are a few third-party accessories such as “Tremor Pak Plus” that do allow the accessory to be changed instantly:
image

When the Save/Rumble switch is toggled, Tremor Pak Plus swaps between its built-in Rumble Pak and the pass-through accessory port.

I will investigate your hunch further and try to confirm the actual meaning behind the controller status bits.

meeq:
Still to be implemented: skipping accessory read and write commands when the "DETECT" latch is tripped until the next reset/identify command.

This was verified on console:
image

ares/n64/controller/gamepad/gamepad.cpp
```diff
@@ -57,6 +57,7 @@ auto Gamepad::allocate(string name) -> Node::Peripheral {
 
 auto Gamepad::connect() -> void {
   if(!slot) return;
+  pakDetectLatched = true;
```
rasky, line 60: Is this really needed? I thought the latch was only set on disconnection.

meeq: I just verified this on real hardware.

If I cold-boot the console with no accessory connected, Identify shows 0x02 status:
Screenshot 2025-08-28 15-41-29

If I connect an accessory and re-identify, status is 0x03:
Screenshot 2025-08-28 15-41-41

The DETECT latch is triggered by both connecting and disconnecting an accessory.

LukeUsher:
merged as https://github.com/ares-emulator/ares/pull/2166

PR #2177, opened 2025-08-27
nall: remove memory ownership from string_view again

RogerSanders:
Removed owning string_view again, this time explicitly removing string_view construction from temporary string objects, which resulted in dangling pointers into temporary string buffers. Temporary strings must now be explicitly saved into named objects with clear lifetimes in order to implicitly convert to a string_view. While implicit conversion to a string_view may be safe in some cases due to the guaranteed lifetime of temporaries in C++ being the full expression, it's easy to still create problems when the conversion to a string_view is implicit, when the caller might believe they have passed in a string which has been copied, when in fact a reference to the provided string objects internal buffers has been captured. It's better to disallow this case, and let the programmer explicitly create an object with a clear lifetime, than risk dangling pointer issues.

### DIFF
--- a/nall/nall/string.hpp
+++ b/nall/nall/string.hpp
@@ -198,6 +198,7 @@ public:
   template<typename T, typename... P> auto prepend(const T&, P&&...) -> type&;
   template<typename... P> auto prepend(const nall::string_format&, P&&...) -> type&;
   template<typename T> auto _prepend(const stringify<T>&) -> type&;
+  template<typename... P> auto append(string&& value, P&&... p) -> string&;
   template<typename T, typename... P> auto append(const T&, P&&...) -> type&;
   template<typename... P> auto append(const nall::string_format&, P&&...) -> type&;
   template<typename T> auto _append(const stringify<T>&) -> type&;
--- a/nall/nall/string.hpp
+++ b/nall/nall/string.hpp
@@ -39,8 +39,7 @@ struct string_view {
   string_view(const char* data);
   string_view(const char* data, u32 size);
   string_view(const string& source);
-  template<typename... P> string_view(P&&... p);
-  ~string_view();
+  string_view(string&& source) = delete;
 
   auto operator=(const string_view& source) -> type&;
   auto operator=(string_view&& source) -> type&;
@@ -57,7 +56,6 @@ struct string_view {
   auto end() const { return &_data[size()]; }
 
 protected:
-  string* _string;
   const char* _data;
   mutable s32 _size;
 };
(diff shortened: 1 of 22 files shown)


RogerSanders:
I'm not sure why the server builds are picking up a bunch of cases here I don't see from a clean build locally through VS. Going to hit pause while I dig through the build scripts for a bit to better understand what's happening differently for me locally vs the CI/CD pipeline.

jcm93:
I'm still seeing some odd issues here, would like a bit more time to look into this more closely.

jcm93:
OK after looking at this more I think this is fine and has indeed caught every instance of problematic `string_view` creation, at least from what I can see.

This does play havoc with Ccache on my systems, which will be worth noting in case people (or CI) have odd build failures after this is merged.

nall/nall/string.hpp
```diff
@@ -39,8 +39,7 @@ struct string_view {
   string_view(const char* data);
   string_view(const char* data, u32 size);
   string_view(const string& source);
-  template<typename... P> string_view(P&&... p);
-  ~string_view();
+  string_view(string&& source) = delete;
```
invertego, line 42: I can construct a std::string_view from a std::string&&. Can you make a case for why nall should be stricter than the C++ standard library?

I'm not convinced that forcing callers to convert a temporary into a local is a comprehensive solution, because if the callee holds onto a reference to the memory pointed to by the string_view it could still very well outlast the local variable in the caller's frame. I feel the onus is on the callee to read the contents of a string_view before returning.

RogerSanders: Sorry, missed this comment. So I added this in as a temporary guardrail. Basically the problem was this - there were places in code which were assigning temporary nall::string objects to non-temporary nall::string_view objects. This worked previously however, since passing in a temporary string into a string_view was using the variadic constructor and creating an owning string_view. I blocked this being possible, by explicitly deleting the constructor that takes in a temporary, which allowed me to flush out all usage of this.

I could have just blocked construction from a temporary for the purposes of making the change, and removed this line for the PR, but I chose to leave it in place at least temporarily, because I was worried about the possibility of a concurrent change being made, which relied on the old behaviour. This wouldn't easily be caught during the review, and would introduce a subtle use-after-free runtime error.

Since there are only a handful of places in the code that relied on being able to construct a string_view from a temporary, and in all but a couple of them using a named variable improved readability IMO, and/or performance, combined with the concerns of breaking other concurrent changes, I felt it was best to leave the restriction in place. I think with the concurrent work now happening to remove nall::string and nall::string_view entirely, this can probably remain in place until that job comes through.

PR #2178, opened 2025-08-27
Start migration of nall::vector to std::vector

rasky:
This PR is the first attempt to start migrating to proper standard C++ containers getting rid of nall specific ones.

It is a huge change because it is really difficult to split. When this change is merged, about 50% of nall::vector uses across the whole codebase will be migrated to std::vector. CI should be green.

The PR is not meant to contain any functional change. It is only mechanical change. Besides literal translations, I did *NOT* attempt to improve the way code is written after migration to std::vector. There are a few instances where the previous code was smaller and more readable, but I forced myself to only do mechanical changes otherwise the scope would get impossible to handle.

~~The only functional change to my knowledge is in commit https://github.com/ares-emulator/ares/commit/ad2482fbc39427bafdeb03a5ffa27bf190724841 where it seems like the change uncovered a latent bug that caused a segfault during cleanup on macOS. I did not investigate the bug properly to avoid being derailed. The one-liner change here is sufficient to avoid the segfault.~~

**Testing will be required after merge**. I can only test Ares lightly on a single operating system. We are at the start of a release cycle anyway so it's a good moment for such a large change to land.


ruby/input/joypad/sdl.cpp
```diff
@@ -120,11 +120,12 @@ struct InputJoypadSDL {
       jp.hid->setProductID(pid);
       jp.hid->setPathID(jp.id);
       for(u32 n : range(axes)) jp.hid->axes().append(n);
+      jp.axisPolled.assign(axes, false);
```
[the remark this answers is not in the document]
rasky: This seems a spurious change, probably a pasto

ruby/input/joypad/sdl.cpp
```diff
@@ -6,17 +6,17 @@ struct InputJoypadSDL {
...
     SDL_Joystick* handle = nullptr;
   };
-  vector<Joypad> joypads;
+  std::vector<Joypad> joypads;
 
   auto assign(Joypad& joypad, u32 groupID, u32 inputID, s16 value) -> void {
     auto& group = joypad.hid->group(groupID);
     if(group.input(inputID).value() == value) return;
-    if(groupID == HID::Joypad::GroupID::Axis && !joypad.axisPolled(inputID)) {
+    if(groupID == HID::Joypad::GroupID::Axis && (inputID < joypad.axisPolled.size()) && !joypad.axisPolled[inputID]) {
```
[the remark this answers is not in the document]
rasky: The bound check helps matching the behavior of `nall::vector::operator() const` that would return a default value for an out-of-bound index access.

ruby/input/joypad/sdl.cpp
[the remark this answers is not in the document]
rasky: Removed

nall/nall/smtp.cpp
```diff
@@ -114,15 +114,15 @@ NALL_HEADER_INLINE auto SMTP::send(s32 sock, const string& text) -> bool {
...
     char c;
     if(::recv(sock, &c, sizeof(char), 0) < 1) break;
-    buffer.append(c);
+    buffer.push_back((u8)c);
     if(c == '\n') break;
   }
-  buffer.append(0);
-  return buffer;
+  buffer.push_back(0);
+  return string((const char*)buffer.data(), (u32)buffer.size());
```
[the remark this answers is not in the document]
rasky: Not a real issue

nall/nall/vfs/cdrom.hpp
```diff
@@ -188,7 +189,7 @@ struct cdrom : file {
             rawDataView = archive->dataViewIfUncompressed(*fileEntry);
           } else {
             rawDataBuffer = archive->extract(*fileEntry);
-            rawDataView = rawDataBuffer;
+            rawDataView = array_view<u8>(rawDataBuffer.data(), rawDataBuffer.size());
```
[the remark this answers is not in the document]
rasky: Removed all redundant array_view constructions that can be implicilyl converted.

ruby/video/video.hpp
```diff
@@ -61,8 +61,8 @@ struct VideoDriver {
 };
 
 struct Video {
-  static auto hasDrivers() -> vector<string>;
-  static auto hasDriver(string driver) -> bool { return (bool)hasDrivers().find(driver); }
+  static auto hasDrivers() -> std::vector<string>;
+  static auto hasDriver(string driver) -> bool { auto v = hasDrivers(); return std::ranges::find(v, driver) != v.end(); }
```
jcm93: Would prefer if this was something more descriptive like `drivers` instead of `v`. Also I think it's fine to split this into multiple lines since it is no longer a one-liner.

ruby/video/video.hpp
```diff
@@ -17,10 +17,10 @@ struct VideoDriver {
   virtual auto hasThreadedRenderer() -> bool { return false; }
   virtual auto hasNativeFullScreen() -> bool { return false; }
   virtual auto hasFlush() -> bool { return false; }
-  virtual auto hasFormats() -> vector<string> { return {"ARGB24"}; }
+  virtual auto hasFormats() -> std::vector<string> { return {"ARGB24"}; }
   virtual auto hasShader() -> bool { return false; }
 
-  auto hasFormat(string format) -> bool { return (bool)hasFormats().find(format); }
+  auto hasFormat(string format) -> bool { auto v = hasFormats(); return std::ranges::find(v, format) != v.end(); }
```
jcm93: Similar comment as with `hasDrivers`.

ruby/video/video.cpp
```diff
@@ -423,7 +423,7 @@ auto Video::monitor(string name) -> Monitor {
     if(monitor.primary) return monitor;
   }
   //if only one monitor is found, but it is not primary, use that
-  if(monitors.size() == 1) return monitors.left();
+  if(monitors.size() == 1) return monitors[0];
```
jcm93, line 426: Elsewhere `front()` is used for this; should be consistent.

ruby/input/input.hpp
```diff
@@ -26,8 +26,8 @@ struct InputDriver {
 };
 
 struct Input {
-  static auto hasDrivers() -> vector<string>;
-  static auto hasDriver(string driver) -> bool { return (bool)hasDrivers().find(driver); }
+  static auto hasDrivers() -> std::vector<string>;
+  static auto hasDriver(string driver) -> bool { auto v = hasDrivers(); return std::ranges::find(v, driver) != v.end(); }
```
jcm93: Similar comments as in `video::hasDrivers()`; too terse variable name, multiple lines since it's no longer a one-liner.

ruby/audio/xaudio2.cpp
```diff
@@ -8,7 +8,10 @@ struct AudioXAudio2 : AudioDriver, public IXAudio2VoiceCallback {
   ~AudioXAudio2() { destruct(); }
 
   auto create() -> bool override {
-    if(hasDevices()) super.setDevice(hasDevices().first());
+    {
```
jcm93: Is there any reason for this to be a scoped block?

ruby/audio/waveout.cpp
```diff
@@ -74,8 +74,11 @@ struct AudioWaveOut : AudioDriver {
   auto initialize() -> bool {
     terminate();
 
-    auto deviceIndex = hasDevices().find(self.device);
-    if(!deviceIndex) deviceIndex = 0;
+    u32 deviceIndex = 0;
+    {
+      auto names = hasDevices();
```
jcm93: Same as in `ruby/audio/xaudio2.cpp`; why the scoped block?. Should also keep the variable name consistent probably; `devices` rather than `names`. Maybe better for both is `deviceNames`. `idx` also seems unnecessarily terse but basically a nit.

ruby/audio/wasapi.cpp
```diff
@@ -66,7 +66,10 @@ struct AudioWASAPI : AudioDriver {
 
   auto create() -> bool override {
     super.setExclusive(false);
-    if(hasDevices()) super.setDevice(hasDevices().first());
+    {
+      auto devices = hasDevices();
```
jcm93: Same comment as in other drivers about the scoped block.

ruby/audio/oss.cpp
```diff
@@ -96,7 +96,12 @@ struct AudioOSS : AudioDriver {
   auto initialize() -> bool {
     terminate();
 
-    if(!hasDevices().find(self.device)) self.device = hasDevices().first();
+    {
+      auto devices = hasDevices();
+      if(std::ranges::find(devices, self.device) == devices.end()) {
```
jcm93: Same comment as in other drivers about the scoped block.

ruby/audio/openal.cpp
```diff
@@ -92,7 +92,7 @@ struct AudioOpenAL : AudioDriver {
     timeBeginPeriod(1);
 #endif
 
-    if(!hasDevices().find(self.device)) self.device = hasDevices().first();
+    if (auto devs = hasDevices(); !devs.empty() && !index_of(devs, self.device)) self.device = devs.front();
```
jcm93, line 96: `devs` is unnecessarily terse; also this seems like it would likely be cleaner if split into multiple lines.

ruby/audio/audio.hpp
```diff
@@ -52,8 +52,8 @@ struct AudioDriver {
 };
 
 struct Audio {
-  static auto hasDrivers() -> vector<string>;
-  static auto hasDriver(string driver) -> bool { return (bool)hasDrivers().find(driver); }
+  static auto hasDrivers() -> std::vector<string>;
+  static auto hasDriver(string driver) -> bool { auto v = hasDrivers(); return std::ranges::find(v, driver) != v.end(); }
```
jcm93: Same comments as in video and input; better variable name than `v` and consider multiple lines.

ruby/audio/audio.hpp
```diff
@@ -12,17 +12,17 @@ struct AudioDriver {
...
+  virtual auto hasChannels() -> std::vector<u32> { return {2}; }
+  virtual auto hasFrequencies() -> std::vector<u32> { return {48000}; }
+  virtual auto hasLatencies() -> std::vector<u32> { return {0}; }
 
-  auto hasDevice(string device) -> bool { return (bool)hasDevices().find(device); }
-  auto hasChannels(u32 channels) -> bool { return (bool)hasChannels().find(channels); }
-  auto hasFrequency(u32 frequency) -> bool { return (bool)hasFrequencies().find(frequency); }
-  auto hasLatency(u32 latency) -> bool { return (bool)hasLatencies().find(latency); }
+  auto hasDevice(string device) -> bool { auto v = hasDevices(); return std::ranges::find(v, device) != v.end(); }
+  auto hasChannels(u32 channels) -> bool { auto v = hasChannels(); return std::ranges::find(v, channels) != v.end(); }
```
jcm93: Same comment as line 56. Since these are in a block of four maybe it's fine to leave as "one line;" but still something better than `v` for the variable.

nall/nall/shared-pointer.hpp
```diff
@@ -3,7 +3,6 @@
 #include <nall/function.hpp>
 #include <nall/maybe.hpp>
 #include <nall/traits.hpp>
-#include <nall/vector.hpp>
```
jcm93, line 6: Was this unused in the first place? If not, is it better to include `<vector>` here rather than whatever uses vector underneath this?

nall/nall/directory.cpp
```diff
@@ -23,10 +23,14 @@ NALL_HEADER_INLINE auto directory::ufolders(const string& pathname, const string
       if(!*p) *p = ';';
       p++;
     }
-    return string{(const char*)utf8_t(drives)}.replace("\\", "/").split(";");
+    auto temp = string{(const char*)utf8_t(drives)}.replace("\\", "/").split(";");
```
jcm93: It took me longer than it ought to have to figure out what was going on here, partially because of the variable name. Would suggest renaming `temp` to something more descriptive, perhaps `parts`.

mia/system/master-system.cpp
```diff
@@ -9,8 +9,8 @@ auto MasterSystem::load(string location) -> LoadResult {
 
   this->location = locate();
   pak = new vfs::directory;
-  if(bios) pak->append("bios.rom", bios);
-
+  if(!bios.empty()) pak->append("bios.rom", bios);
+
```
jcm93: Nit; whitespace.

mia/medium/neo-geo.cpp
```diff
@@ -265,13 +265,13 @@ auto NeoGeo::decryptCmcGraphicsInternal(u8 *r0, u8 *r1, u8 c0,  u8 c1, u8 *table
...
-auto NeoGeo::decryptCmcMusic(vector<u8>& m) -> void {
-  vector<u8> input = m;
+auto NeoGeo::decryptCmcMusic(std::vector<u8>& m) -> void {
+  std::vector<u8> input = m;
 
   //checksum of the first 64k of ROM is used as a key
   u16 key = 0;
   for(auto i = 0; i < 0x10000; i++) key += input[i];
-  for(auto i = 0; i < input.size(); i++) m[i] = input[decryptCmcMusicDescramble(i, key)];
+  for(auto i = 0; i < (int)input.size(); i++) m[i] = input[decryptCmcMusicDescramble(i, key)];
```
jcm93, line 274: Perhaps it's better to just declare `i` as an `int` here rather than perform this cast?

mia/medium/mame.cpp
```diff
@@ -61,13 +61,14 @@ auto Mame::loadRoms(string location, Markup::Node& info, string sectionName) ->
...
 
-auto Mame::loadRomFile(string location, string filename, Markup::Node& info) -> vector<u8> {
+auto Mame::loadRomFile(string location, string filename, Markup::Node& info) -> std::vector<u8> {
   Decode::ZIP archive;
   if(!archive.open(location)) return {};
 
   for(auto& file : archive.file) {
     if(file.name.iequals(filename)) {
-      return archive.extract(file);
+      auto tmp = archive.extract(file); // std::vector<u8>
```
jcm93: Any reason this is like this rather than just returning directly?

hiro/windows/application.hpp
```diff
@@ -14,7 +14,7 @@ struct pApplication {
   static auto initialize() -> void;
 
   struct State {
-    vector<pMenuBar*> menuBarsToRebuild;  //deferred menu bars to update
+    std::vector<pMenuBar*> menuBarsToRebuild;  //deferred menu bars to update
```
jcm93, line 17: Nit; comment alignment.

hiro/extension/table-layout.cpp
```diff
@@ -13,16 +13,17 @@ auto mTableLayout::append(sSizable sizable, Size size) -> type& {
...
 
 auto mTableLayout::cell(u32 position) const -> TableLayoutCell {
-  return state.cells(position, {});
+  if(position < state.cells.size()) return state.cells[position];
+  return {};
 }
 
 auto mTableLayout::cell(u32 x, u32 y) const -> TableLayoutCell {
-  if(auto cell = state.cells(y * columnCount() + x, {})) return cell;
+  if(auto index = y * columnCount() + x; index < state.cells.size()) return state.cells[index];
```
jcm93: Would prefer to not have semicolons in `if` statements generally. I think moving the declaration outside of the if statement is better for readability here.

hiro/extension/horizontal-layout.cpp
```diff
@@ -207,7 +208,7 @@ auto mHorizontalLayout::setPadding(Geometry padding) -> type& {
 }
 
 auto mHorizontalLayout::setParent(mObject* parent, s32 offset) -> type& {
-  for(auto& cell : reverse(state.cells)) cell->destruct();
+  for(auto& it : state.cells | std::views::reverse) it->destruct();
```
jcm93: Should still be named `cell` instead of `it`.

hiro/extension/browser-dialog.cpp
```diff
@@ -252,10 +255,13 @@ auto BrowserDialogWindow::run() -> BrowserDialog::Response {
   if(state.action.beginsWith("select")) acceptButton.setText(tr("Select"));
   cancelButton.setText(tr("Cancel")).onActivate([&] { window.setModal(false); });
 
-  if(!state.filters) state.filters.append("All|*");
+  if(state.filters.empty()) state.filters.push_back("All|*");
   for(auto& filter : state.filters) {
     auto part = filter.split("|", 1L);
-    filters.append(part.right().split(":"));
+    auto right = part.right().split(":");
```
jcm93: Not very happy with the readability here. It wasn't beautiful to begin with but I think it still needs to be improved. I'm not 100% sure what filters are here, but perhaps we could do `filterComponent` and `filterComponentParts` or something? Anything that shows that the thing we're currently calling `right` is in fact multiple things would be the priority for clarity I think.

hiro/cocoa/widget/table-view.cpp
```diff
@@ -62,6 +62,7 @@
   }
 
   for(auto& tableViewColumn : tableView->state.columns) {
+    if(!tableViewColumn) continue;
```
jcm93: Why is this here?

desktop-ui/presentation/presentation.cpp
```diff
@@ -235,11 +235,11 @@ Presentation::Presentation() {
     .show();
   });
 
-  viewport.setDroppable().onDrop([&](auto filenames) {
+  viewport.setDroppable().onDrop([&](std::vector<string> filenames) {
     Program::Guard guard;
     if(filenames.size() != 1) return;
-    if(auto emulator = program.identify(filenames.first())) {
-      program.load(emulator, filenames.first());
+    if(auto emulator = program.identify(filenames[0])) {
```
jcm93: Elsewhere we use `front()` for this; better to be consistent.

ares/md/mcd/megald.cpp
```diff
@@ -11,14 +11,15 @@ auto MCD::LD::load(string location) -> void {
 
   //attempt to locate the requested media in the mmi archive
   string mediaName = mcd.disc->attribute("media");
-  auto mediaIndex = mmi.media().find([mediaName](auto& m) { return m.name == mediaName; });
-  if(!mediaIndex) mediaIndex = 0;
+  auto& mediaVec = mmi.media();
+  auto it = std::ranges::find_if(mediaVec, [mediaName](const auto& m){ return m.name == mediaName; });
+  u32 mediaIndex = it == mediaVec.end() ? 0u : static_cast<u32>(it - mediaVec.begin());
```
jcm93, line 17: It seems like this should probably just be cleaned up to get the reference to the media instead of doing this hooplah with the index, if possible.

ares/ares/node/class.hpp
```diff
@@ -7,23 +7,25 @@ struct Class {
...
     } else {
       throw;
     }
   }
 
   static auto create(string identifier) -> Node::Object {
-    if(auto index = classes().find([&](auto instance) { return instance.identifier == identifier; })) {
-      return classes()[*index].create();
-    }
+    auto& cls = classes();
```
jcm93, line 26: Would prefer to use a non-abbreviated variable name.

ares/ares/node/class.hpp
```diff
@@ -7,23 +7,25 @@ struct Class {
...
-    static vector<Instance> classes;
+  static auto classes() -> std::vector<Instance>& {
+    static std::vector<Instance> classes;
     return classes;
   }
 
   template<typename T> static auto register() -> void {
-    if(!classes().find([&](auto instance) { return instance.identifier == T::identifier(); })) {
-      classes().append({T::identifier(), &T::create});
+    auto& cls = classes();
```
jcm93, line 16: Ibid, abbreviated variable name.

desktop-ui/emulator/emulator.cpp
```diff
@@ -193,7 +193,7 @@ auto Emulator::unload() -> void {
 auto Emulator::load(mia::Pak& node, string name) -> bool {
   Program::Guard guard;
   if(auto fp = node.pak->read(name)) {
-    if(auto memory = file::read({node.location, name})) {
+    if(auto memory = file::read({node.location, name}); !memory.empty()) {
```
jcm93: Noted off-thread, but I think it's better for readability if we don't hide multiple lines in one like this, and instead declare the variable outside the clause so the actual if statement is more quickly readable. There are approximately 11 instances of this pattern in this PR.

hiro/extension/table-layout.cpp
```diff
@@ -277,9 +280,9 @@ auto mTableLayout::setPadding(Geometry padding) -> type& {
 }
 
 auto mTableLayout::setParent(mObject* parent, s32 offset) -> type& {
-  for(auto& cell : reverse(state.cells)) cell->destruct();
-  for(auto& column : reverse(state.columns)) column->destruct();
-  for(auto& row : reverse(state.rows)) row->destruct();
+  for(auto& it : state.cells | std::views::reverse) it->destruct();
```
jcm93: These should probably still be named `cell`, `column`, `row` respectively?

hiro/core/widget/combo-button.cpp
```diff
@@ -66,7 +66,7 @@ auto mComboButton::selected() const -> ComboButtonItem {
 }
 
 auto mComboButton::setParent(mObject* parent, s32 offset) -> type& {
-  for(auto& item : reverse(state.items)) item->destruct();
+  for(auto& it : state.items | std::views::reverse) it->destruct();
```
jcm93: Should probably still be named `item`?

ruby/video/video.cpp
rasky: I think we should keep some local consistency at best, but we can't easily start mandating that either should be the preferred format. I can change this one instance but there will be one hundred other cases where it'll be using [0] anyway. In this specific case it's also checking for 1 in the same line, so it seems quite readable to use the indexing form.

ruby/input/input.hpp
rasky: done

ruby/video/video.hpp
rasky: done

ruby/video/video.hpp
rasky: done

ruby/audio/xaudio2.cpp
rasky: fixed

ruby/audio/waveout.cpp
rasky: done

ruby/audio/wasapi.cpp
rasky: Done

ruby/audio/oss.cpp
rasky: done

ruby/audio/openal.cpp
rasky: done

ruby/audio/audio.hpp
rasky: done

ruby/audio/audio.hpp
rasky: done

nall/nall/shared-pointer.hpp
rasky: it was unused, no vector in shared-pointer.hpp

nall/nall/directory.cpp
rasky: Done.

I'll notice that the original code didn't have a variable name either, and the temporary name is only required because we haven't converted string::split to std::vector yet. Once we convert that (which will be in the next round of vector PRs anyway), this code will go away and whatever name we chosen for a few days will be lost forever.

So whenever you see this pattern of explicit copy of a vector, I would suggest to keep this in mind, as the code will go aways very soon.

mia/medium/neo-geo.cpp
rasky: As far as I can tell, this change you suggest, if applied blindly to all similar cases, would change the semantic of the loop as int is signed, so it can't be done as a general replacement patterns without reconsidering each loop one by one. So I would suggest against this kind of changes.

mia/medium/mame.cpp
rasky: done

hiro/windows/application.hpp
rasky: done, albeit will change again if we drop std qualifier when we're done. Anyway it's not yet decided if we'll do, so better fix the code for now.

hiro/extension/table-layout.cpp
rasky: done

hiro/extension/horizontal-layout.cpp
rasky: done

hiro/extension/browser-dialog.cpp
rasky: this is one of the cases that will change again after migration is completed. It would probably become:

```c++
copy(split(part.right(), ":"), back_inserter(filters));
```

(after/if we drop all namespace qualifiers). 

I've opted for filterParts to go after your suggestion to name it plural, but I wouldn't elaborate too much on this as it'll go away.

hiro/cocoa/widget/table-view.cpp
rasky: See https://github.com/ares-emulator/ares/commit/ad2482fbc39427bafdeb03a5ffa27bf190724841. It's the only functional change mentioned in the PR description.

desktop-ui/presentation/presentation.cpp
rasky: Done for local consistency

ares/ares/node/class.hpp
rasky: this is already a method called `classes()`, so you would then have to use `this->classes()`, and then you would have `classes` and `classes()`. It seems more confusing in this case?

desktop-ui/emulator/emulator.cpp
rasky: Done this instance for now

hiro/extension/table-layout.cpp
rasky: Done

hiro/core/widget/combo-button.cpp
rasky: Done

nall/nall/directory.cpp
jcm93: Indeed; I ignored most boundaries between std::vector and nall vector.

ares/md/mcd/megald.cpp
jcm93: I cannot get the GitHub suggestion UI to co-operate with a multi-line block here. This is my intended cleanup:

```c++
  auto& mediaVec = mmi.media();
  auto media = *std::ranges::find_if(mediaVec, [mediaName](const auto& m){ return m.name == mediaName; });

  // Extract the stream information for analog video and audio
  string analogAudioFileName;
  string analogVideoFileName;
  video.isCLV = false;
  for(const auto& stream : media.streams) {
    if(stream.type == "Redbook") {
      mcd.fd = mcd.pak->read(stream.file);
      video.hasDigitalAudio = true;
    }
    if(stream.type == "RawAudio") analogAudioFileName = stream.file;
    if(stream.type == "RawVideo") {
      analogVideoFileName = stream.file;
      video.leadInFrameCount = stream.framesInLeadInRegion;
      video.activeVideoFrameCount = stream.framesInActiveRegion;
      video.leadOutFrameCount = stream.framesInLeadOutRegion;
      video.isCLV = media.format.endsWith("CLV");
```

hiro/cocoa/widget/table-view.cpp
jcm93: This sounds like it's in fact referring to the bug you have already fixed with removing elements from hiro collections directly rather than preserving calling the `remove` methods in those classes. Is that not the case? Or are we talking about some other separate bug? Removing this line does not seem to cause issues currently in my testing; if it is a different bug, is there a repro?

mia/system/master-system.cpp
rasky: Done

hiro/cocoa/widget/table-view.cpp
rasky: Yes you're right, the change isn't needed anymore. I've removed it from the PR.

ares/md/mcd/megald.cpp
rasky: I've added a fixme here, I will revisit this in next PRs or when the series is finished.

desktop-ui/emulator/emulator.cpp
rasky: Done all instances.

jcm93:
This breaks the build on BSD due to ruby/input/uhid.cpp not being migrated. It also breaks the build with Qt on Linux/BSD.

Issue #1976, opened 2025-05-01
[FC] Famicom keyboard is not connected to the desktop-ui input system


randomsomeone69-real:
**Describe the issue**
PS/2 keyboard input isn't printed in Family BASIC (v3.0) even with the keyboard attached and the input being active (toggled with the F9 key on my settings)

**Rom Information**
Sgame
  sha256: c8c0b6c21bdda7503bab7592aea0f945a0259c18504bb241aafb1eabe65846f3
  name:   Family BASIC V3 (Japan)
  title:  Family BASIC V3 (Japan)
  region: NTSC-J
  board:  HVC-NROM
    mirror
      mode: vertical
    memory
      type: ROM
      size: 0x8000
      content: Program
    memory
      type: RAM
      size: 0x1000
      content: Save
    memory
      type: ROM
      size: 0x2000
      content: Character
    memory
      type: ROM
      size: 0x10
      content: iNES
      data: 4e 45 53 1a 02 01 03 00 00 00 00 00 00 00 00 00

remutro:
The family basic keyboard, which was an additional peripheral that plugged into the Famicom expansion port, is currently not implemented.

randomsomeone69-real:
If the peripheric isn't implemented, why is it selectable in the menu then? Or maybe it is another keyboard?

![Image](https://github.com/user-attachments/assets/ae745d94-a7b0-4f1a-8be8-1b518d4f5014)

LukeUsher:
I think it's more correct to say it is emulated, but it was never connected to the ares input system, and still assumes a higan-style UI is present to map each keyboard key individually.

PR #2185, opened 2025-08-29
feat: add keyboard support

ZengGengSen:
1. rename family-keyboard to family-basic-keyboard
2. support family-basic-keyboard input
3. add split family-basic-keyboard mapping
4. add tape tools
5. add wav::encode
6. add family-basic-data-recorder

fix the issue *[FC] Famicom keyboard is not connected to the desktop-ui input system* #1976
support the issue *Add Family Basic Data Recorder support (HVC-008)*  #1976


LukeUsher:
Interesting approach of adding a virtual keyboard rather than directly using InputKeyboard like SG-1000 and MSX

Given how device-specific layouts can be, I'm not sure a virtual keyboard abstraction is a good idea if I'm honest; a better method long term might be to have per-console keyboard layouts with sane defaults and let the users override.

ZengGengSen:
> Interesting approach of adding a virtual keyboard rather than directly using InputKeyboard like SG-1000 and MSX
> 
> Given how device-specific layouts can be, I'm not sure a virtual keyboard abstraction is a good idea if I'm honest; a better method long term might be to have per-console keyboard layouts with sane defaults and let the users override.

I want to delete all the devices inside VirtualPort and let each device use its own configuration. In this way, listing the keyboard separately is more in line with the design, and each device can add its own default layout. However, if this is done, the code changes in the input part will be a bit significant. I'm not sure if the current situation can accept such a drastic change

PR #2189, opened 2025-08-30
Use QScreen for `Desktop::{size,workspace}`

jchv:
Qt 5.0 adds the QScreen API, which exposes enough information that `Desktop::{size,workspace}` no longer needs to fall back to Win32/Xlib. I manually tested that this seems to give the same exact answers as the Xlib path (though it technically goes a different route for `Desktop::size`.)

Since this API doesn't exist until Qt 5.0, this would break the Qt 4 build. However, Qt 4 is long dead now, and ares no longer has any official way to build Hiro with Qt 4 support. Therefore, I propose we finally clean up all of the old Qt 4 codepaths and avoid bifurcating the code any further: #2187

Mirrored in:
- bsnes: https://github.com/bsnes-emu/bsnes/pull/362
- higan: https://github.com/higan-emu/higan/pull/217

### DIFF
--- a/hiro/qt/desktop.cpp
+++ b/hiro/qt/desktop.cpp
@@ -3,57 +3,19 @@
 namespace hiro {
 
 auto pDesktop::size() -> Size {
-  #if defined(DISPLAY_WINDOWS)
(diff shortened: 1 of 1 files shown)


Ammar-Sadaoui:
is there any reason for keeping compatibility with qt4 ?

PR #2190, opened 2025-08-30
nall: remove unused modules

rasky:


nall/nall/beat/single/create.hpp
LukeUsher, line 1: Can we keep this file? It serves as a reference implementation of creating 'bps' format patches for distributing romhacks; and will be used again once we import 'beat' (the bps patching tool) into our tools

nall/nall/emulation/21fx.hpp
LukeUsher, line 1: This one should be kept as it is (optionally) used by the SNES core to implement the 21fx device, which is essentially a precursor to MSU1 and also a development tool that operated from the expansion port instead of from a cartridge.

LukeUsher (changes requested):
just a couple of files I'd like to keep for now

nall/nall/beat/single/create.hpp
rasky: done

nall/nall/emulation/21fx.hpp
ExperimentT020: I'm gobsmacked the 21fx code is even still there. Thought that was mothballed in bsnes back in 2009/2010.

PR #2192, opened 2025-08-31
gba: fix some warnings from C++20 deprecating bitwise operations between different enums

png183:
Fixes some of the C++20 deprecation warnings in the GBA core.


jcm93:
Hmm. It's only AppleClang complaining (currently), but it looks like there are still a couple ambiguous cases still left: https://github.com/ares-emulator/ares/actions/runs/17353822821/job/49263913044?pr=2192#step:10:627

PR #2208, opened 2025-09-08
stdc++: migrate desktop-ui and mia, add nall::merge/split and migrate some of the uses to them

rasky:
This is next step in the series. The main news here is that we start adding some global functions to nall. In general, nall containers have many more methods than standard C++ and those methods implement useful helpers that need to stay around. Since we can't add methods to standard classes, we'll migrate those to global functions instead. We start with merge/split of strings as they use vectors of strings as well, which we want to migrate to std::vector anyway.

~~Some uses have now been decorated with `::nall::split` for namespace issues that I want to postpone for later. They're pretty easy to find with a grep and fix later anyway.~~

### DIFF
--- a/ares/sfc/memory/memory.cpp
+++ b/ares/sfc/memory/memory.cpp
@@ -42,14 +42,14 @@ auto Bus::map(
   reader[id] = read;
   writer[id] = write;
 
-  auto p = ::nall::split(addr, ":", 1L);
+  auto p = nall::split(addr, ":", 1L);
   p.resize(2);
-  auto banks = ::nall::split(p[0], ",");
-  auto addrs = ::nall::split(p[1], ",");
+  auto banks = nall::split(p[0], ",");
+  auto addrs = nall::split(p[1], ",");
   for(auto& bank : banks) {
     for(auto& addr : addrs) {
-      auto bankRange = ::nall::split(bank, "-", 1L);
-      auto addrRange = ::nall::split(addr, "-", 1L);
+      auto bankRange = nall::split(bank, "-", 1L);
+      auto addrRange = nall::split(addr, "-", 1L);
       u32 bankLo = bankRange[0].hex();
       u32 bankHi = bankRange.size() > 1 ? bankRange[1].hex() : bankRange[0].hex();
       u32 addrLo = addrRange[0].hex();
@@ -78,14 +78,14 @@ auto Bus::map(
 }
 
 auto Bus::unmap(const string& addr) -> void {
-  auto p = ::nall::split(addr, ":", 1L);
+  auto p = nall::split(addr, ":", 1L);
   p.resize(2);
-  auto banks = ::nall::split(p[0], ",");
-  auto addrs = ::nall::split(p[1], ",");
+  auto banks = nall::split(p[0], ",");
+  auto addrs = nall::split(p[1], ",");
   for(auto& bank : banks) {
     for(auto& addr : addrs) {
-      auto bankRange = ::nall::split(bank, "-", 1L);
-      auto addrRange = ::nall::split(addr, "-", 1L);
+      auto bankRange = nall::split(bank, "-", 1L);
+      auto addrRange = nall::split(addr, "-", 1L);
       u32 bankLo = bankRange[0].hex();
       u32 bankHi = bankRange.size() > 1 ? bankRange[1].hex() : bankRange[0].hex();
       u32 addrLo = addrRange[0].hex();
--- a/ares/sfc/memory/memory.cpp
+++ b/ares/sfc/memory/memory.cpp
@@ -42,17 +42,18 @@ auto Bus::map(
   reader[id] = read;
   writer[id] = write;
 
-  auto p = addr.split(":", 1L);
-  auto banks = p(0).split(",");
-  auto addrs = p(1).split(",");
+  auto p = ::nall::split(addr, ":", 1L);
+  p.resize(2);
+  auto banks = ::nall::split(p[0], ",");
+  auto addrs = ::nall::split(p[1], ",");
   for(auto& bank : banks) {
     for(auto& addr : addrs) {
-      auto bankRange = bank.split("-", 1L);
-      auto addrRange = addr.split("-", 1L);
-      u32 bankLo = bankRange(0).hex();
-      u32 bankHi = bankRange(1, bankRange(0)).hex();
-      u32 addrLo = addrRange(0).hex();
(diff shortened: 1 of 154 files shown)


ares/sfc/memory/memory.cpp
```diff
@@ -42,17 +42,18 @@ auto Bus::map(
   reader[id] = read;
   writer[id] = write;
 
-  auto p = addr.split(":", 1L);
-  auto banks = p(0).split(",");
-  auto addrs = p(1).split(",");
+  auto p = ::nall::split(addr, ":", 1L);
+  p.resize(2);
```
jcm93, line 46: Is it necessary to `resize` after splitting here? In most areas this doesn't happen so it seems a bit incongruous.

hiro/extension/browser-dialog.cpp
```diff
@@ -308,7 +309,7 @@ auto BrowserDialogWindow::run() -> BrowserDialog::Response {
     if(batched.empty()) return;
     if(MessageDialog()
     .setTitle("Remove Selected")
-    .setText({"Are you sure you want to permanently delete the selected item", batched.size() == 1 ? "" : "s", "?"})
+    .setText({"Are you sure you want to permanently delete the selected item", batched.size() == 1 ? string{} : string{"s"}, "?"})
```
jcm93: Seems unnecessary.
```suggestion
    .setText({"Are you sure you want to permanently delete the selected item", batched.size() == 1 ? "" : "s", "?"})
```

nall/nall/decode/cue.hpp
```diff
@@ -139,7 +140,8 @@ inline auto CUE::loadTrack(std::vector<string>& lines, u32& offset) -> Track {
   Track track;
 
   lines[offset].itrimLeft("TRACK ", 1L).strip();
-  track.type = lines[offset].split(" ").last().strip().downcase();
+  auto parts2 = nall::split(lines[offset], " ");
+  track.type = parts2.empty() ? string{} : parts2.back().strip().downcase();
```
jcm93: Suffixed variable name looks unnecessary.
```suggestion
  auto parts = nall::split(lines[offset], " ");
  track.type = parts.empty() ? string{} : parts.back().strip().downcase();
```

nall/nall/decode/cue.hpp
```diff
@@ -178,7 +180,8 @@ inline auto CUE::loadIndex(std::vector<string>& lines, u32& offset) -> Index {
   Index index;
 
   lines[offset].itrimLeft("INDEX ", 1L);
-  string sector = lines[offset].split(" ").last().strip();
+  auto parts3 = nall::split(lines[offset], " ");
```
jcm93: Suffixed variable again unnecessary.
```suggestion
  auto parts = nall::split(lines[offset], " ");
  string sector = parts.empty() ? string{} : parts.back().strip();
```

nall/nall/string/transform/cml.hpp
```diff
@@ -71,8 +72,10 @@ inline auto CML::parseDocument(const string& filedata, const string& pathname, u
 
     if(name == "variables") {
       for(auto& line : lines) {
-        auto data = line.split(":", 1L).strip();
-        variables.append({data(0), data(1)});
+        auto dataParts = nall::split(line, ":", 1L);
+        std::vector<string> data;
+        for(auto& part : dataParts) data.push_back(part.strip());
+        variables.emplace_back(data.size() > 0 ? data[0] : string{}, data.size() > 1 ? data[1] : string{});
```
jcm93, line 76: Elsewhere if I'm not mistaken `insert` is used for this sort of operation; is there a reason `emplace_back` is used here instead?

nall/nall/string/transform/cml.hpp
```diff
@@ -88,8 +91,11 @@ inline auto CML::parseDocument(const string& filedata, const string& pathname, u
         continue;
       }
 
-      auto data = line.split(":", 1L).strip();
-      auto name = data(0), value = data(1);
+      auto dataParts2 = nall::split(line, ":", 1L);
```
jcm93: Is it better to reuse `dataParts` here instead?

desktop-ui/presentation/presentation.cpp
```diff
@@ -375,8 +377,10 @@ auto Presentation::loadEmulators() -> void {
     for(u32 index : range(count)) {
       MenuItem item{&recentGames};
       auto entry = settings.recent.game[index];
-      auto system = entry.split(";", 1L)(0);
-      auto location = entry.split(";", 1L)(1);
+      auto parts = nall::split(entry, ";", 1L);
+      parts.resize(2);
```
jcm93, line 381: Same comment here re: resizing.

ares/sfc/memory/memory.cpp
rasky:
```C
  auto p = nall::split(addr, ":", 1L);
  p.resize(2);
  auto banks = nall::split(p[0], ",");
  auto addrs = nall::split(p[1], ",");
```

`nall::split` (just like `string::split`) doesn't guarantee that the minimum number of splits actually happen, or rather, than the returned vector has got at least N+1 elements. So if `addr` doesn't contain a `:`, the vector `p` only has one element.

This was dealt with before with the `operator()` accessor. In the new world, it's either a resize or an explicit bound check, depending on the instance. I haven't been super religious here as it's not even obvious what looks better.

As a followup, I think it would be interesting to check if it'd be a good idea have `nall::split` autoresize the vector to `N+1` on output. It seems not a bad idea if you ask me, but in this conversion step I wanted to keep the same semantic of the helper function to avoid mistakes.

nall/nall/string/transform/cml.hpp
rasky: Normally it's push_back() for single elements at the end of the vector. I tend to use emplace_back() only when I'm slightly worried about performance (eg: vector of larger objects) or when I'm pushing a temporary whose constructor takes more than an argument, like in this case, so that you save the extra typing of the class name to name the constructor and/or a temporary named variable in one additional line.

nall/nall/string/transform/cml.hpp
rasky: Took the chance to migrate this and other cases to nall::split_and_strip.

ares/sfc/memory/memory.cpp
LukeUsher: I would lean towards either explicit bounds checks, or extending nall:split to take an "ouput size" argument rather than hard-coding n+1; so that we can do similar code patterns with any expected number of splits.

PR #2222, opened 2025-09-19
Initial working version of NEC PAC support for the LaserActive

RogerSanders:
This change adds some accuracy fixes and speed improvements to the MegaLD implementation, as well as adding initial working support for NEC LDrom2 titles. The implementation is currently duplicated between Sega and NEC PAC modules. A following change will attempt to unify them with a reusable component for the LaserActive interface IC.

### DIFF
--- a/ares/pce/system/system.hpp
+++ b/ares/pce/system/system.hpp
@@ -2,7 +2,7 @@ struct System {
   Node::System node;
   VFS::Pak pak;
 
-  enum class Model : u32 { PCEngine, PCEngineDuo, SuperGrafx };
+  enum class Model : u32 { PCEngine, PCEngineDuo, SuperGrafx, LaserActive};
   enum class Region : u32 { NTSCJ, NTSCU };
 
   auto name() const -> string { return information.name; }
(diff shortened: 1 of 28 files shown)


ares/pce/system/system.hpp
```diff
@@ -38,7 +38,8 @@ struct System {
 extern System system;
 
 auto Model::PCEngine()    -> bool { return system.model() == System::Model::PCEngine;    }
-auto Model::PCEngineDuo() -> bool { return system.model() == System::Model::PCEngineDuo; }
+auto Model::PCEngineDuo() -> bool { return (system.model() == System::Model::PCEngineDuo) || (system.model() == System::Model::LaserActive); }
```
LukeUsher, line 41: nit: we'll probably want to clean this up at some point because, despite sharing functionality, PC Engine Duo and LaserActive are two different models.

LukeUsher (commented):
My only real concern right now is that this only implements laserdisc video rendering for the pixel accurate VDP, not the scanline based implementation.

This is fine at the moment because we currently disable the scanline based implementation due to it having some rather severe bugs, but if we were ever to restore it, this would need addressing.

Not a problem for right now though, and not something I think is worth delaying the merge for.

PR #2229, opened 2025-09-22
ld: Added hash for NEC PAC-N10 bios

RogerSanders:
I received a bios dump for v1.02 of the NEC PAC-N10 bios today, which is the first US region NEC PAC bios that's been dumped. This change adds the hash to the list of known hash values for detection purposes in Ares.


RogerSanders:
The hash-based detection should be changed in the future, but this will serve for now to get NEC PAC emulation out there in an initial form.

PR #2239, opened 2025-09-28
hiro: enable GTK2 support

### LINKED ISSUE #2235: Question: How can one build Ares with GTK2?
Even if i edit `cmake/finders/FindGTK.cmake` and `hiro/cmake/os-linux.cmake` to refer to GTK2 instead of GTK3, it fails:
```
CMake Error at /usr/local/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find GTK (missing: GTK_LIBRARY) (found version "2.24.33")

      Reason given by package: Ensure GTK libraries are available in local library paths.

Call Stack (most recent call first):
  /usr/local/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  cmake/finders/FindGTK.cmake:59 (find_package_handle_standard_args)
  hiro/cmake/os-linux.cmake:7 (find_package)
  hiro/CMakeLists.txt:13 (include)
```
I tried to check if GTK2 support was dropped, but i only found that GTK3 is now the default, but not the only one. (The source is still full of GTK2 code.)

bsdcode:
This PR enables GTK2 support in the CMake build system.

The new boolean build option `USE_GTK2` enables the GTK2 UI backend. It is prioritized over the `USE_QT6` build option if both are enabled at the same time. The sequence of conditional blocks in `hiro/cmake/os-linux.cmake` is rearranged for more clarity (`if(USE_GTK2)`, `elseif(USE_QT6)`, `else() -> GTK3` instead of `if(NOT USE_QT6 AND NOT USE_GTK2) -> GTK3`, `elseif(NOT USE_QT6) -> GTK2`, `else() -> Qt6`).

The GTK2 CMake package finder is a copy of the GTK3 package finder with necessary changes to path lockups. It creates the `GTK2::GTK2` target instead of `GTK::GTK` from the GTK3 package finder.

`HIRO_BACKEND` is set to `GTK` for both GTK backends and therefore `tools/genius` is also build when using GTK2.

A CI action is added for the GTK2 build. This is a copy of the QT CI action.

Fixes #2235


LukeUsher:
Given that GTK2 has not seen an official update since 2020, and that it was declared end of life around that time, I don't believe we should merge this.

The current support for GTK2 is more incidental (due to not being pruned over time) than intentional, and it is unlikely that we will maintain it over time either.

PR #2241, opened 2025-09-30
ld: Added frame prefetch and video resampling coefficient precalculation

RogerSanders:
This change brings speed enhancements to LaserActive games. The linear resampling coefficient precalculation reduces overall CPU overhead by approximately 30%, making slower CPUs much more likely to achieve full framerate. Additionally, frame prefetch using a background thread makes games much more tolerant of IO latency, making it possible to play games back from platter drives over SATA3.

This should be sufficient to make emulation performance acceptable on 95%+ of systems, and I don't have any further optimizations planned at this stage.


LukeUsher:
Looks like we may need to bump macOS minimum version to 11 (Big Sur) to get access to C++20 std::atomic::wait/notify_all, or alternatively rework to avoid these functions.

Given that macOS 11 is 5 years old already, and that it still supports 2013/2014 computers; I think it's okay to bump our minimum requirements here.

jcm93:
Given that the change enables Mega-LD to work better on older hardware, I think it's unfortunate that this change would exclude a bunch of the older Mac hardware that it may have otherwise given a shot to run the core. I don't know if it's worth rearranging the synchronization code for those machines sake though.

LukeUsher:
I think this discord message from @RogerSanders is enough justification

> It's not impossible, but the code gets more bloated and uglier, and easier to break. I'd have to keep the std::atomic_flag members, but put parallel std::condition_variable members, as well as a std::mutex and an explicit releaseable scoped lock. I'd take a lock, change the flag,  trigger the condition variable, then release the lock. Waiting threads would need to spin on the condition and handle spurious wake events. A lot of bloat, very hard to follow, and easy to break if you don't understand threaded synchronisation inside and out. Not really the kind of code you want to introduce and proliferate if you can avoid it.

PR #2245, opened 2025-10-03
GCC11 - Fix ambiguous overload on mouseCursor types. Always compare with matc…

DirtBagXon:
…hing string types.

Fixes:
```
widget.cpp:142:20: error: ambiguous overload for ‘operator==’ (operand types are ‘const hiro::MouseCursor’ and ‘const nall::string’)
 142 | if(mouseCursor == MouseCursor::Hand) name = "hand1";
  | ~~~~~~~~~~~ ^~ ~~~~~~~~~~~~~~~~~
  | |           |
  | const hiro::MouseCursor const nall::string
```

As `mouse-cursor.hpp` defines `name` as `MouseCursor(const string& name = "");` seems like direct comparison via type should be made on these 3 checks.

### DIFF
--- a/hiro/gtk/widget/widget.cpp
+++ b/hiro/gtk/widget/widget.cpp
@@ -139,9 +139,9 @@ auto pWidget::setMouseCursor(const MouseCursor& mouseCursor) -> void {
 
   if(mouseCursor) {
     string name;
-    if(mouseCursor == MouseCursor::Hand) name = "hand1";
-    if(mouseCursor == MouseCursor::HorizontalResize) name = "sb_h_double_arrow";
-    if(mouseCursor == MouseCursor::VerticalResize) name = "sb_v_double_arrow";
+    if(mouseCursor.name() == MouseCursor::Hand) name = "hand1";
+    if(mouseCursor.name() == MouseCursor::HorizontalResize) name = "sb_h_double_arrow";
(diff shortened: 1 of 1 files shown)


jcm93:
What compiler/toolchain/build environment are you seeing this error under?

DirtBagXon:
Hey, this was under a current LTS Ubuntu: 24.04.3 LTS

cmake version 3.28.3
gcc version 13.3.0 (Ubuntu 13.3.0-6ubuntu2~24.04)

jcm93:
The change seems fine to me. I would be curious to see the complete output from your configuration and build steps that trigger this error (and building with `-v`) in a fresh build directory, though. The gcc version matches the one we use on CI, so I am curious why you in particular would see this error.

DirtBagXon:
Ughh, it appears that the terminal I was compiling in was falling back to using `gcc version 11.5.0 (Ubuntu 11.5.0-1ubuntu1~24.04)`. As you state GCC13 doesn't see this issue. Not certain what your minimal gcc requirement is currently, I see cmake version noted in docs, but not gcc. To note, the above changes were all that was needed to successfully compile on this older gcc, currently.

jcm93:
No worries. Now that we've switched to compiling in C++20, we'll probably have to do some legwork to figure out the actual minimum versions of our supported compilers (GCC, Clang, AppleClang, MSVC) that can compile ares (previously, the answer seemed to be basically anything).

PR #2257, opened 2025-10-15
[SFC] Resize board vector so that there are 2 entries minimum

remutro:
When using the operator() accessor for the nall::vector implementation, it would always allocate up to the offset requested so there was always a valid result returned. std::vector does not behave like this. The code to analyze SFC boards in mia as written always expects this boards vector to always have at least two valid entries. So there is a check now that will resize this vector if it is smaller.

### DIFF
--- a/mia/medium/super-famicom.cpp
+++ b/mia/medium/super-famicom.cpp
@@ -255,6 +255,7 @@ auto SuperFamicom::analyze(std::vector<u8>& rom) -> string {
   s +={"  board:    ", board(), "\n"};
 
   auto board = nall::split(this->board().trimRight("#A", 1L), "-");
+  if(board.size() <= 1) { board.resize(2); }
 
   if(auto size = romSize()) {
     if(board[0] == "SPC7110" && size > 0x100000) {


mia/medium/super-famicom.cpp
```diff
@@ -255,6 +255,7 @@ auto SuperFamicom::analyze(std::vector<u8>& rom) -> string {
   s +={"  board:    ", board(), "\n"};
 
   auto board = nall::split(this->board().trimRight("#A", 1L), "-");
+  if(board.size() <= 1) { board.resize(2); }
```
rasky, line 258:
```suggestion
  board.resize(2);
```

The if is not necessary, and we want to settle on a standard pattern.

remutro: This won't work. The minimum number of entries needs to be two. But there could be more and if we blindly resize to two we will throw away information we need. For example, a perfectly valid board string prior to splitting could read:

LOROM-SA1-RAM-EPSONRTC

rasky: ah right, the split doesn't have a maxsize which is why i hadn't noticed this case before. It's good as is then

PR #2266, opened 2025-10-22
Fix glitch in City Hunter ROM in the intro in the performance-mode VDP

Florian-Simon-RedArtGames:
We were rendering background and sprites in burst mode, which shouldn't happen.

I'm not the author, I'm just upstreaming changes done a member in our team.


Florian-Simon-RedArtGames:
It looks like we're missing the code that serializes the new data members. This needs to be added.

Florian-Simon-RedArtGames:
Everything is here now!

PR #2298, opened 2025-11-23
n64: handle correct R/W to RDP spans in test mode

HailToDodongo:
Fixes a wrong implementation of span buffer access when in RDP test mode.
On hardware, the registers for doing so act as a view into the larger buffer in the RDP itself.
Ares so far assumed it only can hold a single value directly, and was missing masking (https://n64brew.dev/wiki/Reality_Display_Processor/Interface#0x0420_000C_-_DPS_BUFTEST_DATA)

This is tested by a demo i made here: https://github.com/HailToDodongo/repeater64

### DIFF
--- a/ares/n64/rdp/io.cpp
+++ b/ares/n64/rdp/io.cpp
@@ -137,7 +137,7 @@ auto RDP::IO::readWord(u32 address, Thread& thread) -> u32 {
 
   if(address == 3) {
     //DPS_BUFTEST_DATA
-    data.bit(0,31) = test.data;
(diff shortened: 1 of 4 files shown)


LukeUsher (changes requested):
This looks good, although you should bump the serializer version since serialization has changed.

https://github.com/ares-emulator/ares/blob/e415dd9db7093550c7b15b20c660dd789773366f/ares/n64/system/serialization.cpp#L1

HailToDodongo:
increased the version

PR #2337, opened 2025-12-26
gba: only trigger keypad IRQ when input changes

png183:
Fixes a bug where keypad IRQs would repeatedly trigger as long as the required inputs were held.


remutro:
This fixes input issues in all 4 regions of Chessmaster (no open issue currently).

Issue #1853, opened 2025-02-21
[GBA] Kong - The Animated Series (USA) (Europe) Broken Graphics


ExperimentT020:
**Describe the issue**
Kong - The Animated Series (USA) and it's European counterpart are currently unplayable due to completely broken graphics. This behaviour matches Issue #211 and #868 possibly due to being by the same developer/game engine?

**Rom Information**
name:   Kong - The Animated Series (USA) (En,Fr,De,Es,It,Nl)
  title:  Kong - The Animated Series (USA) (En,Fr,De,Es,It,Nl)
  sha256: 594254b0d11ed798b98c96a79335cced86828c57f6c5ccc94c1082179ccdae9d
 
name:   Kong - The Animated Series (Europe) (En,Fr,De,Es,It,Nl)
  title:  Kong - The Animated Series (Europe) (En,Fr,De,Es,It,Nl)
  sha256: dbe775e3d396bddff1a11d77cb4af94c93960bf0519b9f1e810a28917dc3d534

![Image](https://github.com/user-attachments/assets/617ea7a4-41f4-4133-8746-0ca533cbe3f5)

![Image](https://github.com/user-attachments/assets/07434b6f-d083-47f0-b2d0-94323f331ec5)

remutro:
Fixed in commit: https://github.com/ares-emulator/ares/pull/2338/commits/78d77f80dfda0df8db533b727fa525d49214d377

Issue #1872, opened 2025-03-02
[GBA] NFL Blitz 2002 (USA) hangs indefinitely after the Midway logo


ExperimentT020:
**Describe the issue**
This game cannot currently reach the title screen or enter gameplay. It hangs indefinitely after the Midway logo displays and fades.

![Image](https://github.com/user-attachments/assets/074c0914-373a-4aef-95e8-0e1f85d19e20)

**Rom Information**
```
name:   NFL Blitz 2002 (USA)
title:  NFL Blitz 2002 (USA)
sha256: 0a4164d428210b9374729b1de01a79e45ce2a4830783eb859a5b8dbab46b856a
board
memory
type: ROM
size: 0x400000
content: Program
mirror: false

```

remutro:
Fixed in commit: https://github.com/ares-emulator/ares/pull/2338/commits/78d77f80dfda0df8db533b727fa525d49214d377

Issue #868, opened 2022-11-03
[GBA] Green screen with Casper (Europe) (En,Fr,De,Es,It,Nl,Pt) and (U)...


BenjaminSiskoo:
Emulator : Ares 1.29
Nightly build 2022-10-31 08:12:03 UTC
Options : Default
Game :  (Europe) (En,Fr,De,Es,It,Nl,Pt), Casper (USA) (En,Fr,Es)
System : GBA

![image](https://user-images.githubusercontent.com/12481242/199726703-34902afb-b406-4407-a32e-632ab63d796d.png)

sczther:
Retested with v141. Still leaves a green screen on both versions. If you press start, the screen blinks white, but it doesn't help.

remutro:
Fixed in commit: https://github.com/ares-emulator/ares/pull/2338/commits/78d77f80dfda0df8db533b727fa525d49214d377

Issue #211, opened 2021-10-01
[GBA] V.I.P. (Europe): Graphics corrupted


Shideravan:
The game seems to be running in the background but the image is bugged.
![image](https://user-images.githubusercontent.com/12163923/135564130-f41f7525-ac05-484f-b7f1-91abc1c4a7df.png)

sczther:
Still happens in v130.1

sczther:
Retested in v141. On boot, it shows just a green screen and a flash when start is pressed. After a while a black "Press Start" appears and then the screen from the first post with the black pixel mess appears.

![Image](https://github.com/user-attachments/assets/e8ae3e78-f439-4a48-b868-7a479cf50b3e)

Feels very similar to #868

ExperimentT020:
Issue still occurs as of the latest nightly builds.

![Image](https://github.com/user-attachments/assets/10d50a6b-aadb-4354-9ffd-ef99bfc9adbe)

remutro:
Fixed in commit: https://github.com/ares-emulator/ares/pull/2338/commits/78d77f80dfda0df8db533b727fa525d49214d377

PR #2338, opened 2025-12-27
gba: fix handling of repeat bit during immediate DMA transfers

png183:
The repeat bit should not have any effect on immediate DMA transfers.

Fixes multiple issues, including #868 and #1872.


remutro:
Also fixes: https://github.com/ares-emulator/ares/issues/1853 & https://github.com/ares-emulator/ares/issues/211

Issue #2246, opened 2025-10-03
[32X] H32 pixel hangover seems to be slightly off


dgrove727:
**Describe the bug**
The H32-on-32X horizontal adjustment seems to be off somehow. In the ROM I'm working with, I have black sprites on the left side specifically to hide the ~3.25 pixel hangover on where the Genesis video would be visible. On hardware, it's enough to hide it. On Ares, a single column of pixels is still visible.

**To Reproduce**
I have a ROM you can use to easily reproduce the issue if needed.

**Expected behavior**
On hardware, the black sprites successfully cover up the H32 video displacement on the left side. You can't even tell there's anything there.

**Screenshots**
<img width="91" height="273" alt="Image" src="https://github.com/user-attachments/assets/21acc91f-5404-49a8-bb40-0e149c950252" />

**Additional context**

jsgroth:
I was curious to have a more robust test case for this than eyeballing pixels in NFL Quarterback Club, so I threw together a test ROM that exercises this behavior:

[32x-h32.zip](https://github.com/user-attachments/files/24510106/32x-h32.zip)

This simply has both VDPs display a static image while the Genesis/MD VDP is in H32 mode.

The left half of the screen has the 32X VDP rendering solid white over the Genesis VDP rendering alternating red and blue vertical lines, each 1px wide. The right half is reversed, with the Genesis rendering white over the 32X rendering red and blue lines.

Here are some (unfortunately very bad) pictures of this running on actual Genesis+32X hardware, connected via component so that you can actually make out the vertical lines:

Left edge (red/blue lines from Genesis H32):

<img width="1890" height="1421" alt="Image" src="https://github.com/user-attachments/assets/195781f6-b069-4f4b-a176-a3789351aca1" />

Right edge (red/blue lines from 32X):

<img width="1889" height="1418" alt="Image" src="https://github.com/user-attachments/assets/f33ce011-4394-48c7-81e3-d302e752845e" />

There's also a graphical artifact in the center that I couldn't get a good picture of, where it switches from 32X white to Genesis white. That could very possibly just be my video connection or TV though.

This is the current output in ares, as of commit https://github.com/ares-emulator/ares/commit/17ce1bbe0f8e4ba08407505cae96e0e1450caf2b:

<img width="1168" height="976" alt="Image" src="https://github.com/user-attachments/assets/60d63b45-a596-4495-bcdc-ac20a5bd6b9b" />

[...]

PR #2362, opened 2026-01-08
m32x: adjust H32 pixel offset between Gen/MD VDP and 32X VDP

jsgroth:
The adjusted offset is equal to 2.6 H32 pixels / 3.25 H40 pixels.


jsgroth:
More details in https://github.com/ares-emulator/ares/issues/2246#issuecomment-3726320604

PR #2384, opened 2026-01-20
Sort the emulator list in firmware and emulators settings window

remutro:
Present the list of systems in the firmware settings window alphabetically.


desktop-ui/emulator/emulators.cpp
```diff
@@ -258,4 +258,6 @@ auto Emulator::construct() -> void {
   emulators.push_back(std::make_shared<ZXSpectrum>());
   emulators.push_back(std::make_shared<ZXSpectrum128>());
   #endif
+
+  std::sort(emulators.begin(), emulators.end(), emuComparer);
```
LukeUsher: unless we need to reuse this comparer elsewhere; consider a lambda

remutro: Updated to use a lambda instead.

PR #2407, opened 2026-02-04
Windows OpenGL 4.6 Core Profile Driver

remutro:
This adds an OpenGL 4.6 video driver for Windows users. This is to allow for shaders that require support for newer versions of OpenGL in order to function properly. Default/optimal driver on Windows will remain as OpenGL 3.2. 

Linux may benefit from additional changes, but this is currently unclear and more testing on more hardware/drivers is required. Using the latest Mesa drivers, requesting the 3.2 version as we do currently enables 4.5 functionality, which should be new enough for any shader. Nvidia, AMD, & Intel hasn't been tested yet. A debug print statement has been added that will output OGL version and GLSL version to assist with gathering info going forward. 

macOS supports up to OpenGL 4.1 and has been deprecated for a long time. Use the Metal driver.

### DIFF
--- a/ruby/video/glx.cpp
+++ b/ruby/video/glx.cpp
@@ -276,6 +276,8 @@ private:
       return false;
     }
 
+    print("OpenGL Version: ", (const char*)glGetString(GL_VERSION), ", GLSL: ", (const char*)glGetString(GL_SHADING_LANGUAGE_VERSION),"\n");
+
     glXSwapInterval(self.blocking);
 
     //read attributes of frame buffer for later use, as requested attributes from above are not always granted
--- a/ruby/video/video.cpp
+++ b/ruby/video/video.cpp
@@ -197,7 +197,8 @@ auto Video::create(string driver) -> bool {
   #endif
 
   #if defined(VIDEO_WGL)
-  if(driver == "OpenGL 3.2") self.instance = std::make_unique<VideoWGL>(*this);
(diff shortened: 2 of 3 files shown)


Tino-Medina:
OpenGL 3.2 and 4.6 do not display video on v148....but v147 ran just fine with 3.2.
DirectX 9 works normal in v148, but of course, shaders cannot be used.

Running fully updated Win10, 4690K CPU @3.5Ghz, , 950 GTX GPU. ALL VIDEO DRIVERS ARE UP TO DATE.

Again, this wasn't an issue with v147, but since "upgrading" to 148, I can no longer use OpenGL as my drivers.

Tino-Medina:
Update:
After some trial and error, have discovered that OpenGL 3.2 and 4.6 DOES work in v148.
However, the way it works isn't exactly intuitive.

Before, no matter which video driver you chose, if shaders were available, you had the option to choose one, or NONE for the stock pixelated look.

Now, if you choose OpenGL 3.2 or 4.6, and leave the shaders on NONE (expecting the stock pixels), you get....a blank screen.
You actually have to choose one now to get the driver to display!

Again, it works, but, well, feels weird this way, and just discovered this by accident by switching settings on and off all over with regards to video!

Anyways...Ares, still a great emulator...I have loved it for a long time, now with just a touch of awkwardness. =)

PR #2414, opened 2026-02-09
Adding StartupWMClass

Kyuyrii:
I created a Snap version of Ares, but I noticed that KDE wasn't displaying the correct icon; this addition is to try and prevent that problem from happening.

### DIFF
--- a/desktop-ui/resource/ares.desktop
+++ b/desktop-ui/resource/ares.desktop
@@ -6,3 +6,4 @@ Icon=ares
 Terminal=false
 Type=Application
 Categories=Game;Emulator;
+StartupWMClass=Ares


Kyuyrii:
Adding some information here, this WMClass is probably just "Ares" because Ares isn't natively integrated into Wayland in version 147.

When it's migrated to Wayland, I suppose this will change.

remutro:
This is nit-picky, but ares isn't capitalized.

Kyuyrii:
> This is nit-picky, but ares isn't capitalized.

When I checked what the WMClass was in KDE, the WMClass was ```Ares``` instead of ```ares```.

PR #2426, opened 2026-02-23
build: change directory structure

gnaghi:
higan, ares'predecessor was designed to work with external libraries, with one frontend (higan-ui, then lucia, then desktop-ui).
This commit make a clear separation between the frontends (desktop-ui, or any other frontend), and the various library components (ares, mia, hiro, nall, libco, ruby).
This makes it straightforward to add additional frontends. 
To select a frontend, add `-DARES_FRONTEND=nameoffrontend` to the command `cmake .. -G Ninja` . If no frontend is specified, desktop-ui will be compiled by default.
A simple SDL frontend is available on my ares repository.


LukeUsher:
I understand the intention, but we will not be merging this.

Adding a new frontend does not require this kind of overhaul, the current structure is sufficient. The `-ui` suffix is enough to distinguish a frontend from a library, and there is no ambiguity that needs solving.

Right now, we only have a single frontend `desktop-ui`, if we ever introduce another one, it can simply follow the same naming pattern (eg: `sdl-ui` in your case), and this does not require splitting the entire project into new top-level directories.

PR #2430, opened 2026-02-24
desktop-ui: add --batch minimal UI mode

awsms:
Introduce a `--batch` command-line option to run with a minimal presentation for frontend-driven launches (eg EmulationStation, arcade cabinets etc).

Behavior
- `--batch` implies `--no-file-prompt`
- Requires at least one valid game path (otherwise exits with a clear error)
- Uses a minimal presentation:
  - hides menu bar
  - removes status bar
  - black background / centered viewport
  - keeps file-open and drag/drop support
- Skips constructing settings/tools/game-browser windows in batch

Runtime safety
- Batch mode avoids UI-only flows where possible (logs messages instead)
- Driver and load/identify failures trigger a clean batch exit path
- Runtime defocus checks now use `settings.input.defocus` instead of `driverSettings.*.checked()`:
  - avoids dereferencing non-constructed UI widgets in batch
  - fixes batch crash from UI-state reads on non-UI code paths

### DIFF
--- a/desktop-ui/desktop-ui.cpp
+++ b/desktop-ui/desktop-ui.cpp
@@ -80,8 +80,8 @@ auto nall::main(Arguments arguments) -> void {
     program.startPseudoFullScreen = true;
   }
 
-  if(arguments.take("--batch")) {
-    program.batchMode = true;
+  if(arguments.take("--kiosk")) {
+    program.kioskMode = true;
     program.noFilePrompt = true;
   }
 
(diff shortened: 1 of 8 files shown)


flagrama:
"--batch" sounds like a window-ism due to it being something you'd use in batch files. This terminology makes no sense for a multi-platform software.

remutro:
I think there is a better name that can be used for this option other then batch as it really doesn't describe what it does. 

The README.md file will also need to be updated with the new command line option.

LukeUsher:
Sorry for the 'ready to review' toggle, I misclicked.

I like this, but I would recommend using something like `--kiosk` over `--batch`, it's less windows-centric and likely to be understood as full screen, no window frame, exit via hotkey, etc.

awsms:
Oops first time using Github's CLI tools, I thought `gh api -X POST repos/awsms/ares/branches/batch-mode/rename -f new_name=feat/kiosk-mode` would keep this PR opened :smile:

PR #2431, opened 2026-02-24
Linux Qt builds for Wayland + ignore global hotkeys when ares is not focused

awsms:
The first commit prevents ares hotkeys from being triggered when ares is on a different workspace (noticed on Hyprland).

Second one prevents ares' Qt build from crashing on startup on wayland (also noticed on Hyprland):
```
QWidget::paintEngine: Should no longer be called
QWidget::paintEngine: Should no longer be called
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  3 (X_GetWindowAttributes)
  Resource id in failed request:  0x479bf890
  Serial number of failed request:  27
  Current serial number in output stream:  28
QThreadStorage: entry 1 destroyed before end of thread 0x5611475a0ba0
QThreadStorage: entry 0 destroyed before end of thread 0x5611475a0ba0
beginResetModel called; about to emit modelAboutToBeReset
endResetModel called; about to emit modelReset
beginResetModel called; about to emit modelAboutToBeReset
endResetModel called; about to emit modelReset
beginResetModel called; about to emit modelAboutToBeReset
endResetModel called; about to emit modelReset
QBasicTimer::start: QBasicTimer can only be used with threads started with QThread
beginResetModel called; about to emit modelAboutToBeReset
endResetModel called; about to emit modelReset
beginResetModel called; about to emit modelAboutToBeReset
endResetModel called; about to emit modelReset
fish: Job 1, './ares…' terminated by signal SIGSEGV (Address boundary error)
```

### DIFF
--- a/desktop-ui/input/input.cpp
+++ b/desktop-ui/input/input.cpp
@@ -144,7 +144,7 @@ auto InputDigital::value() -> s16 {
     auto& groupID = binding.groupID;
     auto& inputID = binding.inputID;
     auto& qualifier = binding.qualifier;
-    if (device->isKeyboard() && program.keyboardCaptured) continue;
+    if (device->isKeyboard() && (program.keyboardCaptured || (!presentation.focused() && !ruby::video.fullScreen()))) continue;
     s16 value = device->group(groupID).input(inputID).value();
     s16 output = 0;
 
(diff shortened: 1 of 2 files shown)


awsms:
note: wait before a merge as I can still fire the hotkeys from other workspaces if ares is in fullscreen

flagrama:
FYI the reason nobody has committed the "xcb" environment variable "fix" for Qt is because it is deprecated and will be removed from Qt itself in the future.

awsms:
> FYI the reason nobody has committed the "xcb" environment variable "fix" for Qt is because it is deprecated and will be removed from Qt itself in the future.

Ohhhhhh thanks for the info, i'll just ditch it then.

flagrama:
I don't think you have to remove it, it just isn't going to be a permanent fix and the issue will crop up again. Just making sure you're aware of that.

PR #2433, opened 2026-02-26
Introducing --kiosk mode

awsms:
Discussion continuing from #2430


remutro:
The README.md file needs to be updated with the new command line option.

awsms:
I think we're good. Unless I missed another one or more, https://github.com/ares-emulator/ares/compare/bba093de4df3fa45c82713c9ce57166445cbe8bd..15b9bdcce20e54b081c0d18ebdb0c50f236c682f was the last call that could trigger a UI item to appear on-screen while `--kiosk` is active.

desktop-ui/emulator/emulator.cpp
```diff
@@ -253,7 +253,12 @@ auto Emulator::setColorBleed(bool value) -> bool {
 }
 
 auto Emulator::error(const string& text) -> void {
-  MessageDialog().setTitle("Error").setText(text).setAlignment(presentation).error();
+  if(program.kiosk) {
+    fprintf(stderr, "error: %s\n", text);
```
LukeUsher: ares strings are nall::string, not std::string or c-strings and are not compatible; if you want a null-terminated C string you can use .data()

awsms: lol just saw your comment, check out the last one, I refactored the whole error logic instead of repeating the same block 3*.

flagrama: > lol just saw your comment, check out the last one, I refactored the whole error logic instead of repeating the same block 3*.

Once you have created a PR, please don't force push unless requested. It makes it harder for other people to work with and follow.

awsms: > > lol just saw your comment, check out the last one, I refactored the whole error logic instead of repeating the same block 3*.
> 
> Once you have created a PR, please don't force push unless requested. It makes it harder for other people to work with and follow.

arf sorry, first time I work with others on a PR. did it so we could end up with a single commit after merge.

PR #2435, opened 2026-02-27
desktop-ui: prevent hotkey from firing when setting one

awsms:
I was constantly getting trolled, with hotkeys triggering a ms after I set them up in the Hotkeys menu. This fixes it.

### DIFF
--- a/desktop-ui/input/hotkeys.cpp
+++ b/desktop-ui/input/hotkeys.cpp
@@ -179,6 +179,7 @@ auto InputManager::createHotkeys() -> void {
 
 auto InputManager::pollHotkeys() -> void {
   if(Application::modal()) return;
+  if(settingsWindow.focused()) return;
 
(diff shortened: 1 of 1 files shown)


awsms:
Looks like I just had a faulty configfile for months. :melting_face:

awsms:
My bad, it's still reproducible easily, set any Input driver (any but `None` obviously) and enable "Allow Input" when focus is lost. Definitely not the behavior we want, where pressing/typing `spacebar` *ANYWHERE* in the settings window fires the command bind to the `spacebar` key.

awsms:
This seems to be a Linux (GLX)-only thing, which explains why nobody ever mentioned that issue before. Can't reproduce it on windows builds with wine either.

remutro:
Modification allows for setting hotkeys without polling at hotkeys while input is allowed. Works as expected.

Issue #2215, opened 2025-09-13
Upgrade Direct3D driver


randomsomeone69-real:
**Is your feature request related to a problem? Please describe.**
The current Direct3D driver used is based on DirectX 9, which is an outdated version of DirectX, and many emulators doesn't even support DirectX 9 today.

**Describe the solution you'd like**
Upgrade the current Direct3D driver from Direct3D 9 to Direct3D 11.

**Describe alternatives you've considered**
An alternate solution would be to provide a Direct3D 10 or 11 drivers while retaining Direct3D 9 for older systems that cannot run the latest Direct3D drivers. 

**Additional context**
Direct3D 12 is something I did consider but didn't recommend because it isn't compatible with Windows 7.

thekovic:
What version of Direct3D ares uses is mostly irrelevant because the video driver is only used to blit the final image to the screen. All ares cores are purely software rendered (with the exception of N64 which is hardware accelerated via ParaLLEl-RDP and uses Vulkan). DirectX9 is still included by default in latest Windows and supported by GPU drivers. Upgrading DirectX version is low priority.

remutro:
Direct3D 9 doesn't allow for shader use, so there is some incentive to upgrade.

PR #2438, opened 2026-02-28
Direct3D11 Video Driver Testing

remutro:
This introduces a Direct3D 11 video driver. The purpose of this draft PR has two main purposes:

- Farm out testing to different Windows systems with different video drivers and report results (please help test if possible). You will want to start ares with the `--terminal` command so that if issues arise you can pass on the output messages from the terminal. Do anything and everything you can think of that may interfere with video display (sync, full screen, resizing, run-ahead, aspect, shaders, etc.)
- Get some eyes on the source to get feedback by people who have more experience in this area to see if there is something that can be improved or isn't correct. 

As of June 4, 2026 everything in this PR currently works properly except for unloading the driver and switching to a different one.

NOTE: Exclusive full screen is not available with this driver. This is on purpose. If curious about why it is explained here: https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/for-best-performance--use-dxgi-flip-model

Related issue: https://github.com/ares-emulator/ares/issues/2215


thekovic:
Because users (particularly Windows users) are used to exclusive fullscreen being a thing, I've seen some DX11+ games still presenting an "exclusive fullscreen" option but it silently just doesn't do anything different.

PR #2442, opened 2026-03-03
n64: implement eeprom write timing and status flags

rasky:
Currently Ares implements an infinitely fast EEPROM, but alas real EEPROMs can take even 20 ms to save a single block of 8 bytes.

This commit implement approximate timings and the related flag to notify the CPU that the EEPROM is busy writing data.


remutro:
Do the new values in cartridge.hpp need to be serialized out as part of save states?

rasky:
right fixed thanks

remutro:
Also need to bump the serializer version in n64/system/serialization.cpp

rasky:
Done

Issue #2365, opened 2026-01-09
Don't Register keystrokes when window is not focused in fullscreen.


OpenRift412:
I'll be playing a game on Ares and then alt tab to look something up, and all my keyboard inputs are still registering in the emulator (both controls and hotkeys), despite the window not being in focus. This only seems to happen when the game is running in fullscreen, though. The option to block input when the window is inactive doesn't seem to apply in fullscreen at all.

remutro:
Is this with more then one monitor present? Hesitant to disable all hotkeys when only one monitor is present as that is how you normally exist full screen.

OpenRift412:
> Is this with more then one monitor present? Hesitant to disable all hotkeys when only one monitor is present as that is how you normally exist full screen.

It is a multi-monitor setup, yes. I don't understand why disabling keyboard input when the fullscreen application isn't in focus would be a problem, though.

Issue #1201, opened 2023-08-20
[Feature Request] Fullscreen Mode Changes.


kotiesd:
Hello, I'd like to suggest a couple of changes for when fullscreen mode is triggered.

1. Allow activating the Windows Key:
   Pressing the Windows key does not display the taskbar and start menu until you exit fullscreen mode.
   Having this option would allow quick access to them without having to leave the full screen.
   
2. Allow displaying the mouse cursor when in Fullscreen:
   The mouse cursor is hidden when in fullscreen mode.
   But some users (like myself), may simply prefer to have the cursor visible while in fullscreen mode.

These two convenience features could be added in the form of check boxes next to a brief description of what they do.
Thank you for considering these feature requests.

thom-yorkes-brother:
In what situations/platforms does the mouse cursor get hidden in fullscreen?
When I run ares v140 on windows, the mouse cursor does not get hidden in fullscreen (with either OpenGL driver or Direct3D video driver).
I actually came here to request that the cursor be optionally hidden when in fullscreen mode.
In order to hide the cursor in fullscreen mode for the WGL ruby backend, I had to handle `WM_SETCURSOR` messages in `VideoOpenGL32_WindowProcedure` and call `SetCursor(NULL)`.  It also required logic added to the `if(self.fullScreen)` check in WGL.cpp:initalize so that the cursor is shown in windowed mode and hidden in fullscreen.
I can post a PR, but I've only made these changes from the WGL ruby backend - and more changes would be needed for the other backends

jeremiahrose:
The cursor is hidden for me in fullscreen mode on MacOS (Apple silicon). Make it impossible to exit fullscreen mode or access the menus or settings.

Screwtapello:
The only way to enter full-screen is to press the hot-key, the only way to exit full-screen is to press the same hot-key. Even if the cursor were visible, you still wouldn't be able to use it to exit fullscreen mode, or access the menus or settings.

PR #2460, opened 2026-03-14
fix: focus handling for alt-tabbing when fullscreen'd on Windows and X11

awsms:
Combines Linux & Windows fixes for focus lost when alt-tabbing on fullscreen:

#1426
```
Using WS_EX_TOPMOST on a fullscreen window prevents Alt+Tab from switching to any different window. This is a bit of an annoyance as it ruins the point of Alt+Tab, especially in the case of a non-exclusive mode being set. It however is very very bad in any case ares hangs, and thus prevents being able to switch to Task Manager to actually kill ares. This means the average user could potentially run into a situation requiring them sign out or shutdown their PC, and thus potentially lose data in other apps.

It also happens to be the case that if the user attempts to Alt+Tab while in exclusive fullscreen, ares will hang for whatever reason. This commit doesn't fix that hang, but it does at least allow the user to Alt+Tab out of the situation.

Note that the windows are being set with WS_DISABLED more to avoid what I believe is a Windows bug. Normally WS_EX_NOACTIVATE should prevent clicking on the window from actually activating the window. However, if you were to Alt+Tab to some other window, then have that other window being shunk so ares ends up shown, you can proceed to click on the ares window, and it will actually activate that window. This thus takes focus away from ares' presentation window with generally no way to give it focus (assuming inputs get blocked). WS_DISABLED prevents mouse input from working on the fullscreen window (which it normally doesn't do anything due to WS_EX_NOACTIVATE, only in the weird case I laid out does it do something)

Also this commit likely breaks macOS/Linux (haven't tested / can't test those), probably need to check for focus differently depending on platform defines?
```

#2434
```
Prevent keyboard state updates and mouse pointer grab from remaining active when the focused X11 window does not belong to this process.

- Keyboard: ignore XQueryKeymap results unless the active _NET_ACTIVE_WINDOW resolves to our PID.
- Mouse: automatically release pointer grab when focus leaves our window/process to avoid stuck cursor after alt-tab.
- GLX: explicitly set _NET_WM_PID on the GLX child window so focus attribution works reliably.

This ensures input is only processed while ares owns focus and avoids global key state leakage or cursor capture issues under X11 and Xwayland. The fix does not impact the Drivers->Input setting "When focus is lost" behavior; controller will still be able to trigger keys if set as "Allow input".

This commit replaces the whole workaround I previously attempted with https://github.com/ares-emulator/ares/pull/2431, to let the X server handling the windows as usual, instead of messing around with the desktop-ui/ logic.
```

### DIFF
--- a/desktop-ui/input/hotkeys.cpp
+++ b/desktop-ui/input/hotkeys.cpp
@@ -181,7 +181,7 @@ auto InputManager::pollHotkeys() -> void {
   if(Application::modal()) return;
 
   if(settings.input.defocus != "Allow") {
-    if (!presentation.focused() && !ruby::video.fullScreen()) return;
+    if (!presentation.focused()) return;
   }
 
   for(auto& hotkey : hotkeys) {
--- a/desktop-ui/program/platform.cpp
+++ b/desktop-ui/program/platform.cpp
@@ -169,8 +169,8 @@ auto Program::audio(ares::Node::Audio::Stream node) -> void {
 }
 
 auto Program::input(ares::Node::Input::Input node) -> void {
-  if(settings.input.defocus != "Allow") {
-    if(!ruby::video.fullScreen() && !presentation.focused()) {
+  if(!driverSettings.inputDefocusAllow.checked()) {
+    if(!presentation.focused()) {
       //treat the input as not being active
       if(auto button = node->cast<ares::Node::Input::Button>()) button->setValue(0);
       if(auto axis = node->cast<ares::Node::Input::Axis>()) axis->setValue(0);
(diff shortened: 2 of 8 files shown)


remutro:
Full screen window behavior here is exactly how I setup it up in my Direct3D11 draft PR and I don't plan on changing that, so I think making these changes are a good idea. 

I was looking into https://github.com/ares-emulator/ares/issues/2365 and was making most of the same changes proposed here in regards to input. I'm just peaking at my old branch cause I didn't finish but at the time I also needed to remove the full screen check in desktop-ui/program/status.cpp, updateMessag() for proper input block/defocus in the status bar. 

I will try to set some time aside soon to test out these changes on Windows & Linux.

remutro:
There is another old request for changing full screen behavior related to these changes: https://github.com/ares-emulator/ares/issues/1201

It also requests some full screen mouse cursor changes but not sure if that is really necessary.

PR #2462, opened 2026-03-14
chord hotkeys, hotkeys for load/save state to slot 1-9

awsms:
This PR expands the hotkeys functionality by adding support for hotkey combos and user-defined actions.
Still a WIP, I'm not sure where to go with the custom actions. Imo load/save state 1-9 should have been individual actions already.


awsms:
Got rid off the "user custom hotkeys" logic, and instead put the Load/Save State 1-9 behind a "Show Save/Load Slot Hotkeys" checkbox.
Since it was requested, we can also now bind joypad chords (for ex `Xbox guide+LT+A`) and mixed joypad+keyboard (`Ctrl+LT`).

PR #2471, opened 2026-03-18
desktop-ui: fix ws orientation after savestate load at boot

awsms:
WonderSwan and WonderSwan Color could boot with the screen rotated incorrectly when loading a savestate during startup (using `--save-state`).
Reapply the PPU/Screen/Orientation setting after unserializing a state so the screen rotation matches the restored orientation.

### DIFF
--- a/desktop-ui/program/states.cpp
+++ b/desktop-ui/program/states.cpp
@@ -34,6 +34,14 @@ auto Program::stateLoad(u32 slot) -> bool {
   if(!memory.empty()) {
     serializer state{memory.data(), (u32)memory.size()};
     if(emulator->root->unserialize(state)) {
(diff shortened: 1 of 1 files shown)


remutro:
FYI - The Game Boy Advance core supports screen orientation as well and should likely be added along with this. I believe it is currently these 3 cores.

awsms:
> FYI - The Game Boy Advance core supports screen orientation as well and should likely be added along with this. I believe it is currently these 3 cores.

Thanks for the heads up, do you know any games/event that could trigger the GBA to rotate in ares?
edit: just saw the "Orientation" submenu, though I have not been able to reproduce the load state bug

remutro:
GBA orientation is triggered from the Orientation sub-menu, same as WS/WSC.

Issue #2391, opened 2026-01-25
[N64] Disable Video Interface Processing is broken in interlaced modes


meauxdal:
**Describe the bug**
Settings > Video > Disable Video Interface Processing does not work properly in interlaced modes.

**To Reproduce**
1. Check the box for Settings > Video > Disable Video Interface Processing
2. Load an N64 ROM that uses interlaced video

**Screenshots**
<img width="1282" height="1036" alt="Image" src="https://github.com/user-attachments/assets/28d7aabc-ee3a-49fc-8201-2cf381dece79" />

<img width="1282" height="1036" alt="Image" src="https://github.com/user-attachments/assets/c58e136d-a737-4663-a1e0-7af47d1752a6" />

<img width="1282" height="1036" alt="Image" src="https://github.com/user-attachments/assets/f563b527-3cbe-4a20-b916-05a200993450" />

<img width="1282" height="1036" alt="Image" src="https://github.com/user-attachments/assets/0ef12e60-e2aa-44da-a913-1cbf4644e629" />

<img width="1282" height="1036" alt="Image" src="https://github.com/user-attachments/assets/9439861c-4a43-4387-a791-64da667a6dc6" />

<img width="1282" height="1036" alt="Image" src="https://github.com/user-attachments/assets/249734f6-17ac-452e-9602-654e18aad527" />

**Additional context**
Progressive modes work as expected using this function.

flagrama:
Is this different than using the same settings in other emulators using parallel-rdp?

meauxdal:
> Is this different than using the same settings in other emulators using parallel-rdp?

Which emulators have an analogue to this setting?

flagrama:
Good question. RMG doesn't seem to let you set the "serrate" bit to 0 but ares sets all of the vi bits to 0

https://github.com/ares-emulator/ares/blob/daa781706a62718de2ae3c71588e045630814a7b/ares/n64/vulkan/parallel-rdp/parallel-rdp/video_interface.hpp#L66-L74

https://github.com/ares-emulator/ares/blob/daa781706a62718de2ae3c71588e045630814a7b/ares/n64/vulkan/vulkan.cpp#L169-L171

Which seems important for the difference between progressive and interlaced video

https://github.com/ares-emulator/ares/blob/daa781706a62718de2ae3c71588e045630814a7b/ares/n64/vi/vi.cpp#L166-L167

[...]

PR #2489, opened 2026-04-04
n64: fix Vulkan VI scanout and weave setting

Drahsid:
I noticed an issue with Ares when testing James Lambert's WIP homebrew game, [Spellcraft](https://github.com/lambertjamesd/spellcraft).
The image Ares displayed seemed to be cropped (I tested this with the current release, the prerelease, and the previous release):
Ares

I thought this may possibly be a parallel rdp bug, so I checked Gopher64 and my own emulator:
Gopher

N64Emulator

Both of these work as expected, so I knew there was something specific with Ares.

I discovered that `vulkan.scanoutAsync(io.field)` being called before `vulkan.frame()`  could cause the VI scanout to be requested before parallel advanced its frame context. This would cause the scanout to be built against the old context instead of the fresh one.

On the way I also spotted another issue: when `Disable Video Interface Processing` was enabled, it was not actually disabling all of the VI processing flags. It was setting `options.vi` to `{false, false, true, false, false, false}`, which still left `serrate` enabled. Was this intentional?

I implemented these changes:
1. Call vulkan.frame() before vulkan.scanoutAsync() so parallel advances its frame context before scheduling the VI scanout
2. Restore all-disabled VI override when `Disable Video Interface Processing` is enabled

I also made a change to explicitly zero-initialize `RDP::ScanoutOptions`, but this does not change any behavior.
Another change was that the weave setting in the UI seemed to be writing to the wrong key.

### DIFF
--- a/ares/n64/vi/vi.cpp
+++ b/ares/n64/vi/vi.cpp
@@ -85,8 +85,8 @@ auto VI::main() -> void {
         if(++io.leapCounter == 5) io.leapCounter = 0;
         #if defined(VULKAN)
         if (vulkan.enable) {
-          gpuOutputValid = vulkan.scanoutAsync(io.field);
           vulkan.frame();
+          gpuOutputValid = vulkan.scanoutAsync(io.field);
         }
         #endif
         refreshed = true;
--- a/ares/n64/vulkan/vulkan.cpp
+++ b/ares/n64/vulkan/vulkan.cpp
@@ -167,7 +167,7 @@ auto Vulkan::scanoutAsync(bool field) -> bool {
   options.downscale_steps = supersampleScanout ? 16 : 0;
   options.persist_frame_on_invalid_input = true;  //this is a compatibility hack, but I'm not sure what for ...
   if(disableVideoInterfaceProcessing) {
-    options.vi = {false, false, false, false, false, false};
+    options.vi = {false, false, true, false, false, false};
   }
   if(!supersampleScanout){
     options.blend_previous_frame = weaveDeinterlacing;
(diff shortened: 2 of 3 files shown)


flagrama:
re: serrate: https://github.com/ares-emulator/ares/issues/2391

Drahsid:
That makes sense, I reverted that change.

rasky:
Why did you change the weave description? The previous one was the one was picked at the time. Especially mentioning that it's disabled when supersampling is active is an important bit of information

Drahsid:
The description is the same and is lower in the file: `weaveDeinterlacingHint.setText("Doubles the perceived vertical resolution; incompatible with supersampling").setFont(Font().setSize(7.0)).setForegroundColor(SystemColor::Sublabel);`

This can be seen in the UI (left is v147):
image

My understanding when I made this change was that the previous node name was a mistake, but actually looking into it further, it seems that none of the N64-specific `emulator->setBoolean(...)` have matching nodes in the tree. I wrote a quick test and  `emulator->setBoolean(...)` always returns false for every N64 option that uses it, so this appears to just be dead code, (unless I'm missing something?)

flagrama:
It seems to be a current incompatibility with ParaLLEl-RDP and applying settings while running. Normally these `if(emulator)` lines are called when the setting changes while running so that it can be applied to the emulator. As you can see from the "Note" at the bottom, that is not currently possible with any of the N64 Render Settings. If you try this with settings outside of the N64 Render Settings section you will see that actually happen. If that incompatibility is ever resolved though, the code is already in place to handle the live switching of the settings.

Your code is probably right that the text for the boolean "name" should match the setting name rather than its description. It seems that part is only for readability of the code? Someone can correct me on that though.

Drahsid:
I did some more testing on other systems where the image appears cropped and I noticed that it did not happen on AMD GPU. I then discovered that the issue only seems to appear on NVIDIA driver 591.xx - it doesn't appear in 595.xx or 581.xx, so the issue seems to actually be a driver regression for NVIDIA for the drivers from around Dec 2025 -> Jan 2026. This may also only be a Windows (11) issue (or even specific to 25H2) with those specific drivers (I didn't test on Win10/Mac/Linux/BSD.) My initial theory on the cause of this is likely incorrect.
Now knowing this, I think the patch is probably unnecessary, so I am going to close the PR.

PR #2498, opened 2026-04-11
Fix MMI extension loading and detection

remutro:
MMI files were not properly detected/associated with their respective LaserActive cores, so loading via drag & drop or from the command line would fail (from the command line you would need to manual specify the system). Looks like this broke when it was decided to change the name of the emulator core to their PAC names at which point the medium name no longer matched. This will also check system as defined in MediaInfo.json in the MMI file and load into the correct LaserActive core. 
Note that the alternate core name strings are used in other places still, as well as struct & file names not matching the current emulator core name. Since it wasn't necessary to change these, I'm not changing anything else. 

Also cleanup extension definitions for all optical based systems. They are defined in CompactDisc & LaserDisc structs which all optical mediums derive from one or the other so specifying them in individual mediums is redundant. 

I'm including a small fix for an incorrect setting I spotted. I have no idea why the colorbleed setting was used here.

### DIFF
--- a/mia/medium/playstation.cpp
+++ b/mia/medium/playstation.cpp
@@ -1,5 +1,12 @@
 struct PlayStation : CompactDisc {
   auto name() -> string override { return "PlayStation"; }
+  auto extensions() -> std::vector<string> override {
+#if defined(ARES_ENABLE_CHD)
+    return {"cue", "chd", "exe", "ps-exe"};
+#else
+    return {"cue", "exe"};
+#endif
+  }
   auto load(string location) -> LoadResult override;
   auto save(string location) -> bool override;
   auto analyze(string location) -> string;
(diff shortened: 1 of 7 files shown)


mia/medium/playstation.cpp
```diff
@@ -1,12 +1,5 @@
 struct PlayStation : CompactDisc {
   auto name() -> string override { return "PlayStation"; }
-  auto extensions() -> std::vector<string> override {
-#if defined(ARES_ENABLE_CHD)
-    return {"cue", "chd", "exe", "ps-exe"};
```
LukeUsher, line 5: Won't this prevent loading PS1 test programs, etc., that are distributed as .exe files? Can we allow optical media types to be extended here

remutro: Whoops, this would prevent loading PS executables. I guess I glossed over it as all the others looked the same. I've reverted the change to this file.

PR #2502, opened 2026-04-15
n64: many improvements to the RSP JIT

rasky:
This PR features many improvements to the RSP JIT to increase performance of the emulation. In particular:

* More opcodes are now JIT'd
* Cold codepaths are now moved at the end of the block, to avoid polluting cache for the hot codepaths
* Overhead around each opcode is reduced to basically zero for most opcodes:
  * There is no more a function epilogue function
  * PC addresses are computed at JIT time and are only materialized on branches
  * Clock is also computed at JIT time and materialized at C++ boundaries (still keep 100% of clock accuracy)
* DMEM is now kept in big-endian layout in memory. This makes it for faster generated code as RSP uses many misaligned memory accesses that are very complex to do with byte swapped memory layout.
* R0 is not cleared after every opcode. Instead, we avoid polluting that in both JIT'ed code and interpreter code. 
* We track registers assigned to constant values and do some folding in code generation. The main goal is to be able to predict alignment of addresses used in vector load/store opcodes, as that can simplify generated code quite a bit.

This PR has been tested with various test roms for regressions. More testing is always welcome.


remutro:
Need to bump serialization version again.

PR #2506, opened 2026-04-22
Upgrade libchdr and LZMA SDK to allow .7z support for ROM loading via nall archive decode

jpchow26:
Summary
- Add support for loading ROM content from .7z archives alongside existing .zip handling.
- Introduce a generic archive decode abstraction in nall and wire it into mia/desktop flows.
- Vendor and build the required LZMA SDK decoder sources as a dedicated thirdparty target.

Commit Structure
1. thirdparty: vendor LZMA SDK and add sevenzip-sdk target
   - Import LZMA SDK 26.00 C decoder sources under thirdparty/lzma-sdk-26.00.
   - Add new static library target sevenzip-sdk in thirdparty/CMakeLists.txt.
2. nall: add archive decoder abstraction with 7z support
   - Add Decode::Archive wrapper and Decode::SevenZip implementation.
   - Hook nall build files to compile/link sevenzip support.
3. mia: support .7z archives across import and loading flows
   - Update file dialogs/filters and ROM loading paths to accept .7z.
   - Replace zip-only decode calls with generic archive decode where applicable.

Behavior Changes
- Firmware and ROM imports now accept both .zip and .7z in the relevant UI flows.
- Parent ROM lookup in MAME medium supports mixed .zip/.7z parent-child combinations.
- User-facing invalid ROM messages updated to mention .zip or .7z archives.

Notes
- Changes were intentionally split into logical commits to keep vendor import, core plumbing, and app behavior reviewable independently.


flagrama:
7zip is normally licensed under LGPL which I don't believe is compatible with ares' ISC style license, or at least the desire to only use permissive licenses. You didn't use any of that code here right?

jpchow26:
Right, I used the LZMA SDK directly which is public domain. See https://www.7-zip.org/sdk.html and the included license file in the third party folder thirdparty/lzma-sdk-26.00/README.md

remutro:
Do we need to import LZMA again? This gets imported as part of libchdr (for handling CHD files). Can we use what already exists?

ares/thirdparty/libchdr/deps/lzma-24.05

jpchow26:
deleted the new lzma 26 folder and moved what was needed to the lzma 24. Confirmed .7z ROM for GBA loaded and played, as well as a CHD for PlayStation

remutro:
> deleted the new lzma 26 folder and moved what was needed to the lzma 24. Confirmed .7z ROM for GBA loaded and played, as well as a CHD for PlayStation

It looks like you just copied the standalone lzma sdk over the existing lzma version that ships with libchdr, which is not maintainable. We need to update libchdr to the latest version, and it ships with lzma 25.01 so what is there will get overwrriten.  If what is included with libchdr isn't sufficient to provide 7z decryption (although this looks that is exactly what is provides although I haven't dug deeply into it) then we likely need to keep the additional dependency.

jpchow26:
The LZMA included in libchdr isn't complete to do 7zip archive handling. I will try updating libchdr to latest, and then grabbing the LZMA 25.01 SDK missing source files to use for archive opening.

bsdcode:
I think it's not a good idea to depend on the libchdr-provided lzma for this functionality:

1. Users who compile ares with `ARES_ENABLE_CHD=off` will be cut off from the independent 7z support.
2. Downstream packaging policies might demand to use system-provided libraries whenever reasonable possible to reduce duplicate building of the same software. For example, as the current maintainer of the libchdr and ares ports for FreeBSD, I actually patch the ares build to use the libchdr port instead of the vendored one. The suggested change would make this alot harder if ares depends on the vendored lzma from the vendored libchdr. While I do respect that the ares project doesn't need to cater to downstream policies in its decisions, I'd appreciate it if we could find another solution.

jpchow26:
So is my original plan to just have the latest LZMA SDK in third party most desirable?

remutro:
Sounds like it is best to include as its own dependency. When I suggested it, I wasn't aware there were pieces cut from the included version with libchdr. So I would leave the version included with libchdr as-is. The upgrade to the latest libchdr release is good.

jpchow26:
Ok, so leave the libchdr update. Put back the third party LZMA 26 SDK as a third party lib and use that for 7z archives.

jpchow26:
I don't feel comfortable making a PR that has crept up in scope updating libchdr when it's much simpler than that. I'll close this PR and open a new one.

PR #2508, opened 2026-04-23
Add ISViewer RAM to save states

remutro:
Add ISViewer RAM to save states

### DIFF
--- a/ares/n64/cartridge/cartridge.hpp
+++ b/ares/n64/cartridge/cartridge.hpp
@@ -40,7 +40,7 @@ struct Cartridge {
     u32  offset = 0;
   } flash;
   struct ISViewer : Memory::PI<ISViewer> {
-    Memory::Writable ram;  //unserialized
+    Memory::Writable ram;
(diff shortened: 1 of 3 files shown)


Dragorn421:
Can confirm this fixes the problem!

Copypaste of the context from discord https://discord.com/channels/976404869386747954/976463759935696977/1496932432421720105

> I figured out why this (isviewer stops working after loading savestates in libultra roms) happens in libultra roms
> 
> I'm not sure exactly how the original ISViewer worked, but how it works with libultra roms in ares is that `isPrintfInit` sets the 32-bit word at `0xB3FF0000` to `'IS64'`, which ares stores in the emulated isviewer ram, and then `is_proutSyncPrintf` reads back and checks for `'IS64'` at the same address. if the check fails nothing is printed.
> 
> The problem is simply that isviewer ram is not serialized as part of save states. i'm trying to fix that but being not super familiar with the codebase makes that a bit difficult even though it seems like an easy task. If someone could do it that'd be great!

PR #2524, opened 2026-05-01
n64: improve CPU JIT

rasky:
This is a overhaul of the VR4300 JIT to improve the N64 core performance:

 * Most opcodes are now generated in JIT code. This is a huge amount of code. And even so, only happy paths are JIT'd: slow paths are still deferred to C++. For instance, memory accesses stay in JIT if they are in cache. FPU operations stay in JIT if they don't raise exceptions. etc.
 * The JIT now converts larger blocks at a time. Instead of stopping at every branch, it stops only at absolute jumps (J, JR, JAL). Local branches are now mostly resolved in-place with direct jumps in generated code, which is much faster.
 * JIT code is now organized in sections of 4K (which matches the VR4300 TLB page). Sections are invalidated as a whole on memory writes (if the write affect a cacheline that has been recompiled). Within each section, we can resolve local branches like explained above, without risking to cross a TLB page boundary.
 * JIT is now fully GDB aware, so execution should be fast even though a breakpoint is set. Settings a breakpoint basically forces a block termination and that's it (before, GDB was basically causing JIT to compile single-instruction blocks (!)).
 * CPU now has a configurable interleaving with the rest of the hardware. By default, the CPU will run for 4096 cycles before synchronization, though this number can be less if there are explicitly scheduled interrupts or events. The number seems to be high enough to help with performance, and low enough not to cause issues with eg RSP communication.

This is an example of the performance improvements of this PR + the RSP JIT PR, over v147:

```
Mario Tennis opening where Mario first walks onto the court: 
Pre-JIT-Updates:     86 VPS,       PR: 125 VPS
Indiana Jones boot logo sequence with the spinning gear:
Pre-JIT-Updates:     81 VPS,       PR: 126 VPS
Star Wars Episode I Racer rotating selection room:
Pre-JIT-Updates:     125 VPS,      PR: 152 VPS
Snowboard Kids title screen:
Pre-JIT-Updates:     89 VPS,       PR: 115 VPS
Conker can be pretty slow in the opening pre-file select, too:
Pre-JIT-Updates:     112 VPS,      PR: 182 VPS
Crusin USA - car select:
Pre-JIT-Updates:     82 VPS,       PR: 144 VPS
T3D 24 HDR Bloom Demo:
Pre-JIT-Updates:     63 VPS,       PR: 115 VPS 
```
 
Testing is welcome, specifically:

 * Verify if n64-systemtest passes. FPU JIT is using some corner cases of AMD64 SSE, so I worry that it might misbehave on some specific microarchitectures.
 * Making sure commercial and homebrew games don't have regressions
 * GDB integration
 * Timing tests. There's still some development to do to minimize timing regressions

I'll keep it in draft pending further testing


Dragorn421 (commented):
small thing, I think you need to bump the `SerializerVersion` in `ares/n64/system/serialization.cpp`

Dragorn421:
bug? report, more details on ares discord https://discord.com/channels/976404869386747954/976463759935696977/1499960150566502460

behavior first appearing in 349342bd3 based on bisecting:
with homebrew mode on, with or without gdb attached, but if gdb attached without setting some breakpoints (todo: which?), a modern gcc build of oot-ntsc-1.0 crashes on title screen load

rasky:
> bug? report, more details on ares discord https://discord.com/channels/976404869386747954/976463759935696977/1499960150566502460
> 
> behavior first appearing in [349342b](https://github.com/ares-emulator/ares/commit/349342bd3124d145c08f5c6c6d35f894ff82bc62) based on bisecting: with homebrew mode on, with or without gdb attached, but if gdb attached without setting some breakpoints (todo: which?), a modern gcc build of oot-ntsc-1.0 crashes on title screen load

Fixed in 7819bed, thank you very much for the detailed analysis, it was very helpful

SatanNintendo:
I've tested a few games, it works fine for me, but I haven't tested that many.

There are some slowdowns in the V-Rally and Roadsters game, but in my case, the V-Rally game is not running at full speed.

In the Roadsters game, the game starts to slow down in the car selection menu, but during the race on the track, the game works fine.

My CPU: i7 2600K (this is a weak spot in my PC).

rasky:
The current state of this branch is promising: no known regressions. Quoting remutro:


> I've played about 10 minutes in the following games now against the latest cpu_jit branch. I didn't detect any issues. Hopefully good enough as issues popped up fairly quickly previously.
```
40 Winks
Banjo-Kazooie
Blast Corps
Body Harvest
Cruis'n Exotica
Donkey Kong 64
F-Zero X
Golden Eye 007
Jet Force Gemini
Killer Instinct Gold
Legend Of Zelda - Ocarina of Time
Mario Kart 64
Mortal Kombat 4
Paper Mario
Pilotwings 64
Quake
San Francisco Rush - Extreme Racing
SpaceStation Silicon Valley
Wave Race 64
```

There would be several other minor improvements possible, but I would say this is good enough as I need to move on to other projects.

LukeUsher:
Incredible work!

PR #2526, opened 2026-05-06
Remove "eat stdin" behavior on exit

Dragorn421:
Currently, ares "eats" stdin on non-windows. For example:

```
$ ares
OpenGL Version: 4.6 (Core Profile) Mesa 25.2.8-0ubuntu0.24.04.1, GLSL: 4.60
echo Hello World  # typed by me while ares was open
# ares closed
$  # notice no Hello World was echo'd
```

image

This is unlike every other program ever, which just does not do anything with stdin. For example `sleep`:

```
$ sleep 10
echo Hello World  # typed by me during the sleep
$ echo Hello World  # not typed by me; this is from the commands I typed during sleep
Hello World
$
```

image

-------------

This also causes ares to hang on exit when no stdin is attached, for example when running ares in the background from a terminal with `ares &` then closing ares:

image

-------------

This behavior was added in february 2019 without any context besides the comments that remain to this day in the codebase: a1411748bbc1bb3dad0966984d6111b12494cc46 (see `main.hpp`)

-------------

I suggest we just remove this "eat stdin" code which as good-willed as may have been and despite the intent of making the behavior less surprising, actually makes the behavior surprising because different from everything else, and also causes a hang when closing a background-launched ares.

### DIFF
--- a/nall/nall/main.cpp
+++ b/nall/nall/main.cpp
@@ -22,20 +22,6 @@ auto main(int argc, char** argv) -> int {
 
   main(Arguments{argc, argv});
 
-  #if !defined(PLATFORM_WINDOWS)
-  //when a program is running, input on the terminal queues in stdin
-  //when terminating the program, the shell proceeds to try and execute all stdin data
-  //this is annoying behavior: this code tries to minimize the impact as much as it can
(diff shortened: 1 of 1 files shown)


rasky:
Yes please, this behavior is extremely annoying

PR #2533, opened 2026-05-09
Update project to use Visual Studio 2026

### LINKED ISSUE #2308: [Feature Request] Support for building with Visual Studio 18 2026
**Is your feature request related to a problem? Please describe.**
I'd like to be able to build ares without having to modify `CMakePresets.json`.

**Describe the solution you'd like**
Add some handling of generator used on Windows based on if VS 2022 or VS 2026 is present.

**Additional context**
@jcm93 CMake wizard 🙏

remutro:
Visual Studio 2026 runners have finally been released on GitHub ( https://github.com/actions/runner-images/pull/14015 ). Update the project to use the latest version of Visual Studio. 

Fixes: https://github.com/ares-emulator/ares/issues/2308

### DIFF
--- a/.github/workflows/build-aux.yml
+++ b/.github/workflows/build-aux.yml
@@ -12,13 +12,13 @@ jobs:
         - RelWithDebInfo
         platform:
         - name: windows-x64-gcc
-          os: windows-2025-vs2026
+          os: windows-latest
           shell: 'msys2 {0}'
           msystem: mingw64
           install: make git cmake ccache ninja mingw-w64-x86_64-toolchain
           target-cmake-preset: windows-ci-mingw-native
         - name: windows-clang-cl-x64
-          os: windows-2025-vs2026
+          os: windows-latest
           windres: rc
           shell: bash
           target-cmake-preset: windows-ci-cl-native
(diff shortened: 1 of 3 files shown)


CasualPokePlayer:
Note that `windows-latest` just ends up currently being an alias to `windows-2025`, which has this notice when it's used
`NOTICE: windows-2025 requests are being redirected to windows-2025-vs2026 by May 12, 2026` (and looking around it seems this is more a gradual rollout as some repos are already getting this redirect, so this date is more just the completion of this rollout)

It's probably a good idea to not change `windows-latest` as that change is effectively automatic in 4 days (especially as the point of `windows-latest` is to automatically alias to the latest Windows version). Also that the `windows-2025-vs2026` was more of a beta image name in the first place (being available since February 6), so it might be the case it'll be deleted anyways when this rollout completes.

remutro:
windows-latest currently doesn't point to it, you will still get an older image that doesn't even have a cmake version that knows 2026 exists. I'll check again by May 12th and if windows-latest points to the correct image, then I will update this PR.

remutro:
The promotion to windows-latest is not scheduled to be complete until June 15th. So we can camp on this for another month and I can update, or we can accept as-is now and it will work with 2026. 

https://github.com/actions/runner-images/issues/14017

```
Breaking changes
Windows Server 2025 with Visual Studio 2026 image is ready to be the default version for the Windows-latest label in GitHub Actions

The "windows-latest" and “windows-2025” labels in GitHub Actions will be migrated to use Visual Studio 2026 by default. Customers needing Visual Studio 2022 must migrate to the windows-2022 image.

Target date
This change will be rolled out over a week beginning June 8, 2026 and will complete by June 15, 2026.
```

thekovic:
this PR should have been squashed when merging

PR #2540, opened 2026-05-19
Rename VPS to Emu speed, and introduce FPS estimation

rasky:
For consoles up to 4th gen approximately, most games refresh their screen at the video hardware refresh rate (or half of it, but more rarely). So the information most users were concerned about was whether Ares would keep up with the emulation speed on their computer or not. This has been expressed in terms of frames per second (originally using the "FPS" label, later changed into "VPS"). This never conveyed exactly what the user was looking for: for instance, an user seeing "FPS: 48" still wouldn't know if the emulator was almost reaching full speed while emulating a PAL system (~ 50 Hz), or far from full speed on a NTSC system (~ 60 Hz). But it was still useful enough, as the variance of the number would also give a hint whether the emu was full speed or not.

On 5th gen consoles, the number becomes even more confusing because many games don't run at the nominal refresh rate but at a lower rate. Since discussion about the actual game frame rates are common, users are often confused on whether the oscillating "VPS" number is referring to the actual game FPS rather than the Ares refresh rate. And in the homebrew scene, this happens even more often as developers tend to monitor closely the FPS of their productions.

This PR changes this number to be expressed as percentage, with the label "Emu speed":

Screenshot 2026-05-19 alle 17 16 37

This should be universally easy to read for both end users and homebrew developers, without any chance of being confused with the actual game frame rate. It is immediately readable and conveys the required information without having to know the expected system refresh rate (that often changes automatically with the region). 

Moreover, to help homebrew developers and curious retrogamers, we add an API to optionally show a "game FPS" counter, and allow cores to estimate the actual in-game frame rate. In this PR, we change the Nintendo 64 core to opt-in to this new feature:

Screenshot 2026-05-19 alle 17 21 13
Screenshot 2026-05-19 alle 17 23 21

This is a best effort estimation but it should work most of the time for N64 games.


Dragorn421:
When the ares Window is small enough, the left part of the status bar is cropped: 
image

this doesn't happen in master, probably because of the much shorter right part "XX VPS" (ares window as small as possible for me:)
image

ruby/audio/directsound.cpp
```diff
@@ -52,17 +52,19 @@ struct AudioDirectSound : AudioDriver {
...
   }
 
-  auto output(const f64 samples[]) -> void override {
-    if(!ready()) return;
+  auto output(const f64 samples[]) -> bool override {
+    if(!ready()) return false;
 
     _buffer[_offset]  = (u16)sclamp<16>(samples[0] * 32767.0) <<  0;
     _buffer[_offset] |= (u16)sclamp<16>(samples[1] * 32767.0) << 16;
     if(++_offset < _period) return;
```
Dragorn421, line 60: I don't know whether this should return true or false but it needs to return a bool now (this is causing all windows builds to fail)

desktop-ui/presentation/presentation.hpp
```diff
@@ -99,7 +99,7 @@ struct Presentation : Window {
     HorizontalLayout statusLayout{&layout, Size{~0, StatusHeight}, 0};
       Label spacerLeft{&statusLayout, Size{8, ~0}, 0};
       Label statusLeft{&statusLayout, Size{~0, ~0}, 0};
-      Label statusRight{&statusLayout, Size{90, ~0}, 0};
+      Label statusRight{&statusLayout, Size{240, ~0}, 0};
```
Dragorn421, line 102: 180 is enough space on my machine :tm: 
image
do you need more on yours? why could that be?

ruby/audio/directsound.cpp
thekovic: Based on the other backends, it should return `false`

PR #2549, opened 2026-05-28
Fix aspect correction menu radio items in GTK

remutro:
I have no idea why GTK behaves differently then other UI frameworks here and I didn't bother to look into it. I just re-arranged some calls and set them up like the scale menu radio item options which already worked, and that fixed the aspect correction menu radio items. I was able to test that GTK & Qt6 on Linux work, and Windows still works.


Dragorn421:
:heavy_check_mark: fixes the issue for me (GTK, Kubuntu Linux 24.04)

PR #2558, opened 2026-06-02
Restore old symlink behaviour on Linux

### LINKED ISSUE #2557: Relative symlinks broken on Linux; absolute symlinks change last used directory
**Describe the bug**
Relative symlinks to a file end up resolving relative to the directory _ares is called from_. For instance, say I had the following directory layout:
```
/home/castor215
/media/roms/all-n64-roms
/media/roms/curated-n64-roms
```
It's not unreasonable that links in `curated-n64-roms` would have the form `../all-n64-roms/Super\ Mario\ 64.zip`. If I open a terminal in my home directory, that same link would end up resolving to `/home/all-n64-roms/Super\ Mario\ 64.zip`, which obviously breaks things.

As an addition, absolute symlinks do load the ROM, but set the last used directory for that core to the directory the link points to, which is annoying.

**To Reproduce**
1. Create a relative symlink to a file in a different directory
2. Open ares in a terminal from a directory that is _not_ the same one as the symlink
3. Try to open the ROM through the symlink
4. Get the ROM not found message

**Expected behavior**
Relative symlinks open the ROM, and neither relative nor absolute symlinks change the last used directory to the link destination.

**Additional context**
I am on Debian and in v147 the OS handled symlinks just fine. I can make a PR for `desktop-ui/emulator/emulator.cpp` with something like:
```c++
168    #if defined (PLATFORM_LINUX)
169        location = program.openFile(dialog);
170    #else
171        location = directory::resolveSymLink(program.openFile(dialog));
172    #endif
```
but I don't know if this would break other Linux distros. This needs feedback from other users.

Castor215:
Closes #2557 

As expressed in the issue, this needs feedback from other Linux users. Did pre-v148 ares not handle symlinks on your distro? Did the changes in v148 fix any issues? I don't want to cause any regressions so I'm opening this as a draft pending feedback.

### DIFF
--- a/desktop-ui/emulator/emulator.cpp
+++ b/desktop-ui/emulator/emulator.cpp
@@ -165,7 +165,12 @@ auto Emulator::load(std::shared_ptr<mia::Pak> pak, string& path) -> string {
     filters.trimRight(":", 1L);
     filters.prepend(pak->name(), "|");
     dialog.setFilters({filters, "All|*"});
+#if defined (PLATFORM_LINUX) || defined (PLATFORM_BSD)
+    //leave resolving symlinks to the OS on linux
+    location = program.openFile(dialog);
+#else
     location = directory::resolveSymLink(program.openFile(dialog));
+#endif
   }
 
   if(location) {


remutro:
I plan on reviewing behavior again. I think there is likely more changes that we might want to make even if this works, but I'll know more when I have some time to spend on it.

bsdcode:
I can confirm v148 on FreeBSD is also affected. Didn't test previous versions. I couldn't test the proposed fix yet, but an additional `|| defined (PLATFORM_BSD)` would probably be necessary.

Castor215:
> I plan on reviewing behavior again. I think there is likely more changes that we might want to make even if this works, but I'll know more when I have some time to spend on it.

Fair enough. If it helps, this is the problematic line in `nall/nall/directory.cpp`
```c++
158        if(readlink(result.data(), buffer, PATH_MAX) < PATH_MAX) {
```
`readlink` is returning relative links "as is" i.e. keeping them relative. The only way I can see around this is detecting if the first character is not `/`, finding the path of the link itself in that case and constructing the path contained in the link manually.

When execution returns to Emulator::load, you would need to detect if `location` is a symlink (absolute or otherwise) or not so the last used directory gets set to the symlink's location.

bsdcode:
Fix is working on FreeBSD.

PR #2559, opened 2026-06-03
RetroAchievements support

awsms:
rcheevos wip :smile:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "thirdparty/rcheevos"]
+	path = thirdparty/rcheevos
+	url = https://github.com/RetroAchievements/rcheevos.git
--- a/desktop-ui/CMakeLists.txt
+++ b/desktop-ui/CMakeLists.txt
@@ -58,6 +58,13 @@ if(ARES_ENABLE_CHD)
 endif()
 
 if(ARES_ENABLE_RCHEEVOS)
+  target_sources(
+    desktop-ui
+    PRIVATE
+      retroachievements/platform/adapter.cpp
+      retroachievements/platform/adapters/nintendo.cpp
+      retroachievements/platform/adapters/sega.cpp
+  )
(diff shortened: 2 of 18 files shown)


LukeUsher:
Now that all the non-hardware accellerated drivers are gone from ruby; I think it would be great if we could integrate some kind of imgui context layer into ruby with backends for each of our drivers, and use this for rendering achievement unlocked overlays, achievement related status popups, etc.

Not having this is a significant reason I haven't completed the achievements work yet.

LukeUsher:
Consider submitting a PR to [ares-emulator/ares-deps](https://github.com/ares-emulator/ares-deps) for  the rcheevos and libcurl integrations, we should not include third party libraries in-tree unless there is a special reason to (eg: they require modifications and cannot be vendored)

nit: nall already has an HTTP implementation; however, it doesn't support TLS, so it is unsuitable for this
Should we consider expanding nall/http to cleanly wrap libcurl, or is it better to just remove nall/http?

LukeUsher:
Make sure to use developer mode when consuming rcheevos to eliminate the risk of account flags for using in-development software.

awsms:
> Consider submitting a PR to [ares-emulator/ares-deps](https://github.com/ares-emulator/ares-deps) for the rcheevos and libcurl integrations, we should not include third party libraries in-tree unless there is a special reason to (eg: they require modifications and cannot be vendored)

yes I'm well aware about it! I've already put  libcurl in https://github.com/awsms/ares-deps, but since this is just the early draft I didn't want to mess already with upstream ares-deps

> nit: nall already has an HTTP implementation; however, it doesn't support TLS, so it is unsuitable for this Should we consider expanding nall/http to cleanly wrap libcurl, or is it better to just remove nall/http?

that's also why I ended up using curl rather than tweaking nall:HTTP, imo using curl directly would be better since nall:HTTP doesn't seem to be used currently?

PR #2561, opened 2026-06-04
Make SDL the only available audio & controller input driver

remutro:
First steps towards moving to SDL as a single audio/input driver. Eventually the following needs to be done:

- Settings UI reworked/rearranged
- Need to implement mouse/keyboard input based on SDL


thekovic:
From a Windows perspective, I would be apprehensive merging this right now pending improvements to our SDL3 audio driver. At least some users (myself included) encounter issues with using SDL3 as audio driver such as degraded performance or even complete breakdown of emulator timing (causing the emulator to run at slow motion despite 60 VPS being reported).

WASAPI has proven more reliable and resilient to me so far.

remutro:
> From a Windows perspective, I would be apprehensive merging this right now pending improvements to our SDL3 audio driver. At least some users (myself included) encounter issues with using SDL3 as audio driver such as degraded performance or even complete breakdown of emulator timing (causing the emulator to run at slow motion despite 60 VPS being reported).
> 
> WASAPI has proven more reliable and resilient to me so far.

Have you tried again as of v148? One of the last fixes introduced should have resolved the SDL performance issue.

LukeUsher:
I'd go as far as removing ruby's audio and input abstractions here; having desktop-ui go directly to SDL and making SDL a hard dependency; SDL is much better than it was in the past, and only having one code path to develop and maintain without extra abstractions will make our lives easier.

thekovic:
> Have you tried again as of v148? One of the last fixes introduced should have resolved the SDL performance issue.

Yes, even on v148 performance with SDL shows to be 5-10% slower and with 1% lows dropping as low as 20% slower. Concrete numbers - with WASAPI I measure 100 VPS in a given scene in Indiana Jones; with SDL I measure 90-95 VPS with short dips to 80 VPS.

bsdcode:
I encounter issues with the SDL3 audio driver on FreeBSD. It struggles on the N64 core when the render resolution is increased (2x or 4x Native) and gets even worse when the game makes use of the expansion pak for improved graphics.

The game which I test for this is Castlevania: Legacy of Darkness. The VPS drop to ~56 and audio is popping heavily. Another one, Resident Evil 2, is not as bad, but still has degraded performance with noticable audio popping.

Increasing the latency does mitigate it a little bit, but it's still not good.

Other audio drivers work fine here (I mainly use and test OSS on FreeBSD, but ALSA and OpenAL also work fine here as a comparison) and can handle both games without performance problems.

For other less demanding cores like SNES and Genesis I didn't encounter problems with the SDL3 audio driver.

In its current state, I fear removing all other audio drivers and relying only on SDL3 would be a huge regression for some users.

LukeUsher:
merging for now, we will continue to improve/iterate during this development cycle

PR #2570, opened 2026-06-18
ruby: improve SDL3 audio performance

bsdcode:
Some platforms experienced very poor performance with the SDL3 audio driver, especially in combination with a demanding core like the N64. It is assumed that SDL couldn't keep up filling the output device buffer with the audio data from the input queue buffer. Increasing the amount of generated input data before feeding it to the device buffer has shown to greatly improve the performance.

This sparked the following idea and change to how the SDL3 audio driver fills its buffers: request a fixed-sized 10ms device buffer and apply the latency to the input buffer, i.e. the input buffer holds a multiple of 10ms worth of audio. This allows SDL to steadily consume audio from the input buffer and feed it into the 10ms device buffer, while the user can still control the latency.

Additionally add the value 30 to the possible latency options. This allows users to specify a lower latency than 40 if 20 is just a tiny bit shy of being good enough.

This is tested on FreeBSD. The N64 core is now perfectly playable if I set the latency above 20. Other tested cores, SNES and Genesis, are already perfect at 20. 10 is just too low for my system, but other users could have more luck with it.

This needs definitly testing on other operating systems, and more opinions if this proposed change is actually a good idea.

### DIFF
--- a/ruby/audio/audio.cpp
+++ b/ruby/audio/audio.cpp
@@ -9,7 +9,7 @@ auto Audio::create() -> bool {
   return initialize();
 }
 
-auto Audio::hasDevices() -> std::vector<string> { 
+auto Audio::hasDevices() -> std::vector<string> {
   _devices.clear();
   _devices.push_back("Default");
   int count = 0;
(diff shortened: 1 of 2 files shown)


remutro:
I did a quick test on Windows and on the car select screen in Cruisin' USA I saw an increase from 130 VPS in v148 to 145 VPS in this PR build using the SDL audio driver.

PR #2573, opened 2026-06-27
n64: Add GUI for Parallel-RDP overscan crop settings

thekovic:
### Description

This PR adds graphical user interface controls for Nintendo 64 overscan cropping settings in the Parallel-RDP renderer with a toggle option and 4 separate text boxes for each side.

GUI:

image

Example before enabling:

image

After:

image

### Possible improvements

- The overscan value controls should be spin boxes but I believe this control is currently missing in hiro.
- It would be great if this setting could be adjusted while the game is loaded (like in @HailToDodongo 's fork) because adjusting overscan to one's liking is likely to take a lot of fiddling by users.
- Add option to preserve the aspect ratio of the image left after overscan as been cropped. Currently, the remaining image is always forced into 4:3 or 16:9 if `Aspect: Anamorphic`. I think `Aspect: No correction` should mean that if I have 320x240 output and I crop 20px top and 20px bottom and then fullscreen ares on a 16:10 monitor, the image will fill my screen completely.

### DIFF
--- a/ares/n64/system/system.cpp
+++ b/ares/n64/system/system.cpp
@@ -31,6 +31,11 @@ auto option(string name, string value) -> bool {
   if(name == "Supersampling") vulkan.supersampleScanout = value.boolean();
   if(name == "Disable Video Interface Processing") vulkan.disableVideoInterfaceProcessing = value.boolean();
   if(name == "Weave Deinterlacing") vulkan.weaveDeinterlacing = value.boolean();
+  if (name == "Enable Overscan Crop") vulkan.enableOverscanCrop = value.boolean();
+  if (name == "Overscan Crop Top") vulkan.overscanCropTop = value.natural();
+  if (name == "Overscan Crop Bottom") vulkan.overscanCropBottom = value.natural();
+  if (name == "Overscan Crop Left") vulkan.overscanCropLeft = value.natural();
+  if (name == "Overscan Crop Right") vulkan.overscanCropRight = value.natural();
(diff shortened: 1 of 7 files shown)


rasky (changes requested):
As discussed on Discord, I think this is too granular.

We could reduce this to three settings:
1) full overscan (the whole potential display area excluding only the syncs)
2) default display area (what Ares does now). This shows only what 99% of gsmes configure as display
3) CRT emulation. Remove some X% from the previous one to simulate a CRT-like look

Maybe this should be a more general video setting rather than a N64 specific one.

I would start with something like this and then let's see what's missing out

thekovic:
I agree this should ideally be a general video setting rather than N64-only setting because the problem it's solving is one of video presentation on the user's display. I just started with an N64 implementation as proof of concept because Parallel-RDP already exposes the necessary knobs and I just had to write the GUI.

CRT overscan simulation (where equal percentage of the game's image is cut from each side) is just one use case.

Another use case (this is the one that gets expressed by users such as https://discord.com/channels/976404869386747954/1393218503305203722 , https://discord.com/channels/976404869386747954/1520177537202782398 ) is to remove the arbitrarily sized black borders that are presented by the game. Examples are Super Mario 64, GoldenEye, Banjo-Kazooie, Conker, Castlevania 64, Indiana Jones and the Infernal Machine, or even homebrew such as Uncharted Terra 64. To cover this with a preset, we'd have to try interpret them from the configuration of the VI registers.

PR #2577, opened 2026-07-14
n64/pi: index cartridge ROM base-relative so >256MiB ROMs boot

meeq:
The PI bus dispatches cart reads/writes with the full 0x1000'0000-based address and relies on the Readable power-of-two mask to strip the base. That only holds while bit::round(rom.size) <= 0x1000'0000 (<=256MiB): for a larger ROM the mask widens to 0x1FFF'FFFF and the 0x1000'0000 cart base leaks into the buffer index, so every cart access (including the IPL3 bytes the PIF's IPL2 checksum validates) reads ~256MiB into the wrong offset -> checksum fails -> PIF Error/NMI -> CPU parked at the reset vector.

Pass the base-relative offset (address - 0x1000'0000) at the two cart dispatch sites. Identical result for <=256MiB ROMs (mask already stripped the base); fixes >256MiB.

### DIFF
--- a/ares/n64/pi/bus.hpp
+++ b/ares/n64/pi/bus.hpp
@@ -49,7 +49,9 @@ inline auto PI::busRead(u32 address) -> u32 {
     return cartridge.isviewer.read<Size>(address);
   }
   if(address <= 0x1000'0000 + cartridge.rom.size - 1) {
-    return cartridge.rom.read<Size>(address);
+    //index base-relative: for ROMs >256MiB the power-of-two Readable mask no longer
+    //strips the 0x1000'0000 cart base (bit::round jumps past it), so pass the offset.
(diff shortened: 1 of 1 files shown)


meeq:
Need to investigate this further: this unblocked an experiment I was working on, but does not appear to be completely correct. Closing for now.

Jj0YzL5nvJ:
mupen64plus/mupen64plus-core#1103

rasky:
This should be included in a refactoring of PI bus. Each PI device should be a different class that registers into the PI bus, which would keep a list of such devices.
For each request coming from the CPU/DMA, the PI bus would forward the request to all peripherals registered into the bus, and the first one accepting it / replying to it, would win. This way, a "cartridge" can accept requests in any range it'd like, which is exactly how the real hardware works.

PR #2593, opened 2026-07-29
input: preserve relative mouse motion between polls

ZengGengSen:
## Changes

- Accumulate relative mouse movement until it is read and consumed by an input mapping.
- Maintain an independent read position for each `InputRelative` binding.
- Synchronize relative input positions when input bindings or mouse capture state changes.
- Apply the same relative-motion handling to the Raw Input, NSEvent, and Xlib mouse drivers.

## Motivation

In my testing, `InputManager::poll()` is called approximately once every 8ms on my system, with observed intervals generally no shorter than 5ms. At 60Hz, each frame lasts approximately 16.67ms, so input is normally polled twice per frame. 

Each poll retrieves the mouse-axis movement accumulated during the preceding approximately 8ms. However, emulator cores generally read relative input only once per frame. This means that `InputManager` may complete its second poll before the core reads the movement value.

The previous implementation retained only the mouse-axis movement reported by the most recent poll. The second poll therefore overwrote the result of the first, allowing the core to observe only the movement from the latest approximately 8ms. Mouse movement from the earlier approximately 8ms of the frame was lost.

This change accumulates all relative mouse movement produced between consecutive core input reads. As a result, even when `InputManager::poll()` runs multiple times within a single frame, the core can still retrieve the complete mouse-axis movement accumulated during that period.

### DIFF
--- a/desktop-ui/input/hotkeys.cpp
+++ b/desktop-ui/input/hotkeys.cpp
@@ -16,10 +16,10 @@ auto InputManager::createHotkeys() -> void {
   hotkeys.push_back(InputHotkey("Toggle Mouse Capture").onPress([&] {
     Program::Guard guard;
     if(!emulator) return;
-    if(!ruby::input.acquired()) {
-      ruby::input.acquire();
+    if(!inputManager.acquired()) {
+      inputManager.acquire();
     } else {
-      ruby::input.release();
+      inputManager.release();
     }
   }));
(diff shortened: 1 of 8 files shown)


ZengGengSen:
I do not have an Xlib or macOS test environment, so these changes should ideally be tested on the respective platforms after merging.

PR #2596, opened 2026-07-30
fix: EEPROM null pointer access

HailToDodongo:
a game trying to read/write to EEPROM without it being present causes a segfault, since it tries to read from an (almost) null-pointer

### DIFF
--- a/ares/n64/cartridge/joybus.cpp
+++ b/ares/n64/cartridge/joybus.cpp
@@ -1,47 +1,54 @@
 
 auto Cartridge::joybusComm(n8 send, n8 recv, n8 input[], n8 output[]) -> n2 {
   n1 valid = 0, over = 0;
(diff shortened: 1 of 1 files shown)


ares/n64/cartridge/joybus.cpp
```diff
@@ -1,47 +1,54 @@
...
-    //cartridge EEPROM (4kbit)
-    if(cartridge.eeprom.size == 512) {
-      output[0] = 0x00;
-      output[1] = 0x80;
-      output[2] = eepromBusy ? 0x80 : 0x00;
-      valid = 1;
-    }
+  // with no EEPROM present, all commands are invalid.
+  // this also prevents any access to the uninitialized cartridge.eeprom data
+  if(eepromPresent) {
```
rmmh, line 8: any reason not to do `if(!eepromPresent) { return }` ?

HailToDodongo: there is a  check for RTC further down

PR #2598, opened 2026-07-31
N64: add simple overclock setting

rmmh:
This allows playing games with mild overclocks for better FPS, similar to what can be achieved with hardware mods and as is commonly implemented on other emulators.


rasky:
I don't know if we want an overclock setting, but for sure the part about the RDP framecount is not something we want to add. We have a different pending PR to implement a real in-game frame counter which is more sophisticated. Please remove that part and leave only the Overclock setting for discussion.

PR #2601, opened 2026-08-05
N64: Add SC64 USB and SD card emulation

meeq:
Exposes an SC64-ish USB interface as a network socket that is compatible with the `sc64deployer remote` protocol. Also an SC64-flavored SD card interface that can be pointed at a file.

Successfully tested with [Buu342's N64-NetLib](https://github.com/buu342/N64-NetLib) and `sc64deployer test -r`


ares/n64/cartridge/cartridge.hpp
```diff
@@ -120,6 +122,8 @@ struct Cartridge {
     auto writeHalf(u32 address, u16 data) -> void;
   } isviewer;
 
+  std::unique_ptr<SC64> sc64;
+
```
rasky: I think Cartridge is our "generic" class for an actual game cartridge (think the physical grey cartridge you plug into the console). It is currently something like a superset of all possible physical cartridges that is then configured to enable/disable "hardware features" for each game (eg: which save type is present in the cartrdige).

SC64 is a different thing. It is a different cartridge, which is totally different, internally. So I believe SC64 should not be a member of Cartridge but rather another global variable. Then you would attach to the PI bus one or the other.

meeq: https://github.com/ares-emulator/ares/pull/2601/commits/cc89b1c88e22b0f5696ce1f4bb2f1754780b9891

LukeUsher: Normally, Cartridge would be an abstract representation of what goes into the cartridge port, and then have a variety of boards/pcbs that inherit from it that implement each physical type of cartridge, and SC64 would be one of these pcbs, as would any gameshark or any other cartridge port device.

However, we haven't really been doing that with N64 so far, and since @rasky is essentially the codeowner of this core at this point, I'm happy with this solution too.

rasky: We are basically using the PIDevice for that at the moment since in the N64 case there are "interesting" things that can be connected to the cartridge port, including the 64DD, etc. 

Cartridge right now is a superset of all game cartridges (gamepaks). We could rename it as such to avoid misunderstandings.

There is indeed a vacant space for a "generic cartridge device" abstract class because the cartridge pin connector has more wires in addition to the PI bus; but we haven't had a need for that yet.

thekovic: > Cartridge right now is a superset of all game cartridges (gamepaks).

Maybe not necessary to rename but it would be nice to document this

PR #2605, opened 2026-08-14
md: add kmod message log channel

lfelipe:
Mega Drive homebrew and test ROMs can log text through the Gens KMod debug registers. $1e on the VDP control port takes one character at a time, and a zero write ends the line. ares currently ignores those writes, so a ROM that prints its progress that way is silent.

### DIFF
--- a/ares/md/vdp/debugger.cpp
+++ b/ares/md/vdp/debugger.cpp
@@ -31,6 +31,13 @@ auto VDP::Debugger::load(Node::Object parent) -> void {
   tracer.interrupt = parent->append<Node::Debugger::Tracer::Notification>("Interrupt", "VDP");
   tracer.dma = parent->append<Node::Debugger::Tracer::Notification>("DMA", "VDP");
   tracer.io = parent->append<Node::Debugger::Tracer::Notification>("I/O", "VDP");
+
+  //Off until asked for, like every other tracer here: `enabled()` is
+  //`file() || terminal()`, so setting either of them here would have this one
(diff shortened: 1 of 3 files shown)


LukeUsher:
Is this definitely safe on real hardware? Do any known hardware revisions behave badly to writing to this register?

In any case, as it is not a feature of real hardware, it should probably be gated behind the 'Homebrew Mode' option.

lfelipe:
Good catch on the gating. I went looking and there isn't a Homebrew Mode option on the Mega Drive core, only N64 and PS1 declare one. So I've just changed the Message tracer to not be force-enabled in load() via setTerminal(true)

Happy to wire up a proper Homebrew Mode for the MD core if you'd rather have it behind that as well.

On hardware safety: I haven't tested this on a real console, so I can't claim it first-hand. But the writes go to VDP register $1E, which doesn't exist on the 315-5313, only $00–$17 do, and ares was already throwing them away in the default: case of writeControlPort before this patch, so nothing about what's emulated changes. 

Used SGDK as the reference here, see src/kdebug.s, where KDebug_Alert writes 0x9E00 to 0xC00004 and comments it as "write null-terminated string byte by byte to port #0x1e", ending on a zero write. Plenty of SGDK homebrew runs on real hardware, unfortunately I only have a Hyperkin to test "real hardware" with and it's essentially an emulator so don't really consider it valid real hw testing.

LukeUsher:
after digging into it further it does appear real hardware ignores these accesses

PR #2608, opened 2026-08-23
n64: fix builds with ARES_UNITY_CORES=OFF

### LINKED ISSUE #2597: Windows builds using MSBuild (Visual Studio) fail if ARES_UNITY_CORES=OFF
**Describe the bug**
Since https://github.com/ares-emulator/ares/commit/3c63fbcfb445556790eb5e710b5de685a6a2035a Windows builds using MSBuild fail to compile because source files in Parallel-RDP (such as `ares\n64\vulkan\parallel-rdp\parallel-rdp\rdp_renderer.hpp`) that contain `#include "device.hpp"` (intending `ares\n64\vulkan\parallel-rdp\vulkan\device.hpp`) instead resolve to `ares\n64\pi\device.hpp`.

This error has not been observed in the CI builds because the base CMake preset for Windows CI builds `windows-ci-base` includes `"ARES_UNITY_CORES": true` which, for some reason, makes the problem go away. The "regular" developer preset `windows` described in the Wiki doesn't have this so defaults to `OFF` and fails.

**To Reproduce**
Steps to reproduce failing build:
1. Configure the project with `cmake --preset windows`
2. Build the solution
3. See many errors

Steps to reproduce working build:
1. Configure the project with `cmake --preset windows -DARES_UNITY_CORES="ON"` or with `cmake --preset windows-ci-cl-native`
2. Build the solution
3. See successful build

**Expected behavior**
Working build under all supported settings.

**Possible solutions**
Quick and dirty workaround which doesn't address the root cause is to rename `ares\n64\pi\device.hpp` to something else such as `ares\n64\pi\pidevice.hpp` or `ares\n64\pi\hamster.hpp`.

**Additional context**
Unless the root cause is found, this bug is likely to resurface whenever someone creates a source file that has the same name as something included in Parallel-RDP. Examples include fence.hpp, image.hpp, shader.hpp, or context.hpp.

I think we should consider adding non-unity builds to the CI pipeline to check for this or similar issues in the future.

invertego:
Fixes #2597

### DIFF
--- a/ares/n64/vulkan/vulkan.cpp
+++ b/ares/n64/vulkan/vulkan.cpp
@@ -1,5 +1,9 @@
 #include <n64/n64.hpp>
 
+//included here rather than in vulkan.hpp so that parallel-rdp headers are only
+//parsed in this translation unit
+#include "rdp_device.hpp"
+
 namespace ares::Nintendo64 {
 
 Vulkan vulkan;
--- a/ares/n64/vulkan/vulkan.hpp
+++ b/ares/n64/vulkan/vulkan.hpp
@@ -1,5 +1,3 @@
-#include "rdp_device.hpp"
-
 namespace ares::Nintendo64 {
 
 struct Vulkan {


thekovic:
huh that's how simple it was?

PR #2612, opened 2026-08-24
n64: speed up first boot of games with RTC

rasky:
There is no reason to advance ~1.7 billion times from the epoch. We can just initialize the RTC with the current time.

### DIFF
--- a/ares/n64/cartridge/rtc.cpp
+++ b/ares/n64/cartridge/rtc.cpp
@@ -10,7 +10,7 @@ auto Cartridge::RTC::load() -> void {
     present = 1;
     n64 timestamp = ram.read<Dual>(24);
     if(!~timestamp) {
-      time_t t = (time_t)0;
+      time_t t = time(0);
       struct tm tmm = *localtime(&t);
       ram.write<Byte>(16, BCD::encode(tmm.tm_sec));
       ram.write<Byte>(17, BCD::encode(tmm.tm_min));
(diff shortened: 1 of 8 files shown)


ares/n64/cartridge/rtc.cpp
```diff
@@ -20,10 +20,10 @@ auto Cartridge::RTC::load() -> void {
       ram.write<Byte>(21, BCD::encode(tmm.tm_mon + 1));
       ram.write<Byte>(22, BCD::encode(tmm.tm_year % 100));
       ram.write<Byte>(23, BCD::encode(tmm.tm_year / 100));
+    } else {
+      timestamp = time(0) - timestamp;
+      advance(timestamp);
```
invertego: This code also doesn't cope well with time going backward (for example, if the user adjusts their time zone). Not a new problem, but it might be nice to fix now. At present, advance() will count down from a negative value until it wraps to positive and finally reaches 0.

LukeUsher:
This also likely impacts other systems with an rtc, perhaps all of them, they were all built from the same base code

ares/n64/cartridge/rtc.cpp
rasky: time(0) isn't affected by timezone though. The bug is pre-existing and it only happens when clock goes backward. That shouldn't happen frequently if ever at all on modern computers connected to internet. 

I'd rather fix it separately and let this fix in for now

rasky: One thing we could do is to just ignore negative deltas. It should never happen and it's not the end of the world if the RTC clock doesn't move in that case. Worst case, users will readjust it in-game

rasky: Done

rasky:
> This also likely impacts other systems with an rtc, perhaps all of them, they were all built from the same base code

I've updated most of them to the same semantic:
* Initialize with PC local time (including timezone)
* In case of negative time warp (almost impossible), avoid almost infinite loops at boot

The only two I haven't touched are NGP as it is currently not serialized, and MSX as it currently never syncs with the host PC. I assume they should be aligned to the others as well, but maybe we can do in a follow-up.

PR #2622, opened 2026-09-05
linux: Register file types so they can be opened through double-click

### LINKED ISSUE #2621: .desktop file doesn't advertise that Ares can open files
The .desktop file in `desktop-ui/resource/ares.desktop` just says `Exec=ares`.

This means that users can't open files with it. Ideally, it would have `"%F"` added to signal that it can open files, and the `MimeTypes` fields filled in for all the types of ROMs it can support.

This would also make Ares be offered as an option when uninstalled, a Flatpak repository like Flathub is setup, and the user double-clicks on a file that Ares supports.

hadess:
Modify Exec field in .desktop file so that the Ares application advertises that it can open files (well, a single file at a time).

Then add all the relevant ROM and disk mime-types as being supported. Systems list used to select the mime-types to add: https://flathub.org/en/apps/dev.ares.ares
All the registered mime-types:
https://gitlab.freedesktop.org/xdg/shared-mime-info/blob/master/data/freedesktop.org.xml.in

Closes: #2621

### DIFF
--- a/desktop-ui/resource/ares.desktop
+++ b/desktop-ui/resource/ares.desktop
@@ -1,8 +1,9 @@
 [Desktop Entry]
 Name=ares
 Comment=Emulator
-Exec=ares
+Exec=ares "%f"
 Icon=ares
 Terminal=false
 Type=Application
 Categories=Game;Emulator;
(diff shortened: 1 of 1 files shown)


flagrama:
What happens when you compile without a core and try to open a file that uses that core?

hadess:
> What happens when you compile without a core and try to open a file that uses that core?

I didn't know cores were compile-option, I used the Flatpak manifest to build ares, and that was my one and only time compiling it.

The desktop file should then be generated from a `.in` file, it would look something like (in pseudo-code):
```
mimetypes=""
if "a26" in ${ARES_CORES}
  mimetypes += "application/x-atari-2600-rom;"
endif

# Generate ares.desktop from its .in variant, replacing `@ARES_MIMETYPES@` with the mimetypes list
generate_file("ares.desktop", "ares.desktop.in", "@ARES_MIMETYPES@", mimetypes)
```

I suck at cmake though, so can't help with the actual implementation.

I've added the Atari 2600 mime-type to the list though.

hadess:
And I've now added the Atari 5200 mime-type that should get added in:
https://gitlab.freedesktop.org/xdg/shared-mime-info/-/merge_requests/431

PR #2626, opened 2026-09-08
n64: add GameCube controller support

zoncabe:
Adds support for the GameCube controller protocol on the Nintendo 64 controller
ports, so homebrew built with libdragon can use a GameCube pad the same way it
does on real hardware with a passive adapter.

libdragon speaks the GameCube protocol directly (`JOYPAD_STYLE_GCN` in
`src/joybus/joypad.c`), but ares only implemented the N64 protocol, so a title
asking for a GameCube pad found nothing on the port.

### Implementation

A new `GamepadGCN` device alongside the existing `Gamepad` and `Mouse`,
implementing the Joybus commands a GameCube controller answers:

| Command | Purpose |
| --- | --- |
| `0x00` / `0xFF` | identify — reports identifier `0x0900` |
| `0x40` | read, in all eight analog modes |
| `0x41` | read origins |
| `0x42` | recalibrate |
| `0x43` | long read |

libdragon only issues `0x40` and `0x41`; the other two are implemented for
completeness.

Rumble is part of the read command itself (third request byte), not an
accessory in the controller slot, matching `__joypad_gcn_controller_rumble_toggle`.
The identify status byte reports whether the motor is currently running.

### Values

Analog ranges follow libdragon's `JOYPAD_RANGE_GCN_*`: stick ±100, C-stick ±76,
triggers 0-200. Origins are `JOYPAD_GCN_ORIGIN_INIT` (127 for the sticks, 0 for
the triggers), so a controller that never has its origins read still reads
correctly.

The analog mode packing was checked byte-for-byte against Dolphin's
`CSIDevice_GCController::GetData` across all eight modes.

Triggers are analog for most of their travel and engage a separate switch at
the bottom, as on the real controller. The threshold is 90% of travel, which is
the default Dolphin uses for the same control; that number is the most
arbitrary part of this patch and I'm happy to change it.

### Testing

Tested with my own libdragon homebrew: the pad is detected as
`JOYPAD_STYLE_GCN`, and the buttons, the stick and the analog C-stick all
behave correctly.

Rumble and the analog triggers are implemented but not yet exercised; I plan to
test both.

### Known limitation

The chosen device does not persist between sessions — ares reconnects the
default device on every load, for every core, so the GameCube pad has to be
selected again each time. I left that out of this PR to keep it contained, but
I'm willing to follow up with it if there's interest.

I intend to keep working on this if it doesn't meet the project's bar as-is;
happy to take direction on anything here.


ares/n64/controller/gamepad-gcn/gamepad-gcn.cpp
```diff
@@ -0,0 +1,374 @@
...
+  state.r = r->value() || state.analogR >= clickThreshold;
+
+  if(state.l) state.analogL = u8(OriginTrigger + RangeTrigger);
+  if(state.r) state.analogR = u8(OriginTrigger + RangeTrigger);
+
+  //opposing directions cannot be held at once on real hardware
+  state.up    = up->value()    & !down->value();
+  state.down  = down->value()  & !up->value();
+  state.left  = left->value()  & !right->value();
+  state.right = right->value() & !left->value();
```
flagrama: Is this accurate? You can absolutely press both directions with the hardware either disassembled, or a damaged or worn out d-pad.

zoncabe: clarified comment

LukeUsher: ares usually takes into account the physical limitations of the controller in these cases, especially when allowing it can cause unexpected behavior in games

PR #2628, opened 2026-09-08
fc: implement Datach external EEPROM

ZengGengSen:


LukeUsher:
Is there a reason why this adds a separate 24C01 implementation rather than using the existing eeprom component?

ares already has `X24C01`, which appears to match the external EEPROM used by Battle Rush; at least as far as nesdev is concerned: https://www.nesdev.org/wiki/INES_Mapper_157

I do see you implemented the 4-byte page buffer behaviors, if this is something we are genuinely missing, this should be a fix to the ares shared eeprom component.

nesdev also seems to note that the LSB first behavior is a Bandai quirk, not how the chip actually behaves, 
> Note that while the 24C02 serializes data bits from MSB to LSB (i.e. first bit 7 of any byte, then 6..0), Bandai persisted in believing that the X24C01 serializes data bits from LSB to MSB instead. The X24C01 is actually still a big-endian device, and multi-byte reads and writes increment the address as MSB-first.

ZengGengSen:
> Is there a reason why this adds a separate 24C01 implementation rather than using the existing eeprom component?
> 
> ares already has `X24C01`, which appears to match the external EEPROM used by Battle Rush; at least as far as nesdev is concerned: https://www.nesdev.org/wiki/INES_Mapper_157
> 
> I do see you implemented the 4-byte page buffer behaviors, if this is something we are genuinely missing, this should be a fix to the ares shared eeprom component.
> 
> nesdev also seems to note that the LSB first behavior is a Bandai quirk, not how the chip actually behaves,nesdev
> 
> > Note that while the 24C02 serializes data bits from MSB to LSB (i.e. first bit 7 of any byte, then 6..0), Bandai persisted in believing that the X24C01 serializes data bits from LSB to MSB instead. The X24C01 is actually still a big-endian device, and multi-byte reads and writes increment the address as MSB-first.

Initially I felt it diverged enough from the existing EEPROM component that I didn't want to modify the shared code and risk breaking other cores, so I put a standalone LSB-first EEPROM in the FC core. After the material you provided, though, I've merged it into the shared implementation.

PR #2631, opened 2026-09-12
[SNES] Add support for project Dream

### LINKED ISSUE #2629: [SNES] Project Dream: black screen
**Describe the issue**
The Project Dream leak loads but only shows a black screen. Other SNES games work for me just fine. I'm using the Linux flatpak at version v148.

**Rom Information**
*removed link*

remutro:
Adds SNES DB support for project Dream. Credit really goes to @JensRestemeier, who supplied the DB update. I tested it and it worked so I'm submitting it. 

Fixes: https://github.com/ares-emulator/ares/issues/2629


hackgrid:
It is not PAL though?
I guess the heurestics do not work correctly for this game, and this DB entry fixes this somehow by chance?

But it is not correct, the rom has a broken header and simply needs to be fixed, like many prototypes.

This attached patch fixes it correctly:
[DREAM.ips.zip](https://github.com/user-attachments/files/32137691/DREAM.ips.zip)

It does not work correctly by default in ares because it mis-identifies it as LOROM:

> game
>   sha256:   40c08a3c84f46919eb34e5396037d9a942d892ddf7da7882855dc333f7195809
>   name:     DREAM
>   title:    DREAM
>   label:    
>   region:   PAL
>   revision: 1.255
>   board:    LOROM-RAM#A
>     memory
>       type: ROM
>       size: 0x200000
>       content: Program
>     memory
>       type: RAM
>       size: 0x20000
>       content: Save



The correct manifest looks like this:

> game
>   sha256:   45f81811e0d639692ad2c3ec9723bb6af617a48de2b23b88793677689a90820d
>   name:     DREAM
>   title:    DREAM
>   label:    Project Dream
>   region:   NTSC
>   revision: 1.7
>   board:    HIROM-RAM
>     memory
>       type: ROM
>       size: 0x200000
>       content: Program
>     memory
>       type: RAM
>       size: 0x2000
>       content: Save



Not sure why the DB entry in this PR works, I guess it is because removes the LOROM line and ares defaults to some other mapping instead? (maybe because of the PCB entry?)

edit: seems like it, SHVC-1J0N-01 is HiROM:
https://snescentral.com/pcbboards.php?chip=SHVC-1J0N-01

remutro:
The board referenced has the same mappings as marking it as a generic HIROM. I'm not sure why we would want to specify HIROM-RAM here to generate a save file that isn't needed either. Also, the hash you specified doesn't match the original. However I did switch the region to NTSC as I believe this is correct, and just made some minor cleanup.

hackgrid:
The hash is different because the header changed.

remutro ***@***.***> schrieb am Sa., 12. Sept. 2026, 13:57:

hackgrid:
For the sram stuff, i just trusted the snes rom header tool haha

remutro ***@***.***> schrieb am Sa., 12. Sept. 2026, 13:57:

